PR #20, opened 2021-05-31
merge master with dev/sz

sebastian-zieba:


kevin218:
I'm confused by this request.  It looks like you're updating dev/sz from master, which shouldn't need a pull request.

PR #21, opened 2021-06-01
Dev/lk

lkreidberg:
- added __init__.py files to a few subdirectories so that setup.py can find them (and eventually make the package pip installable)
- commented out a few imports for WFC3 legacy code (missing image_registration.py and hst_scan.py)
- added julday.py from POET
- added .ecf to the .gitignore so that we don't have path clashes every time we merge a branch


kevin218 (approved):
I included the WFC3 files for reference when coding.  They should not be used at any point.

PR #27, opened 2021-06-16
Pull request from dev/sz to main: Data, MetaData Classes & AstroPy table

sebastian-zieba:
Changes:
- new class: Data: contains data extracted from the fits files (like data, err, dq, wave) and the respective subarrays.
- new class: MetaData: contains settings from the ecf and other values (but no arrays)
- spectra flux, time and wavelength are being saved into an astropy table using format='ecsv'
- added template ecfs for the hackathon in demos/S3
- "isplots" and "testing" flags have been moved from run_eureka.py to the ecf
-  fixed bug in run_eureka.py for S4 ev.eventlabel, ev.workdir -> ev3.eventlabel, ev3.workdir
- renamed savetable.py into astropytable.py (can also read a table now)
- moved the plot saving in S4 into a new file called plots_s4.py
- moved the unit conversion from S3 to nircam.py


lkreidberg (approved):
THIS LOOKS AWESOME

PR #34, opened 2021-06-22
Changing class and instance name for Data and Meta

sebastian-zieba:
Changes:
- Data & dat and Metadata & md are now: DataClass & data and MetaClass & meta
   - Note that the raw 2D spectrum is now stored in data.data
- fix in s3_reduce.py: median frame is now being saved in data instead of meta
- when the code is run now, a copy of the ecf file is saved into the workdir (if running S3) or lcdir (if running S4) [I think that's a good idea if the user wants to check which settings he used when running Eureka!]
- added astroquery as a requirement (svo_filters needs it)
- added a note in eureka/S5_lightcurve_fitting/utils.py that the EXOCTK_DATA is only important for Stage 5 [might scare people who only want to run S3 or S4]
- docs: more thorough quickstart 
- run_script.py: commented out all sys.path.append

### DIFF
--- a/eureka/S3_data_reduction/s3_reduce.py
+++ b/eureka/S3_data_reduction/s3_reduce.py
@@ -254,8 +254,8 @@ def reduceJWST(eventlabel):
     log.writelog('Copy S3 ecf')
     shutil.copy(ecffile, meta.workdir)
 
-    log.writelog('Generating figures')
     if meta.isplots_S3 >= 1:
+        log.writelog('Generating figure')
         # 2D light curve without drift correction
         plots_s3.lc_nodriftcorr(meta, wave_1d, optspec)
 
--- a/eureka/S3_data_reduction/s3_reduce.py
+++ b/eureka/S3_data_reduction/s3_reduce.py
@@ -24,9 +24,9 @@
 # 17. Produce plots DONE
 """
 
-import sys, os, time
+import os, time
 import numpy as np
-import multiprocessing as mp
+import shutil
 from ..lib import logedit
 from ..lib import readECF as rd
 from ..lib import manageevent as me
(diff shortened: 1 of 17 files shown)


lkreidberg (dismissed):

lkreidberg:
hey @gianninapr - looks like the builds passed for the PR, but we're still waiting on 1 test. Any idea why there's a time lag? I'd like to merge this before too many folks start working on Eureka

eureka/S3_data_reduction/s3_reduce.py
```diff
@@ -24,9 +24,9 @@
 # 17. Produce plots DONE
 """
 
-import sys, os, time
+import os, time
 import numpy as np
-import multiprocessing as mp
```
kevin218, line 29: Why remove multiprocessing as an import?

sebastian-zieba: It's currently not used in s3_reduce.py. I moved the background initialization into lib/util.py a while back

eureka/S4_generate_lightcurves/s4_genLC.py
```diff
@@ -45,42 +45,43 @@ def lcJWST(eventlabel, workdir, md=None):
...
+    if meta == None:
+        meta = me.load(workdir + '/S3_' + eventlabel + '_Meta_Save.dat')
 
     # Load Eureka! control file and store values in Event object
     ecffile = 'S4_' + eventlabel + '.ecf'
     ecf     = rd.read_ecf(ecffile)
-    rd.store_ecf(md, ecf)
+    rd.store_ecf(meta, ecf)
 
     # Create directories for Stage 3 processing
```
kevin218, line 56: Stage 4

eureka/S4_generate_lightcurves/s4_genLC.py
```diff
@@ -45,42 +45,43 @@ def lcJWST(eventlabel, workdir, md=None):
...
-        os.makedirs(md.lcdir)
-    if not os.path.exists(md.lcdir+"/figs"):
-        os.makedirs(md.lcdir+"/figs")
+    meta.lcdir = meta.workdir + '/S4_' + datetime + '_' + str(meta.nspecchan) + 'chan'
+    if not os.path.exists(meta.lcdir):
+        os.makedirs(meta.lcdir)
+    if not os.path.exists(meta.lcdir+"/figs"):
+        os.makedirs(meta.lcdir+"/figs")
 
     # Copy existing S3 log file
```
kevin218, line 64: S4

requirements.txt
```diff
@@ -15,4 +15,5 @@ svo_filters
 nbsphinx
 recommonmark
 myst_parser
+astroquery==0.4.2
```
kevin218, line 18: Should this be ">=" instead of "=="?

kevin218 (commented):
Left a few small comments within a couple of the files.  Should be super quick to fix.

kevin218:
I also don't understand what the third check is since there are only 2 Travis-CI builds (3.7 and 3.8).

requirements.txt
sebastian-zieba: Yeah that's right. Will fix that.

sebastian-zieba:
Adopted minor changes

kevin218 (approved):

Issue #40, opened 2021-06-25
Enable S3 to loop over multiple aperture sizes when specified in the .ecf file (intermediate).  Already implemented in POET, just need to replicate format.


kevin218:

kevin218:
In addition to giving a single value, it would be nice to be able to specify a range of aperture sizes (e.g., [min, max, stepsize]) in the .ecf file that the S3 pipeline can loop over.  The ecf reader can already handle lists, so that shouldn't be a problem.  Saving multiple results could be a challenge and we may want to create multiple sub-directories to handle each aperture size.

kevin218:
We may also want to implement something similar with the background region, but let's take one step at a time.

taylorbell57:
For me this currently breaks during the transition from S3 to S4 with or without multiple aperture sizes. Currently, the run_eureka.py file runs S3 and then S4 sequentially, and since the folder naming convention has changed, S4 can't find which *_Table_Save.txt file it is supposed to work on. For a single aperture, this could be remedied easily enough, but for multiple apertures it becomes unclear what the desired behaviour is: should S4 run over all the same apertures or only whichever was somehow preferred after S3? If it is the former, a glob search of *_Table_Save.txt files could be a clean way forward regardless of whether one or many apertures were considered.

kevin218:
I like the idea of a glob search, but we should make sure to check for existing S4 outputs within a given directory to avoid overwriting the outputs. Maybe throw a warning in that case.

PR #57, opened 2021-08-13
Issue #40: Add aperture/bg range functionality

gianninapr:
Pull request tackling issue #40 

Am unsure if the file naming scheme is what we want. I kept the date up to the minutes (took out the seconds). Here's some ideas I had because the simplicity of the current one worries me (having to escape certain characters) :

**Current One**
---------------------
S3_yyyy-mm-dd_hh-mm_ap=value_bg=value_eventname

**Some Other Ideas**
-----------------------------
S3_yyyy-mm-dd_hh-mm_ap-10_bg-15_eventname
S3_yyyy-mm-dd_hh-mm_ap10_bg15_eventname
S3_yyyy-mm-dd_hh:mm_ap=10_bg=15_eventname
S3_yyyy-mm-dd_ap=10_bg=15_eventname_run1    (2, 3, etc...) incrementally 

or any permutation of these single changes above at once


gianninapr:
Regarding the Travis failure to run, it seems the error is being caused by a bug in newer versions of numpy:

In a forum I found:
"The most recent version of NumPy has a bug that means it generates that error if it's imported twice in the same process."

as well as:

[RuntimeError: implement_array_function method already has a docstring](https://stackoverflow.com/questions/57750655/runtimeerror-implement-array-function-method-already-has-a-docstring-when-imp)

[Numpy github issue](https://github.com/numpy/numpy/issues/14384)

I played around for a bit by limiting the numpy version even down to 1.14.0 but nothing really worked. Then I took a closer look at exactly where the error is raised and the setup file that is failing seems to be from a dependency package, as it occurs in the following log message:

Running batman-package-2.4.8/setup.py

kevin218:
These are some good folder naming suggestions.  Personally, I like: S3_yyyy-mm-dd_eventname_ap10_bg15_runX, where "runX" only gets used if the folder already exists.  I guess you would need a recursive check for run1, run2, etc.

eureka/S3_data_reduction/s3_reduce.py
```diff
@@ -110,165 +99,194 @@ def reduceJWST(eventlabel):
...
-    log.writelog(f'\nFound {num_data_files} data file(s) ending in {meta.suffix}.fits')
-
-    stdspec = np.array([])
-    # Loop over each segment
-    # Only reduce the last segment/file if testing_S3 is set to True in ecf
-    if meta.testing_S3:
-        istart = num_data_files - 1
+    # check for range of spectral apertures
+    if isinstance(meta.spec_hw, list):
+        spec_hw_range = range(meta.spec_hw[0], meta.spec_hw[1], meta.spec_hw[2])
```
kevin218, line 104: Should we be inclusive in our range?  In other words, if someone sets spec_hw = [10,20,2], should the last value in the range be 18 or 20?  I would think 20, which would mean we need `range(meta.spec_hw[0], meta.spec_hw[1]+meta.spec_hw[2], meta.spec_hw[2])`

eureka/S3_data_reduction/s3_reduce.py
```diff
@@ -110,165 +99,194 @@ def reduceJWST(eventlabel):
...
+    else:
+        spec_hw_range = [meta.spec_hw]
+
+    #check for range of background apertures
+    if isinstance(meta.bg_hw, list):
+        bg_hw_range = range(meta.bg_hw[0], meta.bg_hw[1], meta.bg_hw[2])
```
kevin218, line 110: Same comment here as for `spec_hw`

kevin218 (changes requested):
I assume you used the latest version of s3_reduce so there should be no conflicts.

eureka/S3_data_reduction/s3_reduce.py
gianninapr: Valid, I assumed the user would understand python ranges aren't inclusive (since I'm used to it) but it's much better to do things so that they make sense logically instead of assume the user knows the code well! I'll change it

gianninapr:
> These are some good folder naming suggestions. Personally, I like: S3_yyyy-mm-dd_eventname_ap10_bg15_runX, where "runX" only gets used if the folder already exists. I guess you would need a recursive check for run1, run2, etc.

Yes this is also my personal favorite, I'll make it work!

gianninapr:
I worked on the generalized routine for the plotting we talked about. The only worry I have with it is that if the user quits a run halfway and doesn't clear their directories they'll have some folders of run 1 (let's say 4 out of the total 10) and then if they run the code again, 6 out of the 10 directories will say run 1 while 4 of them will say the next run (run 2).

kevin218 (dismissed):

kevin218:
Good point on your concern about having inconsistent run numbers if someone quits midway.  A simple workaround would be the create all of the folders nearly simultaneously (using a separate double for loop near the beginning) before moving to the current double for loop that does all the work.

gianninapr:
Got it done! Let me know what you think

kevin218 (approved):

PR #64, opened 2021-10-20
Big Documentation Update

taylorbell57:
Partly addressing Issue #2, I've gone through and updated lots of the documentation to add my S2 work as well as update the S3-S4 documentation which was old from things I'd added and others had added. I also went through all of the S2-S4 files (except the WFC3 files) and edited the function/class docstrings so that they are all well formatted and uniformly formatted for the readthedocs page. I didn't yet do anything with the docstrings in the lib folder which has a tonne of files, but I figure people are less likely to directly edit/interact with these, so they're a lower priority to get documented. Some of the files will have not-so-great docstrings because I just didn't understand enough of what was going on in the function to make a useful docstring, so if you see something you know about that is poorly commented, please suggest or make an edit!

This edit also changes the copyright statement to reduce upkeep work and adds a copyright page where the main developers are called out by name and a link points to the GitHub contributors page.

The tutorials will need a significant edit at some point as well since S3-S4 have significantly changed (and are likely to keep changing at least some), and S2 has been added, and S1 will be added before long too.

Edits to the code itself have been very minimal - the only real changes were moving some tiny bits of code around to reduce the amount of repeated code.


docs/source/index.rst
```diff
@@ -1,20 +1,19 @@
...
-At the moment the package is under development, and currently works on NIRCam data only.
+``Eureka!`` will eventually be capable of reducing data from any JWST instrument and fitting light curves.
+At the moment the package is under heavy development, and currently works on NIRCam, NIRSpec, and MIRI data only.
 The code is not officially associated with JWST or the ERS team.
 
-The code is separated into three parts or "Stages":
+The code is separated into five parts or "Stages":
 
-- Stage 3: Starts with Stage 2 data and reduces the data (performs background subtraction, etc. in order to convert 2D spectra into 1D information)
+- Stage 1: An optional step which calibrates Raw data (converts ramps to slopes). This step can be skipped if you'd rather use STScI's JWST pipeline's Stage 1 outputs.
```
kevin218, line 14: An optional step *that* calibrates

docs/source/index.rst
```diff
@@ -1,20 +1,19 @@
...
+- Stage 2: An optional step which calibrates Stage 1 data (performs flatfielding, unit conversion, etc.). This step can be skipped if you'd rather use STScI's JWST pipeline's  Stage 2 outputs.
+- Stage 3: Starts with Stage 2 data and reduces the data (performs background subtraction, etc.) in order to convert 2D spectra into 1D information
```
kevin218, line 16: in order to convert 2D spectra into *a time-series of 1D spectra*

eureka/S3_data_reduction/background.py
```diff
@@ -139,26 +152,13 @@ def fitbg(dataim, mask, x1, x2, deg=1, threshold=5, isrotate=False, isplots=Fals
...
                     #Use coefficients from previous row
                 else:
                     # Fit along spatial direction with a polynomial of degree 'deg'
                     coeffs    = np.polyfit(goodxvals, dataslice, deg=degs[j])
                     # Evaluate model at goodexvals
                     model     = np.polyval(coeffs, goodxvals)
-                    #model = smooth.smooth(dataslice, window_len=window_len, window=windowtype)
-                    #model = sps.medfilt(dataslice, window_len)
-                    '''
-                    if isplots == 6:
```
kevin218, line 153: Rather than removing code for this figure, we should make sure it works for Eureka!

eureka/S3_data_reduction/background.py
```diff
@@ -198,108 +197,4 @@ def fitbg(dataim, mask, x1, x2, deg=1, threshold=5, isrotate=False, isplots=Fals
         bg   = (bg.T)
         mask = (mask.T)
 
-    return bg, mask #,variance
-
-# STEP 3: Fit sky background with out-of-spectra data
-def fitbg2(dataim, mask, bgmask, deg=1, threshold=5, isrotate=False, isplots=False):
```
kevin218, line 204: `fitbg2` will likely be needed for NIRISS data reduction, so best to leave it in.

kevin218 (changes requested):

docs/source/ecf.rst
```diff
@@ -240,30 +271,79 @@ The names of the the files needed are given in the header as ``hdulist[0].header
...
 wave_min & wave_max
-'''''''''''''''''''''''
-
+''''''''''''''''''''
 Start and End of the wavelength range being considered
 
 
+allapers
+'''''''''
+If True, run S4 on all of the apertures considered in S3. Otherwise will use the only or newest S3 outputs found in the inputdir
```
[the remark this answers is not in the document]
taylorbell57: How about the following:

If True, run S4 on all of the apertures considered in S3. Otherwise the code will use the only or newest S3 outputs found in the inputdir. To specify a particular S3 save file, ensure that "outputdir" points to the procedurally generated folder containing that save file (e.g. set outputdir to /Data/JWST-Sim/NIRCam/Stage3/S3_2021-11-08_nircam_wfss_ap10_bg10_run1/).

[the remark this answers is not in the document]
kevin218: In thinking more about this, the results from running S3 with each aperture/bg size are saved into individual folders, so there will never be an instance where multiple save files are located in the same directory.  If `allapers == True`, how do we efficiently track down all of the S3 save files in different directories?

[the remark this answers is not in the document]
taylorbell57: There's two things I want to clarify here. The first is that I typically don't specify the inputdir to that level of precision (aka I set inputdir to just /Data/JWST-Sim/NIRCam/Stage3/, not /Data/JWST-Sim/NIRCam/Stage3/S3_2021-11-08_nircam_wfss_ap10_bg10_run1/). This means I don't need to update my ecf every single time I run the code since it will just use the newest one, but if I want to consider a specific run then you can give that level of precision in your inputdir variable. Second, the code uses the metadata file to figure out what folders the other aperture pairs would be located in by reusing the same code that created those different folders in the first place.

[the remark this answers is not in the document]
taylorbell57: The code that does this job of locating the other aperture folders is on this line: https://github.com/kevin218/Eureka/blob/a9f683bd7fc79d6fe3f2ba8debc334985ab18d1a/eureka/S4_generate_lightcurves/s4_genLC.py#L141

eureka/S3_data_reduction/background.py
taylorbell57: How does this plot differ from the one on this line:
https://github.com/kevin218/Eureka/blob/a9f683bd7fc79d6fe3f2ba8debc334985ab18d1a/eureka/S3_data_reduction/background.py#L184

Do we want to be keeping both or just one?

eureka/S3_data_reduction/background.py
taylorbell57: Okay, is there some more detailed information I can put into the docstrings to specify why one might choose fitbg2 over fitbg?

docs/source/ecf.rst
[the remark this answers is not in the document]
kevin218: Ok, this makes a lot more sense.  I'm good with the proposed changes.

eureka/S3_data_reduction/background.py
kevin218: Ah ha, that'll do.  No need for both.

eureka/S3_data_reduction/background.py
kevin218: fitbg2 uses bgmask, a mask for the background region! It enables fitting more complex BG regions than simply above or below a given distance from the trace.  This will help mask the 2nd and 3rd orders of NIRISS.

Issue #42, opened 2021-06-25
Matplotlib RuntimeError() whenever Eureka is imported and plt.show() is called. 


AarynnCarter:
Whenever Eureka is imported into a file, attempting to call plt.show() for a matplotlib.pyplot plot results in the error copied below. 

If Eureka is simply not imported, but the same anaconda environment is used, no such error appears. It's important to note that you don't have to actually use Eureka / specify any latex for this to appear. Simply importing the package whilst trying to call plt.show() results in this error. 

Temporary workaround is to add the following lines at some point in the code prior to any calls to plt.show()
```
from matplotlib import rc
rc('text', usetex=False)
```

However, this isn't the best long term solution. Current thought is that the error is being raised by the use of latex in label text within the Eureka package, and whether 'r' should be placed before the label text itself. I expect we need to discuss whether the 'r' is important or not, personally I've never had to include it and any latex commands (e.g. $\mu$) have been interpreted correctly by matplotlib. 

Final note, it seems this issue isn't unique to me, and at least 3 other people experienced this during the ERS hackathon. 

-------------------------------------------------------------------------------------------------------------------------

Mac OS Catalina 10.15.7
Python 3.9.5 Anaconda Environment
```
Traceback (most recent call last):
  File "/Users/acarter/anaconda3/envs/eureka/lib/python3.9/site-packages/matplotlib/texmanager.py", line 252, in _run_checked_subprocess
    report = subprocess.check_output(
  File "/Users/acarter/anaconda3/envs/eureka/lib/python3.9/subprocess.py", line 424, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
  File "/Users/acarter/anaconda3/envs/eureka/lib/python3.9/subprocess.py", line 505, in run
    with Popen(*popenargs, **kwargs) as process:
  File "/Users/acarter/anaconda3/envs/eureka/lib/python3.9/subprocess.py", line 951, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/Users/acarter/anaconda3/envs/eureka/lib/python3.9/subprocess.py", line 1821, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: 'latex'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
[...]

PR #65, opened 2021-10-21
Update limb_darkening_fit.py to correct matplotlib/TeX errors

AarynnCarter:
I created this branch to directly address issue #42 where matplotlib/TeX causes crashes at a variety of different stages in the pipeline depending on the user. 

After some testing, I'm almost certain that the following lines on one of the Stage 5 files has been editing the underlying matplotlib rcParams whenever Eureka is imported, and is directly causing these issues. I've attached some images below of what a plot looks like with and without these lines. Personally, I actually prefer the default matplotlib font, although we might need to make some changes to the font sizes directly in our plotting routines. 

```
rc('font', **{'family': 'sans-serif', 'sans-serif': ['Helvetica'], 'size': 16})
rc('text', usetex=True)
```

@taylorbell57 Could you install this branch and remove the additional lines you added to the jwst package file to see if the previous error disappears for you? The same issue cropped up again for me this morning (spurring me to try and fix things once and for all) and removing these two lines fixed things. 

@kevin218 @lkreidberg Changing the overall plotting style for Eureka seems pretty significant, but I think it's worth it for broader usability. Given you're leading things I'll let you decide whether to merge this or not!

ERS_DATASIMv2_NIRSpec_PRISM_TINY_x1dints
ERS_DATASIMv2_NIRSpec_PRISM_TINY_x1dints

### DIFF
--- a/eureka/S2_calibrations/__init__.py
+++ b/eureka/S2_calibrations/__init__.py
@@ -1,0 +1,2 @@
+# -*- coding: utf-8 -*-
+from . import s2_calibrate
--- a/eureka/S5_lightcurve_fitting/limb_darkening_fit.py
+++ b/eureka/S5_lightcurve_fitting/limb_darkening_fit.py
@@ -22,9 +22,6 @@ from bokeh.models.widgets import Panel, Tabs
 from . import utils
 from . import modelgrid
 
-rc('font', **{'family': 'sans-serif', 'sans-serif': ['Helvetica'], 'size': 16})
-rc('text', usetex=True)
-
 warnings.simplefilter('ignore', category=AstropyWarning)
 warnings.simplefilter('ignore', category=FutureWarning)
 
--- a/eureka/lib/plots.py
+++ b/eureka/lib/plots.py
@@ -2,15 +2,35 @@ import numpy as np
 import matplotlib.pyplot as plt
 from matplotlib import rc, rcdefaults
 
-#PLOT SETTINGS
-def use_tex_fonts(toggle, family='sans-serif', font='Helvetica', fontsize=16):
-    if toggle == True:
-        rc('font', **{'family': family, family: [font], 'size': fontsize})
-        rc('text', usetex=True)
-    elif toggle == False:
+# Function to adjust matplotlib rcParams for plotting procedures.
+def set_rc(style='preserve', usetex=False, from_scratch=False, **kwargs):
+    if not (isinstance(usetex, bool) and isinstance(from_scratch, bool)):
+        raise ValueError('"usetex" and "from_scratch" arguments must be boolean.')
+
+    if from_scratch:
+        # Reset all rcParams prior to making changes. 
         rcdefaults()
-    else: 
-        raise ValueError('TeX font toggle must be a boolean.')
+
+    # Apply the desired style...
+    if style == 'custom':
+        # Custom user font, kwargs must be from the 'font' rcParams group at the moment. 
+        # https://matplotlib.org/stable/api/matplotlib_configuration_api.html#matplotlib.rcParams
+        rc('font', **kwargs)
+    elif style == 'eureka':
+        # Apply default Eureka! font settings.
+        family='sans-serif'
+        font='Helvetica'
+        fontsize=16
+        rc('font', **{'family': family, family: [font], 'size': fontsize})
+    elif style == 'default':
+        # Use default matplotlib settings
+        rcdefaults()
+    elif style == 'preserve':
+        pass
+    else:
+        raise ValueError('Input style must be one of: "custom", "eureka", "preserve", or "default".')
+
+    rc('text', usetex=usetex) # TeX fonts may not work on all machines. 
 
 #SET PLOTTING FORMAT
 ebfmt   = ['bo',  'go',  'ro',  'co',  'mo',  'yo',
--- a/eureka/lib/plots.py
+++ b/eureka/lib/plots.py
@@ -1,6 +1,10 @@
 import numpy as np
 import matplotlib.pyplot as plt
 
+#PLOT SETTINGS
+def use_tex_fonts(family='sans-serif', font='Helvetica', fontsize=16):
+    rc('font', **{'family': family, family: [font], 'size': fontsize})
+    rc('text', usetex=True)
 
 #SET PLOTTING FORMAT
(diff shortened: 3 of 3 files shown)


taylorbell57:
I agree that having users directly edit the jwst file doesn't make sense, but contributing that change to the `jwst` package and requiring `eureka` to use that newer version might be a good option. But then we might run into issues with which `jwst` version is compatible with the current simulated observations (thankfully that'll be resolved once we actually get some real flight data). I am open to also not requiring tex fonts without an opt-out feature (not everyone will want to install tex with things like overleaf now), but I think it's best to resolve the conflict now rather than side-step it and let others figure out how to solve it down the road if they want to use a nicer font. Either way, setting this in a Stage 5 file that then gets applied to all scripts is definitely not an intuitive or sensible decision, so I suggest we move the two lines you cut to another file somewhere more obvious and add an option to skip them somehow. I'm not sure a .ecf option is the cleanest way though

AarynnCarter:
Yes, I suppose there are a few overlapping issues here. I totally agree that we should provide the solution to the jwst pipeline team and update Eureka to use that new version or higher, in fact, I'll do this now.  

Maybe this would be something to discuss on Monday, but I think it's important to decide between 1) opting-in to TeX fonts, or 2) opting-out of TeX fonts. I think 1) is the better choice, as making installation and code execution as hassle-free as possible seems more valuable to me. Like you say, we definitely shouldn't have these lines a Stage 5 file, my immediate thought is that we move them to the lib/plots.py file under a toggle-able parameter. For example... 

```
from eureka.lib import plots
plots.use_tex_fonts = True
plots.family = 'sans-serif'
plots.font = 'Helvetica'
```

taylorbell57:
I 100% agree to opt-in to tex fonts - opt-out is definitely going to cause more people headaches than slightly less nice plots will

AarynnCarter:
Okay, it's a little different to what I initially put, but to opt in for TeX fonts you can now add to whatever code you're running:

```
from eureka.lib import plots
plots.use_tex_fonts()
```

At the moment you can also pass a `family`, `font`, and `fontsize` argument to this for more manual customisation. Should be relatively easy to expand to other rc parameters if and when we need to.

kevin218:
It doesn't seem like there's a way to set Tex fonts to False. A boolean argument would work to toggle it on or off.

I like the idea of making this a general function for defining the rc parameters. This will make it possible to generate consistent, paper-ready plots. If that's the plan, let's rename the routine.

AarynnCarter:
Good catch Kevin, I didn't think about that. The new function takes a boolean argument to switch to a TeX font, or switch back to the matplotlib default. 

Are you thinking something like: `plots.make_paper_ready(True)`?

kevin218:
I was thinking `plots.set_rc(bunch of stuff)`

taylorbell57:
A **kwargs would be perfect for that

taylorbell57:
Hi all, what do you think about this instead?

**Current Code:**
```
def use_tex_fonts(toggle, family='sans-serif', font='Helvetica', fontsize=16):
    if toggle == True:
        rc('font', **{'family': family, family: [font], 'size': fontsize})
        rc('text', usetex=True)
    elif toggle == False:
        rcdefaults()
    else: 
        raise ValueError('TeX font toggle must be a boolean.')
```

**My Suggestion:**
```
def set_rc(user_custom=False, usetex=False, **kwargs):
    if not (isinstance(user_custom, bool) and isinstance(usetex, bool)):
        raise ValueError('user_custom and usetex arguments must be boolean.')

    if user_custom:
        # Let users do whatever they'd like
        rcdefaults()
        rc('font', **kwargs)
    else:
        # Fall back to some nice Eureka! recommended font settings
        family='sans-serif'
        font='Helvetica'
        fontsize=16
        rc('font', **{'family': family, family: [font], 'size': fontsize})

    rc('text', usetex=usetex)
```

If needed in the future, this could be further generalized upon to allow Eureka! customization to line properties and whatnot with separate font_kwargs, line_kwargs, xtick_kwargs, etc.

AarynnCarter:
Hey @taylorbell57, that looks much better! **kwargs are definitely the way forward here. It also looks pretty straightforward to expand things if we want multiple default plotting styles. 

If @kevin218 is happy, I can test/add this in today and merge with main.

kevin218:
Go for it!

AarynnCarter:
Okay, so I ended up tweaking things a little more to cover a few edge cases:

1) If we were to toggle on the Eureka settings, it didn't seem intuitive to me to call `set_rc(user_custom=True) ` to reapply the default matplotlib rcParams. Instead, I've switched things to a more specific `style` parameter.

2) If we were to set some `user_custom` style, but then decide later on in some code to switch the `usetex` setting, it would wipe all of the applied settings, unless they were provided once over. To avoid this, I've added a `preserve` style to simply not adjust the rcParams at all. 

3) Again for the `user_custom` case, if the function was called multiple times it might wipe some previously applied settings. So I've moved the `rcdefaults()` call to a `from_scratch` boolean so that users have the option to reset the rcParams entirely, or apply further changes. 

If we're happy with this I just need someone to approve and I can merge. 

```
# Function to adjust matplotlib rcParams for plotting procedures.
def set_rc(style='preserve', usetex=False, from_scratch=False, **kwargs):
    if not (isinstance(usetex, bool) and isinstance(from_scratch, bool)):
        raise ValueError('"usetex" and "from_scratch" arguments must be boolean.')

    if from_scratch:
        # Reset all rcParams prior to making changes. 
        rcdefaults()

    # Apply the desired style...
    if style == 'custom':
        # Custom user font, kwargs must be from the 'font' rcParams group at the moment. 
        # https://matplotlib.org/stable/api/matplotlib_configuration_api.html#matplotlib.rcParams
        rc('font', **kwargs)
    elif style == 'eureka':
        # Apply default Eureka! font settings.
        family='sans-serif'
        font='Helvetica'
        fontsize=16
        rc('font', **{'family': family, family: [font], 'size': fontsize})
    elif style == 'default':
        # Use default matplotlib settings
        rcdefaults()
    elif style == 'preserve':
        pass
    else:
        raise ValueError('Input style must be one of: "custom", "eureka", "preserve", or "default".')

    rc('text', usetex=usetex) # TeX fonts may not work on all machines. 
```

taylorbell57 (approved):

PR #85, opened 2022-01-05
BIG TESTING UPDATE: Read for more info

gianninapr:
Hello everybody! In this pull request we have a test for NIRCam up and running. These tests are intended for anyone who is uploading new code to run them and verify that none of their changes have broken the basic functionality of the code (a common example: changes in function names, etc). **It is recommended everyone runs this test whenever they upload new code**. The reason behind this is to make the workflow between collaborators easier, and to be thoughtful of everyone who will use your new code so they don't have to track down small changes and bugs. 

**HOW TO RUN TESTS**
----------------------------------

1. Make sure you're in the eureka/tests folder
2. run 'pytest'
3. If you wish to only run the NIRCam test: pytest -k test_NIRCam

**What about NIRSpec and MIRI?**
-----------------------------------------------
The full code functionality of NIRSpec and MIRI runs is still being tested. The need for smaller files for test runs for NIRSpec and MIRI is still a task at hand, and issue #86 was opened pertaining errors with NIRSpec data, therefore there isn't any working NIRSpec and MIRI tests yet, despite being written into the test file (commented out). But @jbrande will be helping with that.

**RUNDOWN OF CHANGES IN THIS PR**
------------------------------------------------------

- `modified:   ../../demos/S3/S3_template.ecf` - added 'save_output' meta parameter in order to allow tests to still save outputs necessary for also running S4 tests. Controls whether S3 outputs will be saved into files
- `modified:   ../S3_data_reduction/s3_reduce.py` - inlcudes save_output option in order to allow tests to still save S3 outputs (see above)
- `modified:   ../S4_generate_lightcurves/s4_genLC.py` - included aperture values for S4 output filenames (such that it matches the naming case from S3) *see lines 146,147*
- `modified:   ../lib/util.py` - updated description of path function from 'create path' to 'reads path'
- `modified:   test_general.py` - the meat of this PR: where all the new tests are written


eureka/tests/test_general.py
```diff
@@ -0,0 +1,159 @@
...
+        print("NIRCam test:")
+
+    # explicitly define meta variables to be able to run pathdirectory fn locally
+    meta.eventlabel='NIRCam'
+    meta.topdir='../tests'
+
+    # run S3 and S4
+    reload(s3)
+    reload(s4)
+    ev3 = s3.reduceJWST(meta.eventlabel)
```
kevin218, line 70: To avoid confusion, rename `ev3` to `s3_meta`

kevin218 (changes requested):

gianninapr:
**Note for light curve tests**
-------------------------------------
Currently the light curve fitting test does not run (check lines 36-62 in `test_lightcurve_fitting.py`) and has therefore been commented out. Since only a single function of Stage 5 is being run (`lc.fit`) and not the whole S5 process, the metadata from `S5_NIRCam.ecf` and `S5_fit_par.ecf` is not being read into the function properly. The metadata will have to be parsed explicitly for the test to run correctly (or the test will need to be restructured).

taylorbell57:
Just took a look through your work (without running it locally) and things look good to me at this point. I can work on getting a small MIRI/LRS test file generated and run through your test code in the next few days.

One good thing to add (can be done in another PR by @jbrande) would be testing of S2 (and S1 once it is merged in). Since there aren't many python interpretable outputs from S1 and S2, it might be the case that testing them is not so meaningful - that's something we can discuss.

gianninapr:
@taylorbell57 I agree it would be nice to have these! We'd need raw files for NIRCam and other instruments to have it go through the entire pipeline, and we'd need to verify they work. (As of right now NIRSpec is having problems with the files we had to run through stage 2 for example). I'll be around if my help is needed but I'm passing the baton to @jbrande since I have to focus in my main project this semester :)

gianninapr:
Added documentation about the tests and how to run them

kevin218 (approved):

PR #93, opened 2022-01-23
Implement shared parameters in transit model

meganmansfield:
One note here for anyone reviewing this pull request: I had some issues with dynesty taking forever to run, and I'm not too familiar with how dynesty works so I'm not sure if that was a bug in the code or just some issue with fitting so many free parameters at once. If someone who's more familiar with nested sampling could take a look at it and test that part out specifically, that would be great!


eureka/S5_lightcurve_fitting/likelihood.py
```diff
@@ -30,14 +30,19 @@ def ln_like(theta, lc, model, pmin, pmax, freenames):
...
+        Adding ability to do a single shared fit across all channels
     """
     # params[ifreepars] = freepars
     ilow = np.where(theta < pmin)
     ihi = np.where(theta > pmax)
     theta[ilow] = pmin[ilow]
     theta[ihi] = pmax[ihi]
     model.update(theta, freenames)
-    model_lc = model.eval()
+    if lc.share:
```
kevin218, line 42: Can we remove this if statement by passing `share=lc.share` and having `lc.longparamlist=None` by default?

eureka/S5_lightcurve_fitting/likelihood.py
```diff
@@ -148,7 +153,10 @@ def computeRedChiSq(lc, model, meta, freenames):
     - December 29-30, 2021 Taylor Bell
         Moved code to separate file, added documentation.
     """
-    model_lc = model.eval()
+    if lc.share:
```
kevin218, line 156: Same as above

eureka/S5_lightcurve_fitting/models.py
```diff
@@ -354,25 +354,67 @@ def eval(self, **kwargs):
...
+        #                 self.parameters.__setattr__(title,self.parameters.dict[param])
+
+                    
+
         # Set all parameters
-        for arg, val in self.parameters.dict.items():
-            setattr(bm_params, arg, val[0])
-        #for p in self.parameters.list:
-        #    setattr(bm_params, p[0], p[1])
+        if 'share' in kwargs:
```
kevin218, line 375: I think this will need to change to `if ('share' in kwargs) and (share==True):` once `lc.share` is always passed as an argument.

eureka/S5_lightcurve_fitting/plots_s5.py
```diff
@@ -28,27 +28,53 @@ def plot_fit(lc, model, meta, fitter):
...
         Moved plotting code to a separate function.
+    - January 7-22, 2022 Megan Mansfield
+        Adding ability to do a single shared fit across all channels
     """
     if type(fitter)!=str:
         raise ValueError('Expected type str for fitter, instead received a {}'.format(type(fitter)))
     
-    model_lc = model.eval()
-    residuals = (lc.flux - model_lc) #/ lc.unc
+    if lc.share:
```
kevin218, line 37: It might make more sense to leave this function as it was and put `if lc.share:' and the for loop around the call to `plot_fit()`. Does that make sense?

eureka/S5_lightcurve_fitting/plots_s5.py
```diff
@@ -76,35 +102,68 @@ def plot_rms(lc, model, meta, fitter):
     History:
     - December 29, 2021 Taylor Bell
         Moved plotting code to a separate function.
+    - January 7-22, 2022 Megan Mansfield
+        Adding ability to do a single shared fit across all channels
     """
     if type(fitter)!=str:
         raise ValueError('Expected type str for fitter, instead received a {}'.format(type(fitter)))
     time = lc.time
-    model_lc = model.eval()
```
kevin218, line 83: Same comment as above.

eureka/S5_lightcurve_fitting/s5_fit.py
```diff
@@ -161,25 +171,40 @@ def fitJWST(eventlabel, s4_meta=None):
...
 
-                # Get the flux and error measurements for the current channel
-                flux = meta.lcdata[channel,:]
-                flux_err = meta.lcerr[channel,:]
-
-                # Normalize flux and uncertainties to avoid large flux values
-                flux_err = flux_err/ flux.mean()
-                flux = flux / flux.mean()
+                # Temporary normalization to avoid large flux values (FINDME: replace when constant offset is implemented)
+                flux = meta.lcdata[0,:] / np.mean(meta.lcdata[0,:200])
```
kevin218, line 180: I think this normalization scheme conflicts with more recent changes.  Be very careful when merging conflicts in this pull request!

eureka/S5_lightcurve_fitting/s5_fit.py
```diff
@@ -222,4 +247,93 @@ def fitJWST(eventlabel, s4_meta=None):
                 if meta.isplots_S5 >= 1:
                     lc_model.plot(meta)
 
+            else:
```
kevin218, line 250: Can we be more clever here and avoid large if statements that have very similar code blocks???  This is probably worth a longer discussion.

eureka/lib/lsq.py
```diff
@@ -8,7 +8,10 @@ def modelfunc(freepars, lc, model, pmin, pmax, freenames, indep_vars):
     freepars[ilow]    = pmin[ilow]
     freepars[ihi]     = pmax[ihi]
     model.update(freepars, freenames)
-    model_lc        = model.eval()
+    if lc.share:
```
kevin218, line 11: Same comment as other files.

kevin218 (changes requested):

eureka/S5_lightcurve_fitting/s5_fit.py
```diff
@@ -161,25 +171,40 @@ def fitJWST(eventlabel, s4_meta=None):
...
+                    flux = np.concatenate((flux,meta.lcdata[i+1,:] / np.mean(meta.lcdata[i+1,:200])))
+                    flux_err = np.concatenate((flux_err,meta.lcerr[i+1,:] / np.mean(meta.lcdata[i+1,:200])))
 
                 # Load the relevant values into the LightCurve model object
-                lc_model = lc.LightCurve(t_mjdtdb, flux, channel, meta.nspecchan, unc=flux_err, name=eventlabel)
-
+                lc_model = lc.LightCurve(t_mjdtdb, flux, 0, meta.nspecchan, unc=flux_err, name=eventlabel,share=True)
+                
+                #Make a long list of parameters
+                longparamlist=[ [] for i in range(meta.nspecchan)]
```
taylorbell57, line 190: If there are no shared parameters, doesn't this just simplify down to the same stuff as what is at the bottom of the fitters.py file? That should make combining the shared and unshared codes easier

taylorbell57: This code should be in the fitters.py file rather than here though in my opinion

eureka/S5_lightcurve_fitting/s5_fit.py
```diff
@@ -201,7 +226,7 @@ def fitJWST(eventlabel, s4_meta=None):
                 if 'emcee' in meta.fit_method:
                     log.writelog("Starting emcee fit.")
                     model.fitter = 'emcee'
-                    lc_model.fit(model, meta, fitter='emcee')
+                    lc_model.fit(model, meta, fitter='emcee') #need to do all other fitters below here (all but lsq)
```
taylorbell57, line 229: What is this comment referring to?

eureka/S5_lightcurve_fitting/s5_fit.py
taylorbell57: See my comments above and elsewhere - I think we can start thinking of fits without shared parameters as just slightly special cases of shared fits which will simplify a lot of the code

taylorbell57 (changes requested):
Alright, my comments are scattered throughout. Generally though, I think it should be possible to think of unshared fits as just a slightly special case of the shared fits which should simplify a lot of the code's logic and reduce repeated code blocks with only slight modifications. If you want help or clarification on anything, please let me know!

meganmansfield:
Thank you both for the comments! I haven't gotten to this yet because of some proposal deadline stuff for next week, but I'll get on it as soon as possible.

eureka/S5_lightcurve_fitting/fitters.py
```diff
@@ -63,7 +64,10 @@ def lsqfitter(lc, model, meta, calling_function='lsq', **kwargs):
     model.update(fit_params, freenames)
 
     # Save the covariance matrix in case it's needed to estimate step size for a sampler
-    model_lc = model.eval()
+    if lc.share:
```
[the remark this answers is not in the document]
meganmansfield: I've tried updating the model.__init__ routine so that it loads in the parameters model.share, model.longparamlist, and model.nchan so that you can get rid of these extra parameters in model.eval!

eureka/S5_lightcurve_fitting/fitters.py
```diff
@@ -324,7 +333,10 @@ def dynestyfitter(lc, model, meta, **kwargs):
     best_model.components[0].update(fit_params, freenames)
 
     model.update(fit_params, freenames)
-    model_lc = model.eval()
+    if lc.share:
```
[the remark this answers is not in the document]
meganmansfield: Done!

eureka/S5_lightcurve_fitting/fitters.py
```diff
@@ -487,6 +501,15 @@ def group_variables(model):
         #     pinitial.append(param[0])
         #     pmin.append(param[0])
         #     pmax.append(param[0])
+        elif param[1] == 'shared':
```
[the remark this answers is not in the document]
meganmansfield: This was a holdover from a different way I tried to handle shared parameters initially. It's fixed now!

eureka/S5_lightcurve_fitting/likelihood.py
meganmansfield: Fixed

eureka/S5_lightcurve_fitting/likelihood.py
meganmansfield: Done

eureka/S5_lightcurve_fitting/models.py
meganmansfield: I changed it so that in the model "init" routine it will automatically set share=False if you don't specify share=True, so now it just needs to be if share==True here.
The end result of the changes is that a model initialized with share=True will use shared params, while a model initialized with either share=False or share not specified should result in a fit without shared params.

eureka/S5_lightcurve_fitting/plots_s5.py
meganmansfield: The problem with that is that plot_fit just takes in the lightcurve object, not a time and a flux or something like that. So when you call lc.flux in plot_fit it gives you an array that's all of the channels combined. That's why I put this all in the plotting, because you need a way to separate out the different channels from the large lc.flux object. But perhaps I'm missing an easy way to fix this problem?

eureka/S5_lightcurve_fitting/s5_fit.py
meganmansfield: Last time I branched from main, the normalization was thrown off so I was getting crappy fits because the out-of-transit had been normalized to a mean of 1.2 or something instead of 1. I think that until we have a constant offset function in the fit we'll just have to fix this part every time we make changes to something. But I can remove this if it conflicts during the merge.

eureka/lib/lsq.py
meganmansfield: Fixed

eureka/S5_lightcurve_fitting/models.py
```diff
@@ -354,25 +354,67 @@ def eval(self, **kwargs):
         # Generate with batman
         bm_params = batman.TransitParams()
 
+        #check for shared fit
```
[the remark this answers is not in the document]
meganmansfield: Done

eureka/S5_lightcurve_fitting/models.py
```diff
@@ -354,25 +354,67 @@ def eval(self, **kwargs):
...
+
+                # Use batman ld_profile name
+                if self.parameters.limb_dark.value == '4-parameter':
+                    bm_params.limb_dark = 'nonlinear'
+
+                # Make the eclipse
+                tt = self.parameters.transittype.value
+                m_eclipse = batman.TransitModel(bm_params, self.time, transittype=tt)
+
+                lcfinal = np.concatenate((lcfinal,m_eclipse.light_curve(bm_params)))
```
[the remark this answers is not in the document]
meganmansfield: Done, thanks for the suggestion!

eureka/S5_lightcurve_fitting/s5_fit.py
```diff
@@ -161,25 +171,40 @@ def fitJWST(eventlabel, s4_meta=None):
...
-                flux = meta.lcdata[channel,:]
-                flux_err = meta.lcerr[channel,:]
-
-                # Normalize flux and uncertainties to avoid large flux values
-                flux_err = flux_err/ flux.mean()
-                flux = flux / flux.mean()
+                # Temporary normalization to avoid large flux values (FINDME: replace when constant offset is implemented)
+                flux = meta.lcdata[0,:] / np.mean(meta.lcdata[0,:200])
+                flux_err = meta.lcerr[0,:] / np.mean(meta.lcdata[0,:200])
+                for i in np.arange(meta.nspecchan-1):
```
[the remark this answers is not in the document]
meganmansfield: That would work as a way to set it up too. Whatever normalization you had in there before wasn't working for me - it was setting the out-of-transit to something slightly larger than 1, so that when I tried to fit just a transit model it was always failing. But this issue is only because we don't have a constant multiplicative offset in the model at all, and once we add that in the weird normalization here won't be needed.

eureka/S5_lightcurve_fitting/s5_fit.py
meganmansfield: That was an in progress note for myself that I forgot to remove!

meganmansfield:
Ok, I've made some updates and the big remaining thing is the discussion on how to get rid of duplicate code. The only problem with doing that is the way we had the output files set up before. They were originally set up so that fit results from each channel would go into a different folder, with a separate log file, which prevents me from merging the "shared" code with the "not shared" code because then the log stuff all gets pushed into one file. Is it ok if I change this so there's just one output directory per run of the fitting code, all the figures go into that one directory, and there's just one log file with all the output for all of the channels? I wasn't sure if I should mess with this structure since I didn't know if it was set up for a specific purpose, but if we keep the separate folders structure it will be hard to shorten the code more.

eureka/S5_lightcurve_fitting/s5_fit.py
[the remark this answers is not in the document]
taylorbell57: We do have a constant term working now - you need to add polynomial into the list of models you're fitting in the ecf and add a c0 term in the fitted parameters ecf.

taylorbell57:
I am in favour of putting all the outputs from each channel go into the same folder (or at least subfolders nested in one shared folder). I vote go ahead and modify the output folder structure so that we can significantly reduce duplicated code. Having all channels in the same folder will also make a Stage 6 which makes transmission/emission spectra plots easier

eureka/S5_lightcurve_fitting/s5_fit.py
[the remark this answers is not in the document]
meganmansfield: Ah I see, thanks! In that case, I haven't edited the polynomial model yet to allow a shared fit, so for now I'll leave it this way, but I'll keep this in mind for after we figure out the basics of getting the shared format to work with models.

eureka/S5_lightcurve_fitting/s5_fit.py
kevin218: The `PolynomialModel` class with the `c0` free parameter fits for a constant offset.

meganmansfield: Got it now, thanks!

eureka/S5_lightcurve_fitting/s5_fit.py
meganmansfield: Done

meganmansfield:
Ok, I've pushed one more set of changes which (hopefully) fully integrate the shared params capability. I've removed a lot of the duplicated text and tried to condense everything. In order to get rid of that issue with the mean normalization conflicting with the main branch, I also went ahead and made it possible to use the polynomial model with shared parameters.
The only thing I haven't done yet is made it so you can do a shared fit across multiple visits, because I realized I don't really have an idea what the file outputs from s4 would be organized like in that case. But I think we can come back to that question later.
Let me know how everything looks!

eureka/S5_lightcurve_fitting/fitters.py
```diff
@@ -87,11 +88,17 @@ def lsqfitter(lc, model, meta, calling_function='lsq', **kwargs):
...
     # Compute reduced chi-squared
     chi2red = computeRedChiSq(lc, model, meta, freenames)
-
+    
     print('\nLSQ RESULTS:')
+    # if lc.share:
     for freenames_i, fit_params_i in zip(freenames, fit_params):
         print('{0}: {1}'.format(freenames_i, fit_params_i))
-    print()
+    # else:
```
taylorbell57, line 96: If you're done with this commented code, you should remove it

eureka/S5_lightcurve_fitting/fitters.py
```diff
@@ -460,10 +472,19 @@ def group_variables(model):
 
     - December 29, 2021 Taylor Bell
         Moved code to separate function to reduce repeated code.
+    - January 11, 2022 Megan Mansfield
+        Added ability to have shared parameters
     """
     # all_params = [i for j in [model.components[n].parameters.dict.items()
     #               for n in range(len(model.components))] for i in j]
-    all_params = list(model.components[0].parameters.dict.items())
+    if model.components[0].share:
```
taylorbell57, line 480: Can you explain the logic behind this chunk of code and maybe add some comments to the code itself as well? I'm having a hard time figuring out the function and nuance of the code

eureka/S5_lightcurve_fitting/lightcurve.py
```diff
@@ -185,28 +196,55 @@ def plot(self, meta, fits=True):
...
-        ax.set_xlabel(str(self.time_units))
-        ax.set_ylabel('Flux')
-        ax.legend(loc='best')
-        fig.tight_layout()
-
-        fname = 'figs/fig54{}_all_fits.png'.format(str(self.channel).zfill(len(str(self.nchannel))))
-        fig.savefig(meta.outputdir+fname, bbox_inches='tight', dpi=300)
-        if meta.hide_plots:
-            plt.close()
+        if self.share:
```
taylorbell57, line 199: This if/else pair seems like it could easily be merged so that unshared fits are just a special case. As of right now there is so much duplicated code

eureka/S5_lightcurve_fitting/models.py
```diff
@@ -152,7 +151,11 @@ def plot(self, time, components=False, ax=None, draw=False, color='blue', zorder
         label = self.fitter
         if self.name!='New Model':
             label += ': '+self.name
-        ax.plot(self.time, self.eval(**kwargs), '-', label=label, color=color, zorder=zorder)
+        if share:
```
taylorbell57, line 154: You could reduce duplication here by doing something like:
```
flux = self.eval(**kwargs)
if share:
    onelc=len.....
    flux = flux[onelc*......
ax.plot(self.time, flux, .........
```

eureka/S5_lightcurve_fitting/models.py
```diff
@@ -292,28 +304,46 @@ def _parse_coeffs(self, **kwargs):
...
 
     def eval(self, **kwargs):
         """Evaluate the function with the given values"""
         # Get the time
         if self.time is None:
             self.time = kwargs.get('time')
 
-        # Create the polynomial from the coeffs
-        poly = np.poly1d(self.coeffs)
+        longparamlist=self.longparamlist
```
taylorbell57, line 328: Why do this local renaming? Why not use `self.nchan` below, and you don't use longparamlist and paramtitles anymore it seems

eureka/S5_lightcurve_fitting/models.py
```diff
@@ -292,28 +304,46 @@ def _parse_coeffs(self, **kwargs):
...
+        nchan=self.nchan
+        paramtitles=longparamlist[0]
 
         # Convert to local time
         time_local = self.time - self.time.mean()
 
-        # Evaluate the polynomial
-        return np.polyval(poly, time_local)
+        # Create the polynomial from the coeffs
```
taylorbell57, line 335: Can this if/else also be collapsed in a simple way so that unshared fits are just a special case of shared fits? Like what if you set nchan=1 for unshared fits or something like that which would make all the unshared fits work like shared fits without any if/else statements? The only issue I see with that is that some string formatting I wrote uses nchan to figure out how many digits there should be (aka, should if be figure 511 if there's only 9 channels or 5101 if there's more than 10 channels). That should be easy to change though

eureka/S5_lightcurve_fitting/models.py
```diff
@@ -354,25 +354,67 @@ def eval(self, **kwargs):
...
-        for arg, val in self.parameters.dict.items():
-            setattr(bm_params, arg, val[0])
-        #for p in self.parameters.list:
-        #    setattr(bm_params, p[0], p[1])
+        if 'share' in kwargs:
+            longparamlist=kwargs['longparamlist']
+            nchan=kwargs['nchan']
+            paramtitles=longparamlist[0]
+            lcfinal=np.array([])
+            for c in np.arange(nchan):
```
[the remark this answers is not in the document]
taylorbell57: I think this point still stands - if self.nchan were set to 1 for unshared fits, then they would work the exact same as shared fits.

eureka/S5_lightcurve_fitting/plots_s5.py
taylorbell57: Can we just change how this function is called to have it take in arrays rather than model and lc objects to avoid this code duplication?

taylorbell57: Oh wait, my same trick of setting lc.nchannel=1 will also work here to get rid of the if/else pair. The only thing you'll have to be careful about is that zfill function call which would be affected by the nchannel=1 change. Perhaps you can precompute a lc.zfill parameter inside the lc.__init__ function and then set lc.nchannel=1 after you've computed the zfill value?

eureka/S5_lightcurve_fitting/plots_s5.py
```diff
@@ -76,35 +102,68 @@ def plot_rms(lc, model, meta, fitter):
...
-    plt.ylabel("RMS (ppm)", fontsize=14)
-    plt.xticks(size=12)
-    plt.yticks(size=12)
-    plt.legend()
-    fname = 'figs/fig52{}_'.format(str(lc.channel).zfill(len(str(lc.nchannel))))+'allanplot_'+fitter+'.png'
-    plt.savefig(meta.outputdir+fname, bbox_inches='tight', dpi=300)
-    if meta.hide_plots:
-        plt.close()
+
+    if lc.share:
```
taylorbell57, line 112: Same comment as above - it should be possible to remove the if/else

taylorbell57 (changes requested):
Alright, left some more comments throughout. My main suggestion is that if you set lc.nchannels=1 for unshared fits, then all of your shared code works for unshared fits (with the exception of some lines which contain zfill() calls which will need a minor tweak).

The other big thing is that there's been several changes to main since this PR was made, so merging it into main will take a fair bit of careful work to make sure the end result is as anticipated.

eureka/S5_lightcurve_fitting/fitters.py
meganmansfield: Done

eureka/S5_lightcurve_fitting/fitters.py
meganmansfield: Yeah, so the logic here is basically that for a shared fit you want the list of free parameters to include all the parameters for the individual visits (e.g., rp_0, rp_1, rp_2...t0_0, t0_1, t0_2...), but for a fit of a single channel the free parameters should only include that channel's parameter (rp_1, t0_1, etc.). This can also be simplified using your nchan=1 for a non-shared fit, and I'll also add comments to clarify what's going on.

taylorbell57: In my opinion, the latter is greatly preferred to reduce redundant labour. I can't really think of a time where one would want to enforce separate priors on each different wavelength's parameters, but @kevin218 might have a different opinion

eureka/S5_lightcurve_fitting/plots_s5.py
meganmansfield: Done with the nchannel=1 thing. I didn't get any errors from the zfill stuff, but check it out and see how it looks.

eureka/S5_lightcurve_fitting/models.py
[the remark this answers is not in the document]
meganmansfield: Done

eureka/S5_lightcurve_fitting/fitters.py
meganmansfield: Actually I got rid of this bit of the code because it's unnecessary with the new nchannels=1 thing. And to answer your question, the user just has to put in one value of rp and somewhere in the code it duplicates that. Perhaps that is a thing we will want to change at some point if we want to give the user the option to specify individual values for each channel, but for now it's working fine to do it that way and it makes it much less messy on the user end.

eureka/S5_lightcurve_fitting/lightcurve.py
meganmansfield: Done

eureka/S5_lightcurve_fitting/models.py
meganmansfield: I think the nchannel=1 thing fixed it anyway

eureka/S5_lightcurve_fitting/models.py
meganmansfield: I didn't check the string formatting for the titles, but otherwise this is condensed now.

eureka/S5_lightcurve_fitting/plots_s5.py
meganmansfield: Done

meganmansfield:
Ok, much code has been condensed and moved around to get rid of repeats. Please tell me it looks good this time, as I won't have a chance to work on this again before Monday.
Also, I know there will likely be some conflicts with main, but they should be straightforward to work out.

taylorbell57:
> Ok, much code has been condensed and moved around to get rid of repeats. Please tell me it looks good this time, as I won't have a chance to work on this again before Monday. Also, I know there will likely be some conflicts with main, but they should be straightforward to work out.

Things are looking good to me so far! If you want @meganmansfield, I can go through the process of merging main into this branch today so that the PR can be approved and merged?

meganmansfield:
If you have time, that would be wonderful, thanks!

meganmansfield:
Hey Taylor, I think I fixed the issue with the shared parameters implementation. It seems to be working for me now - will you double check that the commit I just pushed fixes things for you?

taylorbell57:
Yup, that works for me now!! I can work on the exponential ramp part if you want

meganmansfield:
Yay! And it's up to you - I'll have time to do it tomorrow morning.

taylorbell57:
I'm free for the rest of the day, so maybe I'll take a stab at it

taylorbell57 (commented):
Alright, I've tested this all as thoroughly as I can and don't see any issues anymore (or at least not any that are relevant to this PR). I suggest we approve and merge this, but what do you think @kevin218?

taylorbell57 (approved):

PR #95, opened 2022-01-25
Moving S5 models to a folder and other bug fixes

taylorbell57:
This PR primarily resolves Issue #90 by moving the eureka/S5_lightcurve_fitting/models.py file to many files in a new eureka/S5_lightcurve_fitting/models folder. All of the old utility is maintained with an __init__.py file inside the new models folder, while the code is now easier to parse through and easier for others to contribute to.

In preparing this PR, I encountered Issue #94 where reload statements break my ability separate models.py into a models folder. For now, I have simply removed the reload statements, but this will need to be discussed in more detail in that issue.

I have also started using a new computer (MacOS 11.6 instead of Ubuntu 20.04) and had significant challenges using Eureka! until I made the included changes to the requirements.txt file. If someone else can make a new conda environment and make sure that this requirements list is correct and ideally platform independent that would be great. I also alphabetized the requirements.txt file to make it easier to read through (before it was hard to see if a requirement was present or not).

I also updated the main 51## figure to include a third panel showing the uncalibrated flux and detector*systematics model which can be very helpful in troubleshooting.

Finally, I made some other small tweaks like ensuring that the logs along the S2->S5 path all get appended to the logs from the previous stage. Previously the S3 log would not include the log from S2, but now it will do so as long as the Stage 2 outputs were made by Eureka and not `jwst`.

PR #96, opened 2022-01-26
Permitting the use of WFC3 spectra

taylorbell57:
Alright, so I now have something that works for HST/WFC3 spectra. I had to make several simplifications over Kevin's original WFC3 code to get it to work within Eureka! simply, but I will work on peeling back those assumptions over the next while. For now, S3-5 have all been tested and work well with WFC3 and MIRI (there is no S1 or S2 needed for WFC3's ima files). See the image below for a rough first fit to a real WASP-12b eclipse (white light) - the fitted model looks weird since the HST data is sporadic and the model isn't evaluated where there isn't data, so matplotlib just connects the dots.

fig510_lc_dynesty

Note that this PR is less complex than the 121 Files changed would suggest - I imported POET's models_c folder which included a tonne of files, and I've merged in the edits in PR #95 already, so it'll look even smaller once that is merged into main.

The most important things to do for WFC3 moving forwards are adding in HST-relevant systematic models to S5, ensuring additional covariates are being passed to S5 for decorrelation, make sure plots still look good for HST's partial time coverage, and removing some of the simplifications I needed to make in S3.


kevin218:
This PR might just kill me...

taylorbell57:
Haha, there's a few easy files near the top and a maybe one at the bottom, and then just a bunch of files that are basically excactly copied from POET, so it's not that bad

taylorbell57:
One thing I haven't checked is whether Adina has written any S3 code for NIRISS that this will break - I'll do that now.

taylorbell57:
There we go - NIRISS's S3 code is still a work in progress, but now it's consistent with the other instruments

eureka/S3_data_reduction/bright2flux.py
```diff
@@ -156,13 +156,15 @@ def bright2flux(data, err, v0, pixel_area):
...
     srperas = arcsec**2.0
 
-    data *= srperas * 1e6 * pixel_area
-    err  *= srperas * 1e6 * pixel_area
-    v0   *= srperas * 1e6 * pixel_area
+    data.subdata *= srperas * 1e6 * pixel_area
+    data.suberr  *= srperas * 1e6 * pixel_area
+    data.subv0   *= srperas * 1e6 * pixel_area
 
     return data, err, v0
```
kevin218, line 169: Only return `data` here? `Err` and `val` are no longer, right?

eureka/S5_lightcurve_fitting/lightcurve.py
```diff
@@ -11,10 +11,6 @@
 from . import models as m
 from . import fitters as f
 from .utils import COLORS
-#FINDME: Keep reload statements for easy testing
```
kevin218, line 14: Possible merge conflict with another PR?

eureka/S3_data_reduction/wfc3.py
```diff
@@ -0,0 +1,498 @@
...
+    if meta.horizonsfile != None:
+        # Apply light-time correction, convert to BJD_TDB
+        # Horizons file created for HST around time of observations
+        bjd_corr = suntimecorr.suntimecorr(ra, dec, jd, meta.horizonsfile)
+        bjdutc      = jd + bjd_corr/86400.
+        data.bjdtdb = bjdutc # FINDME: Update to the lines below after I have the leapdir files
+        #data.bjdtdb   = utc_tt.utc_tt(bjdutc,meta.leapdir)   # FIX: should this be utc_tdb instead?
+    elif meta.firstFile:
+        print("WARNING: No Horizons file found. Using JD rather than BJD_TDB.")
+        data.bjdtdb   = jd
```
kevin218, line 238: Having a variable called `bjdtdb` but it actually being `jd` is a recipe for disaster. Maybe we use `hjdutc` or simply `jd` and leave `bjdtdb` undefined? This may cause headaches later on in the code

eureka/S3_data_reduction/wfc3.py
```diff
@@ -0,0 +1,498 @@
...
+    # jd  = np.array([2400000.5 + 0.5*(data.mhdr['EXPSTART'] + data.mhdr['EXPEND'])])
+    ra  = data.mhdr['RA_TARG']*np.pi/180
+    dec = data.mhdr['DEC_TARG']*np.pi/180
+    if meta.horizonsfile != None:
+        # Apply light-time correction, convert to BJD_TDB
+        # Horizons file created for HST around time of observations
+        bjd_corr = suntimecorr.suntimecorr(ra, dec, jd, meta.horizonsfile)
+        bjdutc      = jd + bjd_corr/86400.
+        data.bjdtdb = bjdutc # FINDME: Update to the lines below after I have the leapdir files
+        #data.bjdtdb   = utc_tt.utc_tt(bjdutc,meta.leapdir)   # FIX: should this be utc_tdb instead?
```
kevin218, line 235: Ask Laura Mayorga for updated utc_tt and leapsecond files.

eureka/S3_data_reduction/wfc3.py
```diff
@@ -0,0 +1,498 @@
...
+        data.subdata[n]     = spni.shift(data.subdata[n],  -1*meta.drift2D_int[-1][n,::-1], order=0,
+                                         mode='constant', cval=0)
+        data.submask[n]     = spni.shift(data.submask[n],  -1*meta.drift2D_int[-1][n,::-1], order=0,
+                                         mode='constant', cval=0)
+        data.subvariance[n] = spni.shift(data.subvariance[n], -1*meta.drift2D_int[-1][n,::-1], order=0,
+                                         mode='constant', cval=0)
+        data.subbg[n]       = spni.shift(data.subbg[n],    -1*meta.drift2D_int[-1][n,::-1], order=0,
+                                         mode='constant', cval=0)
+
+    # FINDME: The following cannot be run since we don't have the full time axis
```
kevin218, line 459: This could be a problem for some datasets if we can't run this step.

kevin218 (changes requested):

kevin218:
For the LC plotting, try using Model.interp() to evaluate model onto new time series.

eureka/S3_data_reduction/wfc3.py
taylorbell57: The issue with that is that we'd need an if/else statement everywhere afterward that uses jd if bjdtdb is undefined. A better option might be to re-write the code for all the instruments to simply use a `time` variable that has an associated `time_units` variable which can be checked to see if it is bdjtdb or jd or whatever. Thoughts?

eureka/S3_data_reduction/wfc3.py
taylorbell57: Yeah, I'm still thinking of how we can best enable this option without writing a completely different S3 for WFC3. Removing this one step was one of the biggest concessions I had to make with this first attempt and is something I plan to continue to work on. I can try to do so for this PR if you want, but it'll likely take me some time to work out a solution

taylorbell57: I ran into this same kind of issue with SPCA earlier. It is cetainly doable to rewrite everything so that we can sigma clip along the full time-axis, but it ends up being very RAM intensive since you need to store at least 2 copies of all the frames in RAM to do sigma clipping. Plenty doable with just a transit/eclipse, but it gets rough with a phase curve (and even worse if it's on JWST's larger detectors).

eureka/S3_data_reduction/wfc3.py
taylorbell57: Got these straight from her commits to the WFC3 repo

taylorbell57:
> For the LC plotting, try using Model.interp() to evaluate model onto new time series.

That won't help - matplotlib is already just interpolating between points, so Model.interp does the same thing (interpolate flux). What we need to do is super-sample the time array (or at least uniformly sample the time array) and then evaluate the model using that new time array so that we can see the full transit shape. And if we have other covariates, then things get even more complicated. I did this before for STIS, but I need to remember how

kevin218:
Ah, got it. Maybe we should update interp to perform a uniform interpolation since the current implementation is not terribly useful.

taylorbell57:
> Ah, got it. Maybe we should update interp to perform a uniform interpolation since the current implementation is not terribly useful.

Yeah, that should work fine for now, but as soon as we have independent variables other than time we'll need to model how those other variables (e.g. trace slope, 1D drift, HST orbital phase) vary with time in order to interpolate

kevin218:
The base use case for interp would be to enable plotting the normalized HST
light curve, where all of the systematics have been removed.,

I agree that it wouldn't work well for other cases, like plotting the raw
light curve.

taylorbell57:
> The base use case for interp would be to enable plotting the normalized HST light curve, where all of the systematics have been removed., I agree that it wouldn't work well for other cases, like plotting the raw light curve.

What do you think about the edit's I've made which make the following plot instead of the one above? I don't interpolate at all in the top plot given the issues with doing that, so I switched to using unconnected points for the top panel. I do re-evaluate the astrophysical model for the second panel though.

fig510_lc_dynesty

taylorbell57:
And any thoughts on my comments above with regards to the bjdtdb variable and full time-axis sigma clipping?

eureka/S3_data_reduction/wfc3.py
kevin218: That's how POET operates, where the time unit can also be orbital phase. Let's do it!

eureka/S3_data_reduction/wfc3.py
kevin218: Let's hold off on making changes inside this PR. A good compromise is to load in chunks of data to avoid ram issues. We could break things up by HST orbit or by a specific number of frames.

kevin218:
> > The base use case for interp would be to enable plotting the normalized HST light curve, where all of the systematics have been removed., I agree that it wouldn't work well for other cases, like plotting the raw light curve.
> 
> What do you think about the edit's I've made which make the following plot instead of the one above? I don't interpolate at all in the top plot given the issues with doing that, so I switched to using unconnected points for the top panel. I do re-evaluate the astrophysical model for the second panel though.


I like this solution. Nicely done!

kevin218 (dismissed):

eureka/S3_data_reduction/wfc3.py
taylorbell57: Working on this right now - I'll push some commits soon and then we can merge this branch into main. Just testing end-to-end with NIRSpec and WFC3 to make sure I haven't broken anything with this time and time_units change

taylorbell57:
Alright Kevin, I think this one is ready to merge. I've tested it with NIRSpec and WFC3 and it works well with the change from bdjtdb to a more generic time, with the units of the time variable being tracked in the meta.time_units parameter. I also found and patched some bugs while testing this PR, and it's possible I just found another but it has nothing to do with the edits I've made and is something I'll work on after this PR

eureka/S5_lightcurve_fitting/lightcurve.py
```diff
@@ -198,11 +198,18 @@ def plot(self, meta, fits=True):
         # Draw the data
         ax = fig.gca()
         ax.errorbar(self.time, self.flux, self.unc, fmt='.', color='w', ecolor=self.color, mec=self.color, zorder=0)
+        
+        # Make a new color generator for the models
+        plot_COLORS = color_gen('Category10')
```
kevin218, line 203: The intention is to avoid different colored model lines on colored data points because each color is associated with a specific wavelength. I suggest using 'Greys9' instead for the models.

eureka/S5_lightcurve_fitting/s5_fit.py
```diff
@@ -130,10 +131,23 @@ def fitJWST(eventlabel, s4_meta=None):
                 params = p.Parameters(param_file=meta.fit_par)
 
                 # Subtract off the zeroth time value to avoid floating point precision problems when fitting for t0
-                t_offset = int(np.floor(meta.bjdtdb[0]))
-                t_mjdtdb = meta.bjdtdb - t_offset
-                params.t0.value -= t_offset
-
+                offset = int(np.floor(meta.time[0]))
```
kevin218, line 134: This code will get messy as we implement more and more models with time offsets, like phase curves. Also, updating params.dict could lead to confusion when users compare inputs vs outputs. Maybe it's better to educate users on not using the full BJD in their initial parameters file?

taylorbell57: Your thinking is that we require the inputs in the parameter ecf be in MBJD (BJD-2400000.5) and then only make any adjustments to the time array in Stage 5? Or even make any time adjustments in Stage 3 only?

kevin218: Thinking about max flexibility, we cold let users define the time offset in the ecf file. If none is given, then use the current definition. We don't want to tie ourselves to choices in S3 that don't impact the reduction. That'll just lead to frustration (speaking from experience). Happy to talk more about this.

taylorbell57: To me, the danger with your suggested method is that everybody needs to be sure to subtract the time_offset from all of the time-based parameters in their parameter ecf, and not doing so will lead to significant issues without any hints as to what could have gone wrong to the user. I'll proceed with your suggested method, but we should think carefully about how we want to protect users from confusion

taylorbell57:
Alright, made those requested changes to the time offset and merged in all of the recent changes to main - any last requests before approval and merging @kevin218?

kevin218 (approved):

kevin218:
Good to go!

PR #101, opened 2022-01-31
Sigma clipping with moving mean replacement to S4

taylorbell57:
A quick little PR to add sigma clipping to S4

The NIRSpec PRISM simulated data had several reads where many/all wavelengths were fully zeroed out which required sigma clipping to remove. For now I have added code to replace clipped data points with the median value from a moving box-car filter, but we can add other options in the future. There is also the ability to replace all outliers with a constant value.

Also now forcing printed log statements to flush to STDOUT as they would somtimes get clogged and not print until the end of the stage when the log was closed.

### DIFF
--- a/eureka/S4_generate_lightcurves/clipping.py
+++ b/eureka/S4_generate_lightcurves/clipping.py
@@ -1,0 +1,92 @@
+import numpy as np
+from astropy.convolution import Box1DKernel, convolve
+from astropy.stats import sigma_clip
+ 
+def clip_outliers(data, log, wavelength, sigma=10, box_width=5, maxiters=5, fill_value='boxcar', verbose=False):
+    '''Find outliers in 1D time series.
+  
+    Be careful when using this function on a time-series with known astrophysical variations. The variable
+    box_width should be set to be significantly smaller than any astrophysical variation timescales otherwise
+    these signals may be clipped.
+   
+    Parameters
+    ----------
+    data: ndarray (1D, float)
+        The input array in which to identify outliers
+    log: logedit.Logedit
+        The open log in which notes from this step can be added.
+    wavelength: float
+        The wavelength currently under consideration.
+    sigma: float
+        The number of sigmas a point must be from the rolling mean to be considered an outlier
+    box_width: int
+        The width of the box-car filter (used to calculated the rolling median) in units of number of data points
+    maxiters: int
+        The number of iterations of sigma clipping that should be performed.
+    fill_value: string or float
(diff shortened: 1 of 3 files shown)


eureka/S4_generate_lightcurves/clipping.py
```diff
@@ -0,0 +1,92 @@
...
+    residuals = data-smoothed_data
+    # Sigma clip residuals to find bad points in data
+    residuals = sigma_clip(residuals, sigma=sigma, maxiters=maxiters)
+    outliers = np.ma.getmaskarray(residuals)
+  
+    if np.any(outliers) and verbose:
+        log.writelog('Identified {} outliers for wavelength {}'.format(np.sum(outliers), wavelength))
+  
+    # Replace clipped data
+    if fill_value=='boxcar':
```
kevin218, line 54: I don't like the idea of replacing outliers with mean values. This can mess with systematic-fitting routines in Stage 5 and lead to poor fits with outliers that normally would have been corrected. The better solution is to flag outliers as bad and leave the data values unchanged.

kevin218 (changes requested):

eureka/S4_generate_lightcurves/clipping.py
taylorbell57: Yeah, initially I wanted to do something like setting the values to NaN (or using numpy masked arrays, but they can use up a lot of RAM), but I wasn't sure how this would impact later stages. I look into this more

kevin218: Masked arrays should be fine at this stage. Using a boolean array of good/bad points also works.

taylorbell57: Alright, I haven't tested that masked arrays from S4 properly get passed to S5, but the new code should work. I'll test it tomorrow

taylorbell57: Alright, masked arrays work well in Stage 4 and 5, so marking as resolved. I kept the replacement code in place in case others want to do something similar in the future, but the code defaults to masking instead now.

taylorbell57: Oops, I was still using the replacement option rather than masking, and masking doesn't seem to work right now. Looking into it more

taylorbell57:
Alright, I've now ensured that masking of bad data in S4 works well and carries over well to S5. In the process I also discovered and resolved a big bug with S4-5 where they would not properly consider all of the different apertures run in S3 despite their claims that they were using those different apertures.

kevin218 (approved):

PR #109, opened 2022-02-05
Origin/dev/adf

afeinstein20:
Merging in the cosmic ray removal routine + 2D background fitting routine for NIRISS. The majority of changes are present under `background.fit_bg3` and `niriss.fit_bg`.

### DIFF
--- a/eureka/S3_data_reduction/background.py
+++ b/eureka/S3_data_reduction/background.py
@@ -425,5 +425,6 @@ def fitbg3(data, order_mask, niters=2, readnoise=5, sigclip=[4,2], isplots=0):
     if isplots or isplots>=5:
         fig,(ax1,ax2,ax3) = plt.subplots(nrows=3, figsize=(14,8))
 
+    data.bkg_removed = bkg_subbed
 
-    return bkg_subbed
+    return data
--- a/eureka/S3_data_reduction/background.py
+++ b/eureka/S3_data_reduction/background.py
@@ -1,7 +1,13 @@
 import numpy as np
-import matplotlib.pyplot as plt
-import multiprocessing as mp
 from tqdm import tqdm
+import ccdproc as ccdp
+from astropy import units
+import multiprocessing as mp
+import matplotlib.pyplot as plt
+from astropy.nddata import CCDData
+from astropy.stats import SigmaClip, sigma_clip
+from photutils import MMMBackground, MedianBackground, Background2D
+
 
 __all__ = ['BGsubtraction', 'fitbg', 'fitbg2', 'fitbg3']
 
@@ -324,7 +330,6 @@ def fitbg2(dataim, meta, mask, bgmask, deg=1, threshold=5, isrotate=False, isplo
                     else:
                         nobadpixels = True      #exit while loop
 
-
                     if isplots == 6:
                         plt.figure(3601)
                         plt.clf()
@@ -351,20 +356,53 @@ def fitbg2(dataim, meta, mask, bgmask, deg=1, threshold=5, isrotate=False, isplo
     return bg, bgmask#, mask #,variance
 
 
+
 def fitbg3(data, isplots=0):
     """
-    Fit sky background with out-of-spectra data. Hopefully this is a faster
-    routine than fitbg2. (optimized to fit across the x-direction)
+    Fit sky background with out-of-spectra data. Optimized to remove
+    the 1/f noise in the NIRISS spectra (works in the y-direction).
 
     Parameters
     ----------
-    dataim : np.ndarray                                          
-       Data image to fit the background to.                                                            
-    isplots : int, optional                                    
-       The amount of plots saved; set in ecf. Default is 0.                                          
+    isplots : bool, optional                                      
+       Plots intermediate steps for the background fitting routine.
+       Default is False.                                          
 
     Returns
     -------
     bg : np.ndarray
        Background model.
     """
+    # Removes cosmic rays
+    # Loops through niters cycles to make sure all pesky
+    #    cosmic rays are trashed
+    rm_crs = np.zeros(data.data.shape)
+    for i in range(len(data.data)):
+        ccd = CCDData((data.data[i])*units.electron)
+
+        for n in range(niters):
+            m1  = ccdp.cosmicray_lacosmic(ccd, readnoise=readnoise, sigclip=sigclip)
+            ccd = CCDData(m1.data*units.electron)
+        rm_crs[i] = m1.data + 0.0
+
+    # Models the background
+    sigma_clip = SigmaClip(sigma=sigclip)
+    if bkgtype.lower() == 'median':
+        bkg = MedianBackground()
+    elif bkgtype.lower() == 'mmmbackground':
+        bkg = MMMBackground()
+    else:
+        print('Background model not implemented.')
+        return
+
+    b = Background2D(data.data, 
+                     box_size=box_size,
+                     filter_size=filter_size,
+                     bkg_estimator=bkg,
+                     sigma_clip=sigma_clip,
+                     fill_value=0.0,
+                     mask=mask)
+
+        
+    if isplots or isplots>=5:
+        fig,(ax1,ax2,ax3) = plt.subplots(nrows=3, figsize=(14,8))
--- a/eureka/S3_data_reduction/background.py
+++ b/eureka/S3_data_reduction/background.py
@@ -357,7 +357,7 @@ def fitbg2(dataim, meta, mask, bgmask, deg=1, threshold=5, isrotate=False, isplo
 
 
 
-def fitbg3(data, isplots=0):
+def fitbg3(data, order_mask, niters=2, readnoise=5, sigclip=[4,2], isplots=0):
     """
     Fit sky background with out-of-spectra data. Optimized to remove
     the 1/f noise in the NIRISS spectra (works in the y-direction).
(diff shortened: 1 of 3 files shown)


afeinstein20:
at some point this may need to be sped up (e.g. moving some components into c) but this now goes from 2D image -> 1D extracted spectra for the first and second orders

eureka/S3_data_reduction/background.py
```diff
@@ -351,20 +360,174 @@ def fitbg2(dataim, meta, mask, bgmask, deg=1, threshold=5, isrotate=False, isplo
...
+                     bkg_estimator='median', sigma=4, box=(10,5), filter_size=(2,2))
+        b2 = bkg_sub(rm_crs[i]-b1, 
+                     order_mask,
+                     sigma=3,
+                     bkg_estimator='median')
+        
+        bkg_subbed[i] = (rm_crs[i]-b1)-b2
+
+        
+    if isplots or isplots>=5:
```
kevin218, line 528: This statement confuses me. Why not just >=5?

eureka/S3_data_reduction/niriss.py
```diff
@@ -459,12 +473,119 @@ def wave_NIRISS(wavefile, meta):
...
+    n_iters : int, optional
+       The number of iterations to go over and remove cosmic
+       rays. Default is 3.
+    readnoise : float, optional
+       An estimation of the readnoise of the detector.
+       Default is 5.
+    sigclip : list, array, optional
+       A list or array of len(n_iiters) corresponding to the
+       sigma-level which should be clipped in the cosmic
+       ray removal routine. Default is [4,2,3].
```
kevin218, line 493: Why is the default 4,2,3? I would expect something like 4,4,4.

eureka/S3_data_reduction/niriss.py
```diff
@@ -459,12 +473,119 @@ def wave_NIRISS(wavefile, meta):
...
     """
     Subtracts background from non-spectral regions.
 
-    # want to create some background mask to pass in to 
-      background.fitbg2
+    Parameters
+    ----------
+    data : object
+    meta : object
+    n_iters : int, optional
```
kevin218, line 484: Can't this be derived from the length of sigClip?

eureka/S3_data_reduction/niriss.py
```diff
@@ -459,12 +473,119 @@ def wave_NIRISS(wavefile, meta):
...
+
+    box_mask = dirty_mask(data.median)
+    data = fitbg3(data, np.array(box_mask-1, dtype=bool), 
+                  n_iters, readnoise, sigclip, isplots)
+    return data
+
+
+def fit_orders(data, meta):
     """
+    Creates a 2D image optimized to fit the data. Currently
```
kevin218, line 530: Can you help me understand what this function is doing?

kevin218 (commented):
General comment: It would be helpful to see more comments in the code to better understand the logic

eureka/S3_data_reduction/background.py
```diff
@@ -351,20 +360,174 @@ def fitbg2(dataim, meta, mask, bgmask, deg=1, threshold=5, isrotate=False, isplo
...
+    """ 
+    A skewed gaussian model.
+    """
+    t = alpha * (x - eta) / omega
+    Psi = 0.5 * (1 + erf(t / np.sqrt(2)))
+    psi = 2.0 / (omega * np.sqrt(2 * np.pi)) * np.exp(- (x-eta)**2 / (2.0 * omega**2))
+    return (psi * Psi)*scale
+
+
+def gauss_removal(img, mask, linspace, where='bkg'):
```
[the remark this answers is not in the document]
afeinstein20: I was debating whether or not along the time axis would be better! Should've gone with my gut. I wonder if it's worthwhile to split all of these routines into their own `cosmic_ray_removal.py` file. I think there was interest in optimizing CR removal from the Google doc Monday, so it would be cool if people were able to implement their own routines there. What do you think?

eureka/S3_data_reduction/niriss.py
afeinstein20: yeah, good point!

afeinstein20: I fixed this in the recent push

eureka/S3_data_reduction/background.py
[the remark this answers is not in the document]
kevin218: Yes, 100% agree.
Also, check out clipping.py in the S3 folder.

eureka/S3_data_reduction/niriss.py
afeinstein20: So basically what it's doing is similar to what I had slack'd you about (e.g. profile fitting for optimal extraction). So what you get out of this function are 2 images: (1) has a weighted mask for the first order and (2) has a weighted mask for the second order. Each of the profiles are now are being approximated with a Gaussian of varying amplitudes.

I hope this image will help explain what the final output of the code is.
Left top: example NIRISS exposure
Left bottom: the 2 fitted masks co-added
Right column: the differences in profile shapes at different rows, where the orange is the data profile, and the black is the profile from the fitted mask 

niriss_2d_mask

The reason why it takes so long is because it's fitting the amplitudes and the standard deviations of the gaussian models to the data (I also want it to be flexible enough to fit the center value, but I need to make it faster before adding more variables. Also, this way the order tracing done before is treated as an initial guess, rather than ground truth. But I digress....)

eureka/S3_data_reduction/background.py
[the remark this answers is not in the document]
afeinstein20: Sorry, I don't see a `clipping.py`? Was it possibly looped into `optspex.py`?

eureka/S3_data_reduction/niriss.py
afeinstein20: Also, here's a comparison between spectra extracted (after all processing) with a box (blue) vs. with the weighted mask derived from this function (black)

order_comparison

eureka/S3_data_reduction/background.py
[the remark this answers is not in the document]
kevin218: See Taylor's current PR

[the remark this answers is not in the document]
kevin218: Sorry, S4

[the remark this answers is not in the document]
afeinstein20: Ooo yeah I see now. I think it makes sense to have the bulk of cosmic ray removal in S3. So maybe we should think about where this step makes the most sense for all instruments, and keep `clipping.py` for nirspec as an additional check for S4

eureka/S3_data_reduction/niriss.py
kevin218: Thanks for the extra info.  I'll take a second look at the code this morning to piece it all together.
You should check out `fitgaussians()` in [gaussian.py](https://github.com/kevin218/Eureka/blob/main/eureka/lib/gaussian.py).  It may prove to be faster than you current implementation.  
When it comes to fitting the center, it's often good to use the previous best fit center as your starting point for the next columns; however, using the trace would probably be just as good, so I leave that up to you.

eureka/S3_data_reduction/background.py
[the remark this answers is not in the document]
kevin218: If `clipping.py` is generic enough, we could move it to the `lib` directory and use it for multiple stages.  It shouldn't matter if we're looking at the white LC flux or flux along a single pixel in time, the code effectively does the same job.  What do you think?

eureka/S3_data_reduction/niriss.py
kevin218: Alternatively, if you want to focus your time on other routines that yield more realistic profiles, we can simply leave this one as is for now.

eureka/S3_data_reduction/background.py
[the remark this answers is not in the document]
afeinstein20: I like this! I think it makes a lot of sense to move it to `lib` since it is pretty generic. I will wait until Taylor's PR is accepted before making any changes, to avoid potential merge issues

eureka/S3_data_reduction/niriss.py
afeinstein20: Ah, I see! I think there are definitely good lessons in `fitgaussian()` to be implemented here and try to wrap them into my implementation in the next PR, with a better profile than a gaussian

eureka/S3_data_reduction/background.py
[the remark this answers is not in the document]
kevin218: PR merged!

kevin218 (approved):

eureka/S3_data_reduction/background.py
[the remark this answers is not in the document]
afeinstein20: Moved `clipping.py` into `lib` and updated the import statements calling that file! And put my cosmic ray removal function in there

eureka/S3_data_reduction/background.py
```diff
@@ -351,20 +360,174 @@ def fitbg2(dataim, meta, mask, bgmask, deg=1, threshold=5, isrotate=False, isplo
...
+       Sets where the mask is covering. Default is `bkg`. 
+       Other option is `order`.
+
+    Returns
+    -------
+    img : np.ndarray
+       The same input image, now masked for newly identified
+       outliers.
+    """
+    n, bins, patches = plt.hist((img*mask).flatten(),
```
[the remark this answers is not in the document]
afeinstein20: Yeah, I think you're right! I've updated this

eureka/S3_data_reduction/niriss.py
afeinstein20: Yeah, I've updated this!

eureka/S3_data_reduction/niriss.py
afeinstein20: So I managed to throw together a routine with `scipy.optimize.leastsqr` that's certainly faster and mildly different from the function in `gaussian.py`. The results between that best-fit model and mine seems, minimal (see below)

The lefthand column has 2 masks: the top is with my optimization, the bottom is with `scipy.optimize`. The top mask extracted the blue spectra and the bottom mask extracted the black spectra (to the right)

order_comparison

What do you think?

kevin218: Is there any way to know the truth? The significant difference in flux between the two methods is worrisome.

afeinstein20: I've asked Joe for the ground truth of what was injected in the simulation. When he gets those to me, that will probably instill some confidence in one method over the other

afeinstein20: I'm worried this `so.leastsqr` method is unreliable, since I reran everything this morning and am now getting different results.... I think the current version I have pushed is better for now (even if it's slower)

afeinstein20: I've pushed a version now with both the old method (`fit_orders`) and a faster version with scipy.optimize (`fit_orders_fast`). I think this is the best I can do right now before I get the ground truth results to compare to.

kevin218: Agreed!

kevin218 (approved):

PR #120, opened 2022-02-17
Resolving niriss_cython import issues

taylorbell57:
Needed to add some special code to the setup.py file to make sure that
the pyx file gets compiled and installed into the eureka site-package

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -138,3 +138,6 @@ dmypy.json
 
 # Ignore installed packages for wfc3 branch
 image_registration/
+
+# Compiled niriss_cython file
+niriss_cython.c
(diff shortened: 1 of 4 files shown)


taylorbell57:
I've tested that this fixes the issues I had with importing anything from Eureka, but I haven't tested that this still works for NIRISS reduction. Can you please test this @afeinstein20?

kevin218 (approved):

afeinstein20:
`pip install .` worked just fine for me in a clean conda environment for both python==3.9.7 and python==3.10. Nothing seems to be broken either. I'd say go ahead and merge @taylorbell57 !

PR #122, opened 2022-02-17
Stage 6 First Attempt

taylorbell57:
I made a first attempt at a Stage 6 which plots the emission spectrum if fp is a fitted parameter, otherwise it plots the transmission spectrum. This will work for now, but once we can fit phase curves we'll want to plot both the transmission and emission spectra (as well as the phase-resolved emission spectrum). Given the current state of Stage 5 though, I felt like this was good enough and worth a PR now. I also added the ability to plot a model spectrum to compare with inputs to simulated data or fitted spectra. Attached is the output from a 20-bin fit to the old NIRCam observations

fig6101


eureka/S6_planet_spectra/s6_spectra.py
```diff
@@ -0,0 +1,281 @@
...
+
+class MetaClass:
+    '''A class to hold Eureka! metadata.
+    '''
+
+    def __init__(self):
+        return
+
+def plot_spectra(eventlabel, s5_meta=None):
+    '''Gathers together diferent wavelength fits and makes transmission/emission spectra.
```
kevin218, line 23: Different

eureka/S6_planet_spectra/s6_spectra.py
```diff
@@ -0,0 +1,281 @@
...
+    if (not meta.s5_allapers) or (not meta.allapers):
+        # The user indicated in the ecf that they only want to consider one aperture
+        # in which case the code will consider only the one which made s4_meta.
+        # Alternatively, S4 was run without allapers, so S6's allapers will only conside that one
+        meta.spec_hw_range = [meta.spec_hw,]
+        meta.bg_hw_range = [meta.bg_hw,]
+
+    if meta.testing_S6:
+        # Only fit a single channel while testing
+        chanrng = [0]
```
kevin218, line 70: This variable doesn't appear to be used, but may to keep this if statement for later.

eureka/S6_planet_spectra/s6_spectra.py
```diff
@@ -0,0 +1,281 @@
...
+            
+            t0 = time_pkg.time()
+            
+            meta = load_specific_s5_meta_info(old_meta, run_i, spec_hw_val, bg_hw_val)
+            
+            # Get the directory for Stage 6 processing outputs
+            meta.outputdir = util.pathdirectory(meta, 'S6', meta.runs_s6[run_i], ap=spec_hw_val, bg=bg_hw_val)
+            run_i += 1
+            
+            # Copy existing S4 log file and resume log
```
kevin218, line 94: S5

eureka/S6_planet_spectra/plots_s6.py
```diff
@@ -0,0 +1,29 @@
+import numpy as np
+import matplotlib.pyplot as plt
+
+def plot_spectrum(meta, wavelength, spectrum, err, wavelength_error=None, model_x=None, model_y=None, ylabel=r'$R_{\rm p}/R_{\rm *}$', xlabel=r'Wavelength ($\mu$m)'):
```
kevin218, line 4: A nice option to have is to include scale heights on the right axis. Perhaps we make this a separate figure with isplots=3?

kevin218 (changes requested):

eureka/S6_planet_spectra/s6_spectra.py
```diff
@@ -0,0 +1,281 @@
...
+            meta.s6_logname  = meta.outputdir + 'S6_' + meta.eventlabel + ".log"
+            log         = logedit.Logedit(meta.s6_logname, read=meta.s5_logname)
+            log.writelog(f"Input directory: {meta.inputdir}")
+            log.writelog(f"Output directory: {meta.outputdir}")
+            
+            # Copy ecf (and update outputdir in case S6 is being called sequentially with S5)
```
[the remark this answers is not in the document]
taylorbell57: Excellent idea!

eureka/S6_planet_spectra/plots_s6.py
taylorbell57: Good thinking - I'll reuse this function and just call it twice if isplots==3 with a boolean flipped

eureka/S6_planet_spectra/s6_spectra.py
taylorbell57: S5 was indeed not using chanrng, but it now is

eureka/S6_planet_spectra/s6_spectra.py
```diff
@@ -0,0 +1,281 @@
...
+            fname = f'S5_{fitter}_samples_{channel_key}.csv'
+            samples = pd.read_csv(meta.inputdir+fname, escapechar='#', skipinitialspace=True)
+            # FINDME: kludge to decide between transmission and emission spectra for now
+            if 'fp' in samples.keys():
+                keys = [key for key in samples.keys() if 'fp' in key]
+                ylabel=r'$F_{\rm p}/F_{\rm *}$ (ppm)'
+                scalar = 1e6
+            else:
+                keys = [key for key in samples.keys() if 'rp' in key]
+                ylabel=r'$R_{\rm p}/R_{\rm *}$'
```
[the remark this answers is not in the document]
taylorbell57: I guess setting what units to plot in would be a cleaner way of choosing between fp/fs and rp/rs than what I have too. If it can take a list, then it will even kind of work for phase curves (or at least be closer to working)

eureka/S6_planet_spectra/plots_s6.py
kevin218: But only for transits!

taylorbell57: How do you want to convert between radii and H? Do you want people to pass in the scale height, pass in some of the values and look up the rest from the exoplanet archive?

taylorbell57:
Alrigh, I've addressed the above and done my best to support as much flexibility as possible (you can even plot your x-axis in Hz if you should desire to do so for some very strange reason).

Attached is the additional figure that can be made with isplots>=3, and I've used transit depth as the y-axis and also plotted it in percent (y_scalar=10; if y_scalar==1e6, it'll say ppm, and for any other number it'll say `f'.... * {y_scalar}'`). I set the zero-point of the scale-height axis to be the mean radius, but can do something else if you'd prefer

fig6301

taylorbell57:
I should also mention I merged @cpiaulet's PR #131 into mine to test it and just left it there as I saw no issues. I should've undone that merge to make things cleaner, but I forgot.

kevin218:
> I set the zero-point of the scale-height axis to be the mean radius, but can do something else if you'd prefer

That works well as a default. Could have it as an optional parameter in case someone wants something different.

kevin218 (commented):
Nice work on these changes!

demos/JWST/S6_template.ecf
```diff
@@ -0,0 +1,34 @@
...
+x_unit			um # Options include any measurement of light included in astropy.units.spectral (e.g. um, nm, Hz, etc.)
+
+# This section is relevant if isplots_S6>=3
+# Scale height parameters (if you want a second copy of the plot with a second y-axis with units of scale height)
+star_Rad		0.6506 # The radius of the star in units of solar radii
+planet_Teq		1400 # The equilibrium temperature of the planet in units of Kelvin
+planet_Mass		2 # The planet's mass in units of Jupiter radii (used to calculate surface gravity)
+planet_Rad		None # The planet's radius in units of Jupiter radii (used to calculate surface gravity); Set to None to use the average fitted radius
+planet_mu		2.3 # The mean molecular mass of the atmosphere (in atomic mass units)
+
```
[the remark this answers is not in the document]
taylorbell57: Can you clarify what you mean here?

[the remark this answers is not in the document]
kevin218: Add H_0 here so that people can define their own zero point for scale height. Can leave line commented out to apply mean as offset.

kevin218 (approved):

Issue #129, opened 2022-02-18
AttributeError: 'PolynomialModel' object has no attribute 'longparamlist'


jbrande:
On a clean install of Eureka (48c0443), when running S5 with the polynomial model specified, the fitting steps fail because the PolynomialModel is missing the attribute 'longparamlist'. This attribute is present in TransitModel, but didn't make it over to the PolynomialModel yet.

taylorbell57:
Can you explain more where this error is coming up? I can see that longparamlist isn't in the PolynomialModel, but I don't know why it would need to be or what would throw an error

taylorbell57:
Never mind, I see how it could be an issue and it was an easy fix

Issue #128, opened 2022-02-18
transit-only models break plot_fit


jbrande:
On a clean install of Eureka (48c0443), when running S5 with run_myfuncs set to [transit], plot_fit() in plots_s5.py breaks. The code looks to use a variable called model_sys to plot the detrended lightcurve, and on line 54 selects a subarray of this model. However, if there's no systematics model specified in the ECF, this model_sys is given a default OOT flux of 1.0 in model.syseval(), which pulls the systematics array out of the model object. The default return value here should probably be an ndarray of the same length as the model, with all the values set to 1.0, rather than just a single value.

PR #130, opened 2022-02-18
Resolving Issues #128 and #129

taylorbell57:
A quick little bug patch PR to resolve Yoni's 2 discovered issues #128 and #129


jbrande:
I just tested this and it looks like both issues are fixed for now. @kevin218 , if there's nothing else you need from this, we can probably go ahead and merge this back in.

kevin218 (approved):

PR #131, opened 2022-02-18
add lc scatter as fit param + new residuals distribution plot + minor plot fixes

cpiaulet:
- new fitting parameter: scatter_ppm
- parameter used for chi squared and log likelihood calculations
- make code robust to hitting upper or lower bound in initial lsq
- add a new plot: distribution of residuals normalized by the single-point scatter in the light curve
- fix xlabel for lc plot
- fix uncertainties in residuals panel following lc fit

### DIFF
--- a/demos/Run_Eureka.ipynb
+++ b/demos/Run_Eureka.ipynb
@@ -230,7 +230,7 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "Python 3",
+   "display_name": "Python 3 (ipykernel)",
    "language": "python",
    "name": "python3"
   },
@@ -244,7 +244,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.7.6"
+   "version": "3.9.7"
   }
  },
  "nbformat": 4,
--- a/eureka/S5_lightcurve_fitting/fitters.py
+++ b/eureka/S5_lightcurve_fitting/fitters.py
@@ -184,6 +184,9 @@ def emceefitter(lc, model, meta, log, **kwargs):
     
     if lsq_sol.cov_mat is not None:
         step_size = np.diag(lsq_sol.cov_mat)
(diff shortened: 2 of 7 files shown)


taylorbell57:
Have you tested whether the new scatter_ppm parameter works well with shared fits? I've yet to do any testing with the new parameter, but I can confirm that the code behaves as before if you don't choose to fit for it

taylorbell57 (changes requested):
As of right now this breaks for shared fits, but I have confirmed it works well for unshared fits. Let me know if you would like help getting it to work

cpiaulet:
I have never ran shared fits with Eureka, would love some help with at least testing so I understand how the shared fits are coded and where it breaks :)

taylorbell57:
You could run S4 with more than one wavelength bin (can just do two to make things quick to test) and then set any parameter in the S5 param ecf to 'shared' instead of 'free' (e.g. t0 is a good one since the transit timing should be wavelength independent). I tried it once and got an array shape error when evaluating the new model and it wasn't immediately clear to me why model.eval() would give the wrong shape so I just left a review here for now since I have to referee a paper and write a reference letter before the end of the week. Hopefully it'll be an easy fix for you, otherwise I (in a day or few) or someone else can help if you get stuck

taylorbell57:
Then again I just tried a shared fit without scatter_ppm and I get the same error, so maybe it is caused by my changes rather than yours or yours breaks regardless of whether you fit scatter_ppm in a shared fit

taylorbell57:
Well my issue may or may not be related to your PR, but I can already tell it won't work with shared fits as the different scatter_ppm parameters that get made for each wavelength don't get applied to each wavelength's lightcurve. Instead the first one gets used for all wavelengths and all the others are only constrained by the priors and are unused.

taylorbell57:
@cpiaulet this PR ended up getting merged in with mine which is fair enough since your edits don't really break the code for shared fits any more than it already was. Have you made any progress on getting the scatter_ppm to work with shared fits?

PR #132, opened 2022-02-18
Adding a .ecf file to the media folder

AarynnCarter:
In my commit for the Stage 1 documentation I forgot to add this .ecf file which is used as a preview on the webpages.


taylorbell57:
Sorry, I hope you don't mind but while you were at it I added in updated ecfs for S2-5 as well since they were out of date. I haven't updated the documentation about those ecfs yet, but at least the ecfs will be up-to-date

taylorbell57 (approved):
Looks good to me, but Kevin might want to read through it as well before we merge

kevin218 (approved):
Looks good!

Issue #123, opened 2022-02-17
Bug in S1 when ramp_fit_max_cores is all


taylorbell57:
If you set ramp_fit_max_cores to all in the ecf file, the readecf function interprets it as the python built-in `all` function rather than the string all. Perhaps quotes are needed? Using half works fine without quotes, but I haven't tested it with quotes

PR #137, opened 2022-02-22
Resolving Issue #123

taylorbell57:
Without quotation marks, all gets interpreted as the built-in function all. Edited ecf templates to encourage people to use quotation marks. Closes Issue #123


AarynnCarter (approved):
Looks good!

PR #146, opened 2022-02-25
Fixing error message for S3 when no files found

taylorbell57:
The error message in S3 when no S2 files were found is out-of-date, and this tiny PR fixes that issue


taylorbell57:
Threw in another bug patch that fixes a typo that prohibits plotting emission spectra

kevin218 (approved):

Issue #148, opened 2022-02-26
Need a model for phase variations


taylorbell57:
We can currently only fit for a transit **or** an eclipse, and also cannot fit for phase variations. This needs to be addressed to handle the new MIRI/LRS phase curve simulations in the ERS data challenge. I'll open a PR that addresses this issue shortly

PR #149, opened 2022-02-26
Adding phase curve model

taylorbell57:
I've now added the ability to fit both a transit and eclipse model as well as phase variations, addressing Issue #148. I've adopted the phase variation model used by SPCA (with parameters A, B, C, and D). I also made it easier to fit just an eclipse and now you don't need to specify a form of limb darkening. I'd wait until @jbrande's PR is merged before approving this one so that I can update the S5 tests as needed, but I've already confirmed that this PR does not introduce any more failures than are already present in the main branch.


taylorbell57:
One issue I've just realized is that I haven't yet added any physicality priors, so the phasecurve can currently be "negative" (phase variations dip below the bottom of the eclipse)

taylorbell57:
I think a cleaner option moving forwards might be to have something like eclipse_batman for our astrophysical model names so that we could have different method options if others wanted.

demos/JWST/S5_template.ecf
```diff
@@ -30,6 +30,7 @@ interp			False	# Should astrophysical model be interpolated (useful for uneven s
 # Diagnostics
 isplots_S5      5 # Generate few (1), some (3), or many (5) figures (Options: 1 - 5)
 testing_S5      False # Boolean, set True to only use the first spectral channel
+testing_model   True # Boolean, set True to only inject a model source of systematics
```
kevin218, line 33: Do we want this set to `False` for the template?

docs/media/S5_template.ecf
```diff
@@ -30,6 +30,7 @@ interp			False	# Should astrophysical model be interpolated (useful for uneven s
 # Diagnostics
 isplots_S5      5 # Generate few (1), some (3), or many (5) figures (Options: 1 - 5)
 testing_S5      False # Boolean, set True to only use the first spectral channel
+testing_model   True # Boolean, set True to only inject a model source of systematics
```
kevin218, line 33: Same question here

demos/JWST/S5_template.ecf
```diff
@@ -5,7 +5,7 @@ rescale_err     False    # Rescale uncertainties to have reduced chi-squared of
 fit_par         ./S5_fit_par_template.ecf # What fitting ecf do you want to use?
 run_verbose     True
 fit_method      [dynesty] #options are: lsq, emcee, dynesty (can list multiple types separated by commas)
-run_myfuncs     [transit,polynomial] #options are: transit, expramp, and polynomial (can list multiple types separated by commas)
+run_myfuncs     [transit,polynomial] #options are: transit, eclipse, phase, expramp, and polynomial (can list multiple types separated by commas)
```
kevin218, line 8: I'd rather use `phasecurve` since `phase` could be confused with orbital phase.  
This actually begs the deeper question: do we want to reference BATMAN in the transit, eclipse, and PC model names (e.g., batman_tr, batman_ecl, and batman_pc)?  In the near future, I foresee other transit model implementations that apply for sophisticated parameterizations to account for differences between ingress and egress.

eureka/S5_lightcurve_fitting/models/BatmanModels.py
```diff
@@ -0,0 +1,210 @@
...
+                bm_params.t_secondary = m_transit.get_t_secondary(bm_params)
+            
+            # Make the eclipse model
+            m_eclipse = batman.TransitModel(bm_params, self.time, transittype='secondary')
+
+            lcfinal = np.append(lcfinal, m_eclipse.light_curve(bm_params))
+
+        return lcfinal
+
+class PhaseVariationsModel(Model):
```
kevin218, line 133: `PhaseCurveModel` makes more sense to me

eureka/S5_lightcurve_fitting/models/BatmanModels.py
```diff
@@ -0,0 +1,210 @@
...
+                    m_transit = batman.TransitModel(bm_params, self.time, transittype='primary')
+                anom = m_transit.get_true_anomaly()
+                w = self.parameters.dict['w'][0]
+                #the planet is on an eccentric orbit
+                phi  = anom + w*np.pi/180. + np.pi/2.
+            
+            #calculate the phase variations
+            if pc_params['C']==0. and pc_params['D']==0.:
+                #Skip multiplying by a bunch of zeros to speed up fitting
+                phaseVars = 1. + pc_params['A']*(np.cos(phi)-1.) + pc_params['B']*np.sin(phi)
```
kevin218, line 204: Why do you subtract 1 from cos(phi)?  I'm not used to this format, so I might just be missing something.

eureka/S5_lightcurve_fitting/models/BatmanModels.py
```diff
@@ -0,0 +1,210 @@
...
+            else:
+                phaseVars = 1. + pc_params['A']*(np.cos(phi)-1.) + pc_params['B']*np.sin(phi) + pc_params['C']*(np.cos(2.*phi)-1.) + pc_params['D']*np.sin(2.*phi)
```
kevin218, line 206: From what I can tell, this implementation doesn't remove the curvature during secondary eclipse.  While this is a good start, we'll want to implement more advanced versions down the road.

kevin218:
> For reasons like Issue #150, I think a cleaner option moving forwards might be to have something like eclipse_batman for our astrophysical model names so that we could have different method options if others wanted.

Just seeing this message now. I made a similar comment in the review.

eureka/S5_lightcurve_fitting/models/BatmanModels.py
taylorbell57: By this you mean that it doesn't model eclipse mapping type observations? That is definitely true and a limitation of this quick and simple first attempt

eureka/S5_lightcurve_fitting/models/BatmanModels.py
taylorbell57: With this formulation you get just fp as the eclipse depth, so there's no degeneracy with A or B since both terms are now zero for phi=0 during mid-eclipse

eureka/S5_lightcurve_fitting/models/BatmanModels.py
taylorbell57: My hesitation with doing so (and for using phase above instead of phasecurve) was that to me phase curve implies the entirety of the phase curve (transit and eclipse included) while phase variations implies just the sinusoidal modulations to me. I can still change this and/or the above if you prefer though

eureka/S5_lightcurve_fitting/models/BatmanModels.py
taylorbell57: Or are you meaning that there would be phase curvature during the bottom of eclipse? I'll need to double check that, but I'm pretty sure the eclipse function goes to zero during the bottom of eclipse so that there doesn't end up being any curvature during the flat part of the eclipse. That's what SPCA does, but I haven't checked to make sure I carried that part over

taylorbell57: Oops, it does not - working on that now

taylorbell57:
Alright, I tried to implement all of your requested edits and have confirmed that the code passes pytest. To remove the curvature during eclipse I had to reformulate things so that the transit and eclipse models were components of the phase curve model, but I did this is a way that should be independent of the transit and eclipse model used (aka I don't assume a batman transit/eclipse model anywhere). And I don't require that one fits for a transit or eclipse model either, but if either are fitted, then they become a component of the phase curve model

demos/HST/S5_wfc3_template.ecf
```diff
@@ -2,33 +2,35 @@
 
 allapers        True # Run S5 on all of the apertures considered in S4? Otherwise will use newest output in the inputdir
 rescale_err     False    # Rescale uncertainties to have reduced chi-squared of unity
-fit_par         ./s5_fit_par_wfc3_template.ecf # What fitting ecf do you want to use?
+fit_par         ./S5_fit_par_wfc3_template.ecf # What fitting ecf do you want to use?
 run_verbose     True
 fit_method      [dynesty] #options are: lsq, emcee, dynesty (can list multiple types separated by commas)
-run_myfuncs     [transit,polynomial] #options are: transit, expramp, and polynomial (can list multiple types separated by commas)
+run_myfuncs     [transit,polynomial] #options are: transit, eclipse, phasecurve, expramp, and polynomial (can list multiple types separated by commas)
```
kevin218, line 8: Should this be Batman_tr, etc?

eureka/S5_lightcurve_fitting/models/SinusoidPhaseCurve.py
```diff
@@ -0,0 +1,122 @@
...
+
+    def eval(self, **kwargs):
+        """Evaluate the function with the given values"""
+        # Get the time
+        if self.time is None:
+            self.time = kwargs.get('time')
+        
+       #Initialize model
+        bm_params = batman.TransitParams()
+        pc_params = {'A':0., 'B':0., 'C':0., 'D':0.}
```
kevin218, line 73: I still think we need better variable names for the PC model. These letters don't convey "PC amplitude" when I see them.

kevin218 (commented):
Nice implementation of the PC model.

eureka/S5_lightcurve_fitting/models/SinusoidPhaseCurve.py
kevin218: AmpCos1, AmpSin1, AmpCos2, AmpSin2?

taylorbell57: Sure, that works!

kevin218 (approved):

PR #152, opened 2022-02-28
Issue 145

### LINKED ISSUE #145: Absolute ECF pathing
Currently when running a Eureka Stage, you pass in an eventlabel which tells Eureka what files to look for in the current working directory. It would be nice to be able to give an absolute path to the ECF storage folder, to make it easier to compartmentalize different runs of Eureka. 

For example, the current unit testing scheme has unit tests and ECFs all in the same folder, e.g.
tests/
|___SX_NIRCam.ecf
|___SX_NIRSpec.ecf
|___test_general.py (which has test cases for both NIRCam and NIRSpec)

To have separate subfolders to keep track of the various instrument specific ECFs, you'd have something like:
tests/
|___NIRCam/
|     |_____SX_NIRCam.ecf
|___NIRSpec/
|     |_____SX_NIRSpec.ecf
|___test_general.py (which has test cases for both NIRCam and NIRSpec, and absolute paths to their ECFs in the subfolders)

For these specific testing case, there may be other ways to split off the instrument specific tests and keep them in the subfolders as well. Still the general capability to specify paths to ECFs not in the current working directory would be useful.

taylorbell57:
Alright, I made those edits I was talking about that Yoni had requested, and this PR will close #145. Now you can specify a path to the ecfs in the run_eureka.py file (or equivalent), and if you don't it defaults to just './' as it already does. I've tested it with pytest now and found no issue.


jbrande (approved):
This passes pytest locally for me and I don't see anything that gives me red flags in the commits here. Looks good! Thanks Taylor.

kevin218 (approved):

PR #157, opened 2022-03-02
New 2D light curve with drift/jitter correction

kevin218:
Added figure 4102, which is nearly identical to 3101, but includes the benefits of drift/jitter correction.  The plot also draws vertical dashed lines to indicate breaks between spectroscopic channels.

fig4102-2D_LC

To get the drift/jitter correction working, I set the weights to unity for now.  The previous mask was the wrong size.

### DIFF
--- a/eureka/S4_generate_lightcurves/plots_s4.py
+++ b/eureka/S4_generate_lightcurves/plots_s4.py
@@ -91,23 +91,17 @@ def lc_driftcorr(meta, wave_1d, optspec):
     wmin = wave_1d.min()
     wmax = wave_1d.max()
     n_int, nx = optspec.shape
-    # iwmin       = np.where(ev.wave[src_ypos]>wmin)[0][0]
-    # iwmax       = np.where(ev.wave[src_ypos]>wmax)[0][0]
     vmin = 0.97
     vmax = 1.03
-    # normspec    = np.mean(ev.optspec,axis=1)/np.mean(ev.optspec[ev.inormspec[0]:ev.inormspec[1]],axis=(0,1))
     normspec = optspec / np.mean(optspec, axis=0)
     plt.imshow(normspec, origin='lower', aspect='auto', extent=[wmin, wmax, 0, n_int], vmin=vmin, vmax=vmax,
                cmap=plt.cm.RdYlBu_r)
     ediff = np.zeros(n_int)
     for m in range(n_int):
         ediff[m] = 1e6 * np.median(np.abs(np.ediff1d(normspec[m])))
-        # plt.scatter(ev.wave[src_ypos], np.zeros(nx)+m, c=normspec[m],
-        #             s=14,linewidths=0,vmin=vmin,vmax=vmax,marker='s',cmap=plt.cm.RdYlBu_r)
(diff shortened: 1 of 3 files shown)


eureka/S4_generate_lightcurves/plots_s4.py
```diff
@@ -70,9 +70,60 @@ def drift1d(meta):
...
+    Returns
+    -------
+    None
+    '''
+    plt.figure(4102, figsize=(8, 8))  # ev.n_files/20.+0.8))
+    plt.clf()
+    wmin = wave_1d.min()
+    wmax = wave_1d.max()
+    n_int, nx = optspec.shape
+    # iwmin       = np.where(ev.wave[src_ypos]>wmin)[0][0]
```
taylorbell57, line 94: Can we clean out this old commented out code?

eureka/S4_generate_lightcurves/s4_genLC.py
```diff
@@ -165,14 +165,16 @@ def lcJWST(eventlabel, s3_meta=None):
                 # Correct for drift/jitter
                 for n in range(meta.n_int):
                     # Need to zero-out the weights of masked data
-                    weights = (~optspec[n].mask).astype(int)
+                    #weights = (~optspec[n].mask).astype(int)
```
taylorbell57, line 168: Replace this with the following:
`weights = (~np.ma.getmaskarray(optspec[n])).astype(int)`

eureka/S4_generate_lightcurves/plots_s4.py
```diff
@@ -70,9 +70,60 @@ def drift1d(meta):
...
+    ediff = np.zeros(n_int)
+    for m in range(n_int):
+        ediff[m] = 1e6 * np.median(np.abs(np.ediff1d(normspec[m])))
+        # plt.scatter(ev.wave[src_ypos], np.zeros(nx)+m, c=normspec[m],
+        #             s=14,linewidths=0,vmin=vmin,vmax=vmax,marker='s',cmap=plt.cm.RdYlBu_r)
+    plt.title("MAD = " + str(np.round(np.mean(ediff), 0).astype(int)) + " ppm")
+    # plt.xlim(wmin,wmax)
+    # plt.ylim(0,n_int)
+    if meta.nspecchan > 1:
+        xticks = np.unique(np.concatenate([meta.wave_low,meta.wave_hi]))
```
taylorbell57, line 111: This doesn't look very good if you have many wavelength bins (ignore the fact that some of the wavelength bins are outside of the spectrum - I messed up the wavelength limits in my ecf). Perhaps you can angle the x-ticks by 45 degrees or so to help avoid them getting too crowded?

fig4102-2D_LC

taylorbell57 (changes requested):

eureka/S4_generate_lightcurves/plots_s4.py
kevin218: I picked 90 degrees to allow for maximum flexibility.  For my test case, there was no overlap in text with 40 channels.

taylorbell57: We could instead just change the dimensions of the figure rather than swapping the axes if we get to that point

taylorbell57 (approved):

Issue #116, opened 2022-02-15
Use reparameterized limb darkening coefficients


taylorbell57:
As of right now we are directly fitting for the quadratic limb darkening coefficients, but this is not very efficient and is hard to place priors on. We'd be better off using the reparameterized versions (q<sub>1</sub> = (u<sub>1</sub> + u<sub>2</sub>)<sup>2</sup> and q<sub>2</sub> = 0.5*u<sub>1</sub>(u<sub>1</sub> + u<sub>2</sub>)<sup>−1</sup> with uniform priors in the interval [0, 1]) à la [Kipping (2013)](https://academic.oup.com/mnras/article/435/3/2152/1024138).

PR #160, opened 2022-03-02
Edits to get Eureka to work with new MIRI simulations

taylorbell57:
The new MIRI simulations also do not have time stamps, so I added some code to automatically generate those. I also added some useful error messages in astropytable.py which previously would throw a very unhelpful error message if the time array (or any other array) had the wrong shape.


taylorbell57:
Ahh, it seems Caroline's commit statements still made it into my git log despite my attempts to remove them - her edits have not been included in this PR despite what the log suggests

taylorbell57:
Alright, while I was at it I added in the reparameterized quadratic limb darkening which closes #116. I also added code to close #161. And I added a new set of figures to show the evolution of the emcee walkers to be able to visually look for strange behaviour. Finally, I moved the code that saves the fits to as early in the code as I could to make sure any bugs in the plotting code do not mean a total loss of the fit.

kevin218 (approved):

PR #164, opened 2022-03-03
Minor enhancements to Stages 5 + 6

kevin218:
@taylorbell57 In case you're wondering what's up with `plots_s6.py`, I changed all the tabs to spaces for code indentation.  I also noticed that many other files used tabs.  When we first started, everything was supposed to use 4 spaces for indentation.  We need to figure out a consistent solution.


taylorbell57 (approved):
Everything looks fine to me. I'm totally fine with using spaces or tabs but often prefer tabs since it makes aligning columns in ecfs far easier to maintain. I've just been using the default behaviour of VS Code around tabs/spaces, and I think it uses spaces if a file already has spaces (and automatically uses the same number of spaces - e.g., some old files already had 2 spaces as their indent, so that's what it used). If I make a new file though (like all the S6 files), then I think it defaults to a tab

Issue #154, opened 2022-03-01
Have analyses print MAD value to terminal


iancrossfield:
When optimizing the analysis parameters, it can be tedious to individually open the output figures for each analysis session to see how things worked out. To slightly streamline things: why not just print the MAD value to the terminal as each run concludes? E.g., something like:
> MAD for this run: 1234 ppm
> Total time (min): 0.04
> Saving results as astropy table
> Generating figure
> Saving Metadata

kevin218:
While we're at it, we should also make the MAD calculation robust against NaNs. I hope this will address #155.

Issue #155, opened 2022-03-02
MAD Calculation Fails for some S3 Runs


jbrande:
For some not quite predictable reasons, the MAD calculation for S3 will fail and give some obviously incorrect, massively negative value.

Compare these two S3 plots:
![nirspec_mad_reasonable](https://user-images.githubusercontent.com/5041139/156267852-4cc89b3e-61f0-40e5-a7d9-68a0f6470066.jpg)
![nirspec_mad](https://user-images.githubusercontent.com/5041139/156267862-8eaafc9d-f059-4dfa-96d5-a00669d0314c.jpg)

This doesn't seem to happen with the S4 MAD calculation that I've noticed. I can tell that the calculation in plots_s3.py and plots_s4.py are slightly different, but not whether this could be the cause of the issue.

taylorbell57:
Well apparently this issue also happens sometimes in Stage 4 (see the plot below made using the code from tjb_current, aka the code as of commit 209e41fd1d5b44829d74b9dda22072df77e22ffa). Looking at the figure, it seems there's a line of NaNs (or more likely mased values) at very low integration numbers - these would have been clipped in Stage 4 as they were strong outliers. I guess we need to make sure we're using np.nanmedian or np.ma.median and functions like those in Stage 4 as well (@kevin218 do you do this in PR #199?).

![fig4102-2D_LC](https://user-images.githubusercontent.com/24514122/159540341-228111cd-776e-45ff-bc26-99676389164c.png)

kevin218:
I did not use masked arrays, but we should.  After some discussions, I think the right path forward is to breakout the MAD calculation into a separate function in the lib directory.  The Stage 3/4 codes can then call the same subroutine.  Note, the results can be different between Stage 3/4 if the wavelength edges are provided.

taylorbell57:
The recent PR #229 fixed my S4 MAD calculations, so I'm pretty confident this was an issue with NaN masking which is now handled, so I'm closing this issue now

Issue #121, opened 2022-02-17
Tiny error bars out of S4 when the S2 photom step is run


cpiaulet:
If I run S3, then S4 of Eureka! from the S2 jwst pipeline products (I tried NIRCam and NIRSpec so far), I get very small error bars on the points in individual wavelength-dependent light curves. 

Is there some other unit conversion that needs to happen on the uncertainties to correct for this? @taylorbell57 tells me this does not occur if the photom step is skipped.

On a related note, do we have plans to add the single-point scatter as a free parameter in the S5 fit? It could be useful if errors are over- or underestimated in previous steps.

cpiaulet:
Looking at it more closely, there seems to be something else at play other than the error bars themselves. When I run S3 and S4 again after running S2 myself without the phantom step, the background subtraction looks a lot different, as well as the light curves.
<img width="746" alt="fromS2jwst_S3output_Int10" src="https://user-images.githubusercontent.com/38121553/154551056-ccbd3783-8ddd-4585-9ed7-06c31b95f742.png">
<img width="740" alt="fromS2customnophotom_S3output_Int10" src="https://user-images.githubusercontent.com/38121553/154551061-d453db0a-7432-40d9-97b6-c34a91be7478.png">


<img width="991" alt="fromS2jwst_S4output_bandpass15" src="https://user-images.githubusercontent.com/38121553/154550822-83d2cf85-934c-452b-800c-4e5a91691e7c.png">
<img width="984" alt="fromS2customnophotom_S4output_bandpass15" src="https://user-images.githubusercontent.com/38121553/154550827-6bfa54f5-4b08-4902-bdd9-ce8297f23ed8.png">

taylorbell57:
Are the first plots with photom and the second without photom?

cpiaulet:
Yes

cpiaulet:
This is for the NIRSpec PRISM data set for the 2022 data challenge

taylorbell57:
Hmm, seems to me either the photom step is doing some irrecoverable damage or our attempts at converting units after the photom step are significantly flawed

cpiaulet:
Sure! and I'd be happy to help with this as well :)

kevin218:
The background subtraction is different between these two cases. I'm guessing the bad one uses median frame while the good one uses column-by-column.

kevin218:
[...]

PR #166, opened 2022-03-04
Automating ancil file download

### LINKED ISSUE #143: Unclear instructions on obtaining ancillary files for Stage 3
It is currently unclear what is supposed to be contained in the S3 ancildir or how users should go about getting those files.

taylorbell57:
Closes #143 by using code already written in the jwst pipeline to determine the needed reference files and download them

### DIFF
--- a/eureka/S3_data_reduction/bright2flux.py
+++ b/eureka/S3_data_reduction/bright2flux.py
@@ -1,5 +1,3 @@
-import sys
-from io import StringIO
 import numpy as np
 import scipy.interpolate as spi
 from scipy.constants import arcsec
--- a/eureka/S3_data_reduction/bright2flux.py
+++ b/eureka/S3_data_reduction/bright2flux.py
@@ -5,6 +5,39 @@ from astropy.io import fits
 from jwst import datamodels
 from jwst.photom import PhotomStep
 
+def rate2count(data):
+    """This function converts the data, uncertainty, and variance arrays from rate units (#/s) to counts (#).
+
+    Parameters
+    ----------
+    data:   DataClass
+        Data object containing data, uncertainty, and variance arrays in rate units (#/s).
+
+    Returns
+    -------
+    data:   DataClass
+        Data object containing data, uncertainty, and variance arrays in count units (#).
+
+    Notes
+    -----
+    History:
+    
+    - Mar 7, 2022 Taylor J Bell
+        Initial version
+    """
+    if "EFFINTTM" in data.mhdr.keys():
+        int_time = data.mhdr['EFFINTTM']
+    elif "EXPTIME" in data.mhdr.keys():
+        int_time = data.mhdr['EXPTIME']
+    else:
+        raise ValueError('No FITS header keys found to permit conversion from rate units (#/s) to counts (#)')
+    
+    data.subdata *= int_time
+    data.suberr  *= int_time
+    data.subv0   *= int_time
+
+    return data
+
 def dn2electrons(data, meta):
     """This function converts the data, uncertainty, and variance arrays from raw units (DN) to electrons.
 
@@ -86,8 +119,10 @@ def bright2dn(data, meta):
     foo = fits.getdata(meta.photfile)
     if meta.inst == 'nircam':
         ind = np.where((foo['filter'] == data.mhdr['FILTER']) * (foo['pupil'] == data.mhdr['PUPIL']) * (foo['order'] == 1))[0][0]
-    if meta.inst == 'miri':
+    elif meta.inst == 'miri':
         ind = np.where((foo['filter'] == data.mhdr['FILTER']) * (foo['subarray'] == data.mhdr['SUBARRAY']))[0][0]
+    else:
+        raise ValueError(f'The bright2dn function has not been edited to handle the instrument {meta.inst}, and can currently only handle miri and nircam observations.')
 
     response_wave = foo['wavelength'][ind]
     response_vals = foo['relresponse'][ind]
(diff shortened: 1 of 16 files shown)


eureka/S3_data_reduction/bright2flux.py
```diff
@@ -183,19 +184,30 @@ def convert_to_e(data, meta, log):
...
+        meta.photfile, meta.gainfile = download_ancil(data.filename)
+    
     if data.shdr['BUNIT'] == 'MJy/sr':
         # Convert from brightness units (MJy/sr) to flux units (uJy/pix)
         # log.writelog('Converting from brightness to flux units')
         # data = b2f.bright2flux(data, data.shdr['PIXAR_A2'])
         # Convert from brightness units (MJy/sr) to DNs
-        log.writelog('  Converting from brightness units (MJy/sr) to electrons')
-        meta.photfile = os.path.join(meta.topdir, *meta.ancildir.split(os.sep)) + '/' + data.mhdr['R_PHOTOM'][7:]
+        log.writelog('  Converting from brightness units (MJy/sr) to electrons per second (e/s)')
```
kevin218, line 198: This function is intended to return electrons, not electrons/sec.

eureka/S3_data_reduction/bright2flux.py
```diff
@@ -183,19 +184,30 @@ def convert_to_e(data, meta, log):
     meta:   MetaClass
         The metadata object.
     """
+    if data.shdr['BUNIT'] != 'ELECTRONS/S':
```
kevin218, line 187: If the units are already in electrons/s, why do we need to download any files?

eureka/S3_data_reduction/bright2flux.py
```diff
@@ -183,19 +184,30 @@ def convert_to_e(data, meta, log):
...
         data = dn2electrons(data, meta)
     if data.shdr['BUNIT'] == 'DN/s':
         # Convert from DN/s to e/s
         log.writelog('  Converting from data numbers per second (DN/s) to electrons per second (e/s)')
-        meta.gainfile = os.path.join(meta.topdir, *meta.ancildir.split(os.sep)) + '/' + data.mhdr['R_GAIN'][7:]
         data = dn2electrons(data, meta)
+    
     return data, meta
+
+def download_ancil(fitsname):
```
kevin218, line 208: We should not require internet access for users to be able to reduce data, so these downloaded files need to go somewhere (like an ancil directory) that can be checked later.

kevin218 (changes requested):
Unless I've missed something, it looks like Eureka! is downloading the gain and photom files every time it runs.  The most robust algorithm would be to check for files in the ancil directory.  If a file is missing, download and save it to that directory.  This way, an internet connection isn't required every time someone wants to reduce some data.  If the downloaded files are automatically saved in `crds_cache` then we need to be checking there first.

eureka/S3_data_reduction/bright2flux.py
taylorbell57: The statement is !=, so no files are downloaded if the units are already good.

eureka/S3_data_reduction/bright2flux.py
taylorbell57: The code uses the CRDS variables set in your environment (.bashrc file or equivalent), and it always checks to see if you have the required files first before trying to download anything from the internet. If you have the required files, no internet connection required. The files get saved wherever your CRDS cache is setup to be, which saves you the hassle of redownloading files if you ran stage 2 which would've downloaded the files already. The slight inconvenience is that it checks that you have the right gain and photom files for every file you reduce, but that could also be a feature if you have multiple visits or instruments in the same folder.

kevin218: Got it.  There's a lot going on under the hood here that isn't obvious.  Consider adding more comments and renaming `download_ancil` to something like `retrieve_ancil` so that there's no confusion.

eureka/S3_data_reduction/bright2flux.py
kevin218: so it is!

kevin218:
Before we completely ditch the `ancil` directory, is there any other reason someone would want to use it in the future?  For example, I put some forward models in there for S6.  Maybe someone will want to put stellar Phoenix models in there.

eureka/S3_data_reduction/bright2flux.py
taylorbell57: Huh, well HST data comes in with e/s (or so the FITS header claims) and the dn2electrons function does not multiply by the integration time, so there might be more issues here than just the log messages

taylorbell57:
> Before we completely ditch the `ancil` directory, is there any other reason someone would want to use it in the future? For example, I put some forward models in there for S6. Maybe someone will want to put stellar Phoenix models in there.

Fair enough, but I don't see how those files would be relevant to S3, so why still specify the ancildir in that stage in particular? And the S6 code doesn't need to know an ancildir - it just looks in the path specified by topdir+model_spectrum (as set in S6). There's nothing saying you can't make an ancilliary files folder, but as of right now there's no code that needs it

kevin218:
> Fair enough, but I don't see how those files would be relevant to S3, so why still specify the ancildir in that stage in particular? And the S6 code doesn't need to know an ancildir - it just looks in the path specified by topdir+model_spectrum (as set in S6). There's nothing saying you can't make an ancilliary files folder, but as of right now there's no code that needs it

You're right, those are later on.  Ok, let's move forward without ancil and see what happens.

eureka/S3_data_reduction/bright2flux.py
kevin218: Look at lines 65-67 `hst_scan.py`.  When the files are read in, the flux is multiplied by the sample time.

taylorbell57:
Alright, I partially resolve some issues and have addressed all your comments above. I've hopefully addressed Issue #121 and at least added error messages to alert users to Issue #169.

taylorbell57:
I just added a few small tweaks to also close https://github.com/kevin218/Eureka/issues/154 and maybe also https://github.com/kevin218/Eureka/issues/155, but I don't really know what's causing https://github.com/kevin218/Eureka/issues/155 so I can't guarantee that issue will be closed (we'll just have to wait and see if anyone encounters that issue again)

kevin218 (approved):

Issue #167, opened 2022-03-04
Light curve fitting doesn't work with masked arrays


kevin218:
When `sigma_clip=True` in Stage 4, the masked flux values are not normalized at the beginning of S5 (which is fine), but then the masked array gets dropped at some point and `lsq` tries to fit the light curve with some very large flux outliers.  I've found a couple places that use `np.copy(flux)` (which doesn't copy masked arrays) and switched to using `np.ma.MaskedArray.copy(flux)`, but the fitting still doesn't work right.  I'll continue tracking down the problem and report on progress here.

PR #168, opened 2022-03-04
Masked arrays in S5

kevin218:
This PR addresses Issue #167


taylorbell57:
I just added a few small tweaks to also close #154 and maybe also #155, but I don't really know what's causing #155 so I can't guarantee that issue will be closed (we'll just have to wait and see if anyone encounters that issue again)

PR #171, opened 2022-03-08
Patching bug with plotting chain, saving chain as HDF5

taylorbell57:
This PR builds upon #166, so that one should be merged first.

I maintained backwards compatibility with old saved csv chains, but I
now save chains as HDF5 files which are far more compressed and still
independent of programming language.

I also made a much better chain plot to show the evolution of the emcee
walkers with several visual aids. See below for a super-short emcee fit I
just did to the MIRI/LRS phase curve that is obviously nowhere near
burned in, but gives you the general idea of what the plot looks like.
There will be multiple figures if there are more than 12 parameters to
keep the figure size manageable. I'll try to do a longer run to show you
what the chain looks like when it's at least near converged, but those
horizontal dotted lines can be used as rulers to see if the median or 1-3
sigma level of the ensemble of walkers has changed.

fig550_fullchain_emcee_plot1of2

### DIFF
--- a/demos/JWST/S3_miri_lrs_template.ecf
+++ b/demos/JWST/S3_miri_lrs_template.ecf
@@ -34,4 +34,3 @@ topdir      /home/User
 # Directories relative to project dir
 inputdir    /Data/JWST-Sim/MIRI/Stage2	# The folder containing the outputs from Eureka!'s S2 or JWST's S2 pipeline (will be overwritten if calling S2 and S3 sequentially)
 outputdir	/Data/JWST-Sim/MIRI/Stage3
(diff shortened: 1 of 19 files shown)


taylorbell57:
Here's a bit more burned-in chain. Two figures are made, one with the burn-in steps included:

fig550_fullchain_emcee_plot1of2

And one without

fig550_chain_emcee_plot1of2

With that second figure it's far easier to tell if the median or standard deviations of the walkers have moved during the sampling thanks to the horizontal dotted lines which start at the starting value for the -3 sig,-2 sig,-1 sig,median,+1 sig,+2 sig,+3 sig levels

kevin218 (approved):

PR #173, opened 2022-03-08
Allowing user defined wavelength bins in Stage 4

taylorbell57:
Allowing user defined wavelength bins in Stage 4. This is beneficial to create outputs consistent with the requests of the ERS Data Challange


kevin218 (approved):

taylorbell57:
Oops, I realized I hadn't made the wave_hi and wave_low lists into arrays before which causes issues later, but I do now.

kevin218 (approved):

PR #174, opened 2022-03-08
Update fitters.py

iancrossfield:
lightcurve channels should be enumerated as 001, 002, 003,.... etc.  This way, directory sorting (e.g. via `ls`) returns the LC outputs in the proper channel order.

### DIFF
--- a/eureka/S5_lightcurve_fitting/fitters.py
+++ b/eureka/S5_lightcurve_fitting/fitters.py
@@ -661,7 +661,7 @@ def save_fit(meta, lc, fitter, fit_params, freenames, samples=[]):
         if lc.share:
             fname = f'S5_{fitter}_samples_shared.csv'
         else:
-            fname = f'S5_{fitter}_samples_ch{lc.channel}.csv'
+            fname = f'S5_{fitter}_samples_ch%03i.csv' % lc.channel
         np.savetxt(meta.outputdir+fname, samples, header=','.join(freenames), delimiter=',')
     
     return


iancrossfield:
PS: Addresses #175

eureka/S5_lightcurve_fitting/fitters.py
```diff
@@ -661,7 +661,7 @@ def save_fit(meta, lc, fitter, fit_params, freenames, samples=[]):
         if lc.share:
             fname = f'S5_{fitter}_samples_shared.csv'
         else:
-            fname = f'S5_{fitter}_samples_ch{lc.channel}.csv'
+            fname = f'S5_{fitter}_samples_ch%03i.csv' % lc.channel
```
taylorbell57, line 664: Instead use `str(lc.channel).zfill(len(str(lc.nchannel))))` which is consistent with how all the plot names from S5 are formatted. The change will also need to be made to the similar looking line above for the fitparams csv file, and S6 will also need to be updated similarly [here](https://github.com/kevin218/Eureka/blob/7458a371a644b69c9e842aaccadd47ec62dbe571/eureka/S6_planet_spectra/s6_spectra.py#L140)

taylorbell57 (changes requested):

eureka/S5_lightcurve_fitting/fitters.py
taylorbell57: Making this change now also has the disadvantage of obfuscating all chains that people have previously run (e.g. I've done some many hour long fits already, and making this change would mean those chains are all garbage). I'm not fully convinced that the slightly nicer formatting in the terminal is worth the losses it will result in

iancrossfield: That kind of sounds like arguing that the code shouldn't be improved pre-launch, because it will temporarily inconvenience current users. There have already been plenty of other syntax changes that people have to take the time to re-learn.  

Anyway, this isn't about 'looking nicer in the terminal' -- it affects analysis of the data.  If I want to load all the S5 outputs, I have to do
`import os
import pandas as pd
outputfiles = [line.strip() for line in os.popen('ls S5*ch*.csv')]
outputtables = [pd.read_csv(filename) for filename in outputfiles]
transitspectrum = [table['# rp']**2 for table in outputtables]
... etc.
`
Unless there's something I'm missing in the CSV outputs that includes what wavelength channel is which, the snippet above currently results in a mis-ordered transit spectrum.

taylorbell57: Sorry, my point wasn't that the change should never be made (I agree that it is an improvement), just that it's likely worth waiting a week or few until the data challenge is over so that I and others don't have to redo fits. We've been trying to avoid changing the inputs/outputs of any of the stages in a way that breaks backwards compatibility while so many people are using eureka for the data challenge since that'll likely result in confusion and frustration. Kevin is away until the weekend too, and we typically wait for his review before merging anything into main anyway. For the time being, you or others could do something like the following:

```
import glob
import numpy as np
import pandas as pd
import matplotlib.pyplot as plt
import eureka.lib import manageevent as me

# Get list of output files
fnames = glob.glob('S5*ch*.csv')
# Get the ordering of channels
channels = [int(fname[fname.find("ch")+2:-4]) for fname in fnames]
# Sort fnames properly
fnames = fnames[np.argsort(channels)]
# Read in the fitted values from S5
outputtables = [pd.read_csv(fname) for fname in fnames]
transitspectrum = [table['# rp']**2 for table in outputtables]

# Load in the metadata object to get wavelength information
meta = me.loadevent('S5_nircam_wfss_Meta_Save.dat')
wavelengths = np.mean(np.append(meta.wave_low.reshape(1,-1), meta.wave_hi.reshape(1,-1), axis=0), axis=0)
wave_errs = (meta.wave_hi-meta.wave_low)/2

# Do something like plotting the data
plt.errorbar(wavelengths, transitspectrum, xerr=wave_errs)
plt.show()
```

Note that if you want uncertainties on the fitted parameters to add y errors to the plot above, you'll need to load in the chain from emcee or dynesty and use np.percentile or np.std to determine the uncertainty.

AarynnCarter:
Note that I'm currently getting the following error from this file. Should be a quick fix, but worth incorporating with this pull request. 

```
  File "/Users/acarter/Documents/TRANSITS/EUREKA/DC_NIRSPEC_FULL/../Eureka_dev/eureka/S5_lightcurve_fitting/fitters.py", line 522
    return pos, nwalkers
TabError: inconsistent use of tabs and spaces in indentation
```

kevin218:
Now is probably the best time to break backwards compatibility. There will likely be more big changes this month as we move towards our first release.

AarynnCarter:
@taylorbell57 Sorry, wasn't aware it was already addressed. I just pulled from main and it cropped up, then saw this in the pull requests. Looks like you're on top of it.

PR #180, opened 2022-03-09
Added CRDS env var instructions to installation page.

jbrande:
Also, added /docs/build/ to .gitignore, and tweaked text of contribution page to reflect the current unit testing setup.


docs/source/installation.rst
```diff
@@ -46,6 +46,20 @@ With Git/GitHub
...
 
+CRDS Environment Variables
+--------------------------
+
+Eureka! installs the JWST Calibration Pipeline as part of its requirements, and this also requires users to set the proper environment variables so that it can download the proper reference files needed to run the pipeline. 
+For users not on the internal STScI network, two environment variables need to be set to enable this functionality:
+
+	.. code-block:: bash
+
+		export CRDS_PATH=/PATH/TO/FOLDER/crds_cache
```
taylorbell57, line 57: It's likely worth specifying that these need to be in the user's .bashrc or similar file (e.g. .bash_profile) otherwise they're only active for that particular terminal instance

jbrande: Ah, you're totally right. I'll add that.

taylorbell57 (approved):

PR #181, opened 2022-03-09
S3 mansfield

meganmansfield:
Minor updates to 1) make the .ecf background and source position specifications a bit clearer, and 2) allow bright2flux.py to extrapolate the response function to the edges of the image when they are at wavelengths beyond those defined by the response function.


meganmansfield:
Ok actually making this change led to some plotting errors where there were NaNs, but those are all fixed now. I just added a few statements to replace NaNs with zeros after the optimal extraction.

eureka/S3_data_reduction/s3_reduce.py
```diff
@@ -334,7 +334,9 @@ def reduceJWST(eventlabel, ecf_path='./', s2_meta=None):
                                                                              deg=meta.prof_deg, n=data.intstart + n,
                                                                              isplots=meta.isplots_S3, eventdir=meta.outputdir,
                                                                              meddata=data.medapdata, hide_plots=meta.hide_plots)
-
+                #Replace NaNs with zero
```
taylorbell57, line 337: Zeros won't have the same impact on later steps though since things like averages will actually use those points rather than ignoring them. Wouldn't a better option would be numpy masked arrays?

eureka/S3_data_reduction/plots_s3.py
```diff
@@ -206,6 +206,7 @@ def profile(eventdir, profile, submask, n, hide_plots=False):
     -------
     None
     '''
+    profile[np.where(np.isnan(profile))] = 0
```
taylorbell57, line 209: Zeros won't have the same impact on later steps though since things like averages will actually use those points rather than ignoring them. Wouldn't a better option would be numpy masked arrays?

eureka/S3_data_reduction/s3_reduce.py
meganmansfield: Yes, that sounds like a much better solution - how do you make a numpy masked array?

taylorbell57: https://numpy.org/doc/stable/reference/generated/numpy.ma.masked_invalid.html

meganmansfield: Sweet thanks, fixed now!

eureka/S3_data_reduction/plots_s3.py
meganmansfield: Fixed as above.

eureka/S3_data_reduction/plots_s3.py
```diff
@@ -206,6 +206,7 @@ def profile(eventdir, profile, submask, n, hide_plots=False):
     -------
     None
     '''
+    profile = np.ma.masked_invalid(profile)
     vmax = 0.05*np.max(profile*submask)
```
taylorbell57, line 210: You'll likely want to change this to np.ma.max

eureka/S3_data_reduction/s3_reduce.py
taylorbell57: You should replace the comment about replacing NaNs with zeros as well

eureka/S3_data_reduction/plots_s3.py
meganmansfield: Fixed, thanks

eureka/S3_data_reduction/s3_reduce.py
meganmansfield: Ah yes, fixed now.

kevin218 (dismissed):

taylorbell57 (approved):

PR #183, opened 2022-03-10
S5 gp merge

evamariaa:


eureka/S5_lightcurve_fitting/likelihood.py
```diff
@@ -43,7 +43,11 @@ def ln_like(theta, lc, model, freenames):
                 lc.unc_fit[chan*lc.time.size:(chan+1)*lc.time.size] = theta[ind[chan]] * 1e-6
     else:
         lc.unc_fit = deepcopy(lc.unc)
-    ln_like_val = (-0.5 * (np.sum((residuals / lc.unc_fit) ** 2+ np.log(2.0 * np.pi * (lc.unc_fit) ** 2))))
+    if model.GP:
+        ln_like_val = model.GP_model_likelihood(model_lc)
+    else:
+        residuals = (lc.flux - model_lc) #/ lc.unc
+        ln_like_val = (-0.5 * (np.sum((residuals / lc.unc) ** 2+ np.log(2.0 * np.pi * (lc.unc_fit) ** 2))))
```
taylorbell57, line 50: Should be `lc.unc_fit` for chi-squared calculation here

eureka/S5_lightcurve_fitting/models/Model.py
```diff
@@ -304,13 +328,19 @@ def physeval(self, interp=False, **kwargs):
                     flux *= model.interp(new_time, **kwargs)
                 else:
                     flux *= model.eval(**kwargs)
-        
         return flux, new_time
+    
+    
+    def GP_model_likelihood(self,  fit, **kwargs):
```
taylorbell57, line 334: I don't think this belongs in the gerenal Model.py file - why not just a general function inside the new GPModel.py (no need for it to be a function inside a class).

taylorbell57: Or it could be in likelihood.py

eureka/S5_lightcurve_fitting/fitters.py
```diff
@@ -307,7 +310,11 @@ def emceefitter(lc, model, meta, log, **kwargs):
         if len(ind)>1:
             for chan in np.arange(lc.flux.size//lc.time.size):
                 lc.unc_fit[chan*lc.time.size:(chan+1)*lc.time.size] = fit_params[ind[chan]] * 1e-6
-
+        
+    #Plot GP fit + components
+    if model.GP:
+        plots.plot_GP_components(lc, model, meta, fitter='emcee')
```
taylorbell57, line 316: Is there a reason the lsqfitter doesn't get a similar plot?

demos/JWST/S5_template.ecf
```diff
@@ -27,6 +27,10 @@ run_bound       'multi'
 run_sample      'unif'
 run_tol         0.1
 
+#GP inputs
```
taylorbell57, line 30: Should note that a GP can be used in run_myfuncs comment above

eureka/S5_lightcurve_fitting/plots_s5.py
```diff
@@ -324,5 +311,87 @@ def plot_res_distr(lc, model, meta, fitter):
...
+        if lc.share:
+            flux = flux[channel*len(lc.time):(channel+1)*len(lc.time)]
+            unc = unc[channel*len(lc.time):(channel+1)*len(lc.time)]
+            model = model[channel*len(lc.time):(channel+1)*len(lc.time)]
+            model_sys = model_sys[channel*len(lc.time):(channel+1)*len(lc.time)]
+            model_phys = model_phys[channel*len(new_time):(channel+1)*len(new_time)]
+            model_GP_component = model_GP_component[channel*len(lc.time):(channel+1)*len(lc.time)]
+        
+        residuals = flux - model
+        fig = plt.figure(int('51{}'.format(str(channel).zfill(len(str(lc.nchannel))))), figsize=(8, 6))
```
taylorbell57, line 374: Plot number 51 has already been used - we should likely use a different number here and in the plot file name

eureka/S5_lightcurve_fitting/models/GPModel.py
```diff
@@ -0,0 +1,250 @@
+import numpy as np
+import george
```
taylorbell57, line 2: If george isn't already in the requirements.txt file, you should add it. Same for tinygp when you get around to testing that one out

taylorbell57 (changes requested):
In addition to my comments in the code, please also delete all of the zip files you have uploaded

demos/JWST/S5_template.ecf
taylorbell57: Should also update the ecf templates in the docs/media folder and the eureka/testing/NIRCam and eureka/testing/NIRSpec folders

eureka/S5_lightcurve_fitting/fitters.py
evamariaa: Problem is that the lsq fit doesn't seem to be fitting the GP and I'm not entirely sure why, that's the reason there's no GP plot for that yet. I'll raise it as an issue and assign myself to look into that after this.

eureka/S5_lightcurve_fitting/plots_s5.py
evamariaa: I thought it'd be good to have the two light curve plots in packs, but can change the number of course if you think that's confusing.

eureka/S5_lightcurve_fitting/models/Model.py
evamariaa: Good point, I'll move it to likelihood.py

eureka/S5_lightcurve_fitting/likelihood.py
```diff
@@ -272,3 +255,29 @@ def computeRMS(data, maxnbins=None, binstep=1, isrmserr=False):
...
+    likelihood of the model
+    
+    Notes
+    -----
+    History:
+    
+    - March 11, 2022 Eva-Maria Ahrer
+        moved code from Model.py
+    """
+    for model in self.components:
```
taylorbell57: This needs to be updated since self is no longer defined. And careful as you have two different `model` variables now

eureka/tests/NIRCam_ecfs/S5_fit_par.ecf
```diff
@@ -13,3 +13,8 @@ u1			0.3				'free'			-1				1				U
 u2			0.1				'free'			-1				1				U
 transittype	'primary'	 	'independent'
 c0			1.009			'free'			0.95			1.05			U
+#GP parameters
+A			-5.			'free'			-15.			5.			U
```
taylorbell57, line 17: These parameters likely need to be commented out in all the ecfs unless GP is in run_myfuncs, right? Otherwise the fitter will try to fit these unused parameters which will slow things down for no reason

eureka/S5_lightcurve_fitting/plots_s5.py
taylorbell57: Hmm, let's see what Kevin thinks

eureka/S5_lightcurve_fitting/fitters.py
```diff
@@ -113,6 +113,9 @@ def callback_full(theta, lc, model, prior1, prior2, priortype, freenames):
     # Save the covariance matrix in case it's needed to estimate step size for a sampler
     model_lc = model.eval()
 
+    # Save the covariance matrix in case it's needed to estimate step size for a sampler
+    model_lc = model.eval(incl_GP=True)
```
kevin218, line 117: It looks like we're calling this function twice. Do we need both?

eureka/S5_lightcurve_fitting/plots_s5.py
kevin218: No to repeating 51. My suggestion would be to use 52 and relabel the current 52 to something else.

evamariaa: Okay, thanks Kevin!
I've now changed the GP plot to 55 (i.e. the last number) so that it won't get confusing to change the 52 plots, especially when users don't run GPs that there aren't missing numbered plots.

taylorbell57: Note that you're still using plt.figure(51...) though, so that line should also be 55

taylorbell57 (changes requested):

eureka/S5_lightcurve_fitting/plots_s5.py
taylorbell57: Can you change this line to the following so that only one plot gets opened (and future channels overwrite the plot) to limit RAM usage.

`fig = plt.figure(int('51{}'.format(str(0).zfill(len(str(lc.nchannel))))), figsize=(8, 6))`

taylorbell57 (changes requested):
Sorry, one last change and then I think this should be ready to merge - I'd forgotten about it until just now when I checked which PRs were still open

taylorbell57 (approved):

kevin218 (approved):

PR #199, opened 2022-03-21
Computes MAD in S4 from wave_min to wave_max

kevin218:


taylorbell57:
How does this differ from the MAD computed in Stage 3? And do you mean to have included the fits file in the test folder?

meganmansfield (approved):
This looks good to me! And I get almost the same MAD as I did using my old cropping method before. Same question about the included fits file, but otherwise it looks good!

kevin218:
> How does this differ from the MAD computed in Stage 3? And do you mean to have included the fits file in the test folder?

S3 computes MAD over the entire wavelength range specified by meta.xwindow, meta.ywindow (including the edges where flux values can get really low).  S4 computes MAD only over the range specified by meta.wave_min, meta.wave_max.

I don't know why the fits file is there.

taylorbell57:
Oh wait, that fits file is supposed to be there for pytest - it get's edited during pytest though, so you have to revert the changes to that file (annoyingly).

taylorbell57 (approved):

taylorbell57:
FYI I added back in that FITS file myself, so your PR is good to go

PR #200, opened 2022-03-21
Resolve #182 by auto setting wave_min/max if None

### LINKED ISSUE #182: S4 wave_min and wave_max option for using full S3 extracted region
It would be nice if instead of defining specific wave_min and wave_max in the S4 .ecf file you could say something like 'full' and then wave_min/wave_max would just be set to the minimum/maximum wavelength value for the subarray region defined in S3.

taylorbell57:
Automatically sets the wave_min and wave_max values to the min and max
of the extracted spectra if the values are None in the S4 ecf.

I've confirmed that this still passes the pytest tests.


meganmansfield (approved):
This worked great for me!

PR #215, opened 2022-03-22
Cleaning up ecf templates

taylorbell57:
Removing all the tabs from ecf templates and replacing them with spaces to make sure the templates look nicely formatted for everyone regardless of the text file viewer. Also made some minor tweaks to ecf defaults and added a bit of documentation to the FAQ page to answer some frequently asked issues on Slack.

### DIFF
--- a/demos/JWST/S1_miri_template.ecf
+++ b/demos/JWST/S1_miri_template.ecf
@@ -1,40 +1,40 @@
 # Eureka! Control File for Stage 1: Detector Processing
 
-suffix 						uncal
+suffix              uncal
 
 # Control ramp fitting method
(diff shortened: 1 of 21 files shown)


jbrande:
Generally looks okay, but in the various S2 ECFs, some have a `skip_resample` parameter and some call that `skip_resample_spec`. I think `skip_resample` is the correct version of the parameter? @taylorbell57 could you double check this?

jbrande (changes requested):
Sorry, my point about `skip_resample` should've been in a review. Here it is again.

taylorbell57:
Great catch, I think that should have fixed it

jbrande (approved):

Issue #236, opened 2022-03-28
remove jwst dependency for S3


lkreidberg:
THis will make a eureka-lite install much more functional, and the jwst pipeline isn't crucial for S3 ( I think)

sebastian-zieba:
I will hard code the wavelengths for the simulated MIRI data in miri.py and add a warning that the hard coded values will be used if the user does not want to install the jwst package.

Issue #231, opened 2022-03-24
Failed to install dependencies


ommnmm:
![Screen Shot 2022-03-23 at 12 53 25 PM](https://user-images.githubusercontent.com/23083481/160026024-08d3e8a5-4e47-4a0c-bffe-bf128c8f0fad.png)
![Screen Shot 2022-03-23 at 12 56 42 PM](https://user-images.githubusercontent.com/23083481/160026045-0ae922fd-a155-4a6f-8b50-25b857005867.png)

Dependencies were installed in new conda environments using pip install / pip install -r requirements.txt

MacOS 11.5.2  ,  conda 4.10.3,  processor: Intel
Note: conda 4.11.0 seems to generate worse errors -  Failed to build Eureka jwst stsci.image batman-package stsci.imagestats stsci.stimage

taylorbell57:
Hi Yucian, thanks for posting this error here so we can continue our discussion from Slack. Can you try something for me? In a terminal, please run the following lines and then post a screenshot (or several screenshots if needed) here showing your commands and the messages they print to the console? I want to isolate and understand this issue as much as possible.

`conda create -n eureka python==3.9 -y`
`conda activate eureka`
`conda info`
`which pip`
`pip install jwst==1.3.3`

ommnmm:
got error for building jwst, stsci.image, stsci.imagestats, stsci.stimage

<img width="758" alt="Screen Shot 2022-03-24 at 5 20 51 PM" src="https://user-images.githubusercontent.com/23083481/160031002-07b73df4-8f4f-4495-9fab-9d08df83494b.png">
<img width="669" alt="Screen Shot 2022-03-24 at 5 21 24 PM" src="https://user-images.githubusercontent.com/23083481/160031008-3ae3fa14-0ad0-4680-9b8f-4667bbf59ad8.png">
 
<img width="663" alt="Screen Shot 2022-03-24 at 5 21 38 PM" src="https://user-images.githubusercontent.com/23083481/160031031-e7b8adb0-99
<img width="1266" alt="Screen Shot 2022-03-24 at 5 22 40 PM" src="https://user-images.githubusercontent.com/23083481/160031049-438041f6-aa78-4834-948c-ac11d934a358.png">
3a-4c00-9e77-f2f78acdcd48.png">
[...]

PR #243, opened 2022-03-29
Making jwst an optional install

### LINKED ISSUE #237: add a eureka-lite install option
This would be an optional install that works for S4 - S6 (so that users aren't completely stuck if they fail to install the jwst pipeline)

taylorbell57:
This PR resolves #237 and gets close to resolving issue #236, but the wave_MIRI function in eureka/S3_data_reduction/miri.py still uses jwst and I don't know how to reproduce it's function without the jwst package. I believe @sebastian-zieba wrote this code, so perhaps he might be able to think of a way to do the same thing without the jwst package.

I also updated the docstrings for Stages 1-2, edited the documentation about how to install eureka (also resolves an issue where the git:// URL wouldn't work for some people and uses https:// instead), and I added a bit of documentation in the FAQ page about what to try to do if jwst doesn't install properly.

One last thing I want to do before merging this PR is adding clear instructions to make a new conda environment to the installation page and the README file so that we can at least reduce the number of issues caused by that.


eureka/tests/test_general.py
```diff
@@ -5,7 +5,10 @@
 
 import numpy as np
 from ..lib import util
-from ..S2_calibrations import s2_calibrate as s2
+try:
```
kevin218, line 8: I imagine anyone testing the pipeline will have the jwst packaged installed.

eureka/S3_data_reduction/wfc3_1reduce.py
```diff
@@ -1,954 +0,0 @@
-#! /usr/bin/env python
```
kevin218, line 1: Did you intend to delete everything from this file ?

kevin218 (commented):
Did you intend to change the fits file?

eureka/tests/test_general.py
taylorbell57: Well I wanted to be able to test that the pipeline would work as intended without jwst installed too though, otherwise it was hard for me to tell what I might have broken

eureka/S3_data_reduction/wfc3_1reduce.py
taylorbell57: Yeah, this file was first uploaded when Eureka was made and hasn't been used since then

taylorbell57:
I did intend to add that Stage 2 output FITS file for the NIRSpec testing data, otherwise one couldn't start the Stage 3 test without jwst installed since we'd only uploaded a Stage 1 output FITS file

kevin218:
Ok, I'm good with this PR

taylorbell57:
Great! I just found the solution to some of the jwst installation issues (e.g. https://github.com/kevin218/Eureka/issues/231) which I'm in the process of adding documentation for. This will help in cases where people just couldn't install jwst at all.

I've also been looking into making an environment.yml file that people could use to install most of the required packages using conda (excluding crds, myst-parser, sphinx-rtd-theme, stdatamodels, svo-filters, jwst, and stcal which are all only hosted on PyPI). With this new yml file, one would simply do `conda env create -f environment.yml` which would create a new conda environment named eureka, install all the requirements it can using conda, and then install the remaining requirements using pip. The benefit would be that most of the requirements would be conda controlled which would hopefully reduce some issues caused by people installing things after eureka. The issue with this technique though is that conda doesn't allow the specification of optional requirements, so we'd either have to force the installation of jwst with this technique or have people separately run a `pip install jwst==1.3.3 stcal` command after the `conda env create ...` command above if they want to run Stages 1-2. We can either just keep the pip install instructions we have, just use the environment.yml file and change the installation instructions, or allow both methods and explain both options in the documentation. Thoughts on the best solution?

kevin218:
I think the conda environment will give us the least headaches down the road. If the jwst install issue has really been resolved (I saw the other thread) then this should be the preferred mode of installation. 

How much work would it be to support both for a few months and see how things go?

taylorbell57:
I'd want to test the conda environment more before we set that as the recommended installation method (probably just need one more day of testing on my end), but I think it wouldn't be too hard to maintain both installation methods for some time. It'll just make the installation instructions a bit more busy, but that's not the worst. The main thing I want to look into first is other isses with jwst like those Megan encountered where installing something after eureka broke her jwst installation. If I can reproduce that, then I can set somewhat stringent requirements in the environment.yml file which should hopefully help avoid at least that one issue Megan encountered if not other similar issues

taylorbell57:
Alright, well I've added some more documentation, a new conda-based installation method, new installation instructions, and added some version requirements that I and/or others found would break the code if not obeyed. I've tested the installation as rigourously as I can on macOS 11.6.5 and in a Docker running Ubuntu 20.04. I can't tell if any of this resolves the issues that Megan encountered where installing `pip install jupyterlab` would break Stage 2 for her, but so far I cannot reproduce that issue with these installation requirements. Megan said she'd see if she can reproduce the error she was experiencing, and if she can then I'll add in some extra requirements to try to avoid that issue.

I still don't know how to make S3 independent of jwst for MIRI, but Sebastian said he'd look into it soon. We can either wait for that change or merge these edits in for now if you're happy with them.

kevin218 (approved):

PR #244, opened 2022-03-29
Issue template

meganmansfield:
Hey, @kevin218, this is my first attempt at making an issue template. What I'm trying to set up now is a workflow script that will automatically assign an issue to different people based on whether the issue is submitted for one of the four instruments or for general lightcurve fitting. However, after setting this up I realized I can only view the template and test it out if it's merged into the main branch. Would you be ok with going ahead and merging this? I promise I won't make changes to any of the other files. :)


kevin218 (approved):
Sure, approved!

Issue #193, opened 2022-03-18
S3: NIRSpec full dataset causes background flagging procedure to assume all pixels are bad. 


AarynnCarter:
When moving from the Tiny dataset to the full dataset for NIRSpec, the extra frames introduce more NaN's per pixel over the full timespan (which get converted to 0's) which in turn causes the background flagging procedure to think every pixel is bad. 

The solution for me was to duplicate the NIRCam function currently being used into the NIRSpec flag_bg function and adjust the following two lines, so that they mask out any values that are equal to 0.  

```
    bgerr1  = np.ma.median(np.ma.masked_equal(data.suberr[:,  :y1],0))
    bgerr2  = np.ma.median(np.ma.masked_equal(data.suberr[:,y2:  ],0))
```

Bigger questions underpin this...

1) Do we still need to be converting the NaN's to zeroes? 
2) Why are there so many NaN's in the NIRSpec simulated data
3) Will this be an issue for the real data? I.e. do we need to spend a lot of time on this

kevin218:
With the switch to masked arrays, it doesn't make sense to convert bad pixels to 0.

AarynnCarter:
Okay unfortunately this wasn't as quick of a fix as removing the lines converting the NaN's to zeroes. If I do so the MAD plot is empty. Interestingly, the BGSUB images look quite different between the adjusted flag_bg I described above, and the default NIRCam flag_bg function. With the default method, it looks like the background is barely even subtracted, and the background pixel flux changes a lot between integrations (same scaling), for the adjusted method it looks much more uniform. 

Will take a look later this week, but wanted to provide a quick update. 

![fig3101-2D_LC](https://user-images.githubusercontent.com/23636747/160507779-ee2b967d-d020-4c49-b6de-3b27adbe0ecb.png)

AarynnCarter:
On a separate point I also think this issue may be what drove the strict requirements on bg_hw that @erinmmay metioned in #121, as a smaller bg_hw could provide some pixels where there are no NaN's present throughout the time series (i.e. where the spectrum is).

AarynnCarter:
Okay, I've figured it out, and this is an issue with the NIRSpec data itself. Whilst the NaN's are being correctly masked in the SCI data array, the ERR array is filled with zeroes. I've looked back at data I've made going back to May 2021 and they all exhibit a similar abundance of zeroes. 

[...]

Issue #99, opened 2022-01-28
Poor NIRSpec Performance


taylorbell57:
Hi all, with my new computer I have much more RAM and am now able to process the NIRSpec data. My first attempt at doing so turned out very poorly (see attached image below). I will note that this was reduced with `jwst` v1.3.3, but I don't think that is likely to be the source of any issues. I've included the ecf files used in Stage 2 and 3 for reference (in a zip file to pacify GitHub). If @AarynnCarter or others have gotten better results with NIRSpec and can suggest better ecf settings, that would be much appreciated.

![fig3101-2D_LC](https://user-images.githubusercontent.com/24514122/151602116-2132cc00-cbbc-4bc7-b792-f68023786b76.png)

[NIRSpec ECFs.zip](https://github.com/kevin218/Eureka/files/7961003/NIRSpec.ECFs.zip)

AarynnCarter:
Hey Taylor, I think @kevin218 got pretty much the same result as this at some point in the past. I'm in the process right now of finalising a new NIRSpec dataset / tiny version and don't want to spend too much (if any) time looking at the old dataset as there are a few changes since then. Once I have that done (I'm hoping today / Monday next week), I'll do a quick run through Eureka to see whether the same bug crops up again and let you know. 

One thing I am confident of is that the data isn't *this* bad, and it's some pipeline step that's going wrong / not interacting with the data in the way we expect it to.

taylorbell57:
Very much understand that - thanks for looking into this when you get the chance! One thing to note is that I started from the Stage 1 outputs, and given your comment on the ERS Slack I suspect these observations also do not have a transit injected into them at this stage? Good to keep in mind, but the data should indeed look far cleaner than this

taylorbell57:
Oh right, these plots look almost the exact same as in Issue #35

AarynnCarter:
Yeah if you started from the uncal.fits file, there won't be any transit injected.

taylorbell57:
Ahh, I started from the _rateints.fits file, so there is actually a transit injected. Kevin's guess on Issue #35 was right that our threshold constraints were too tight - increasing p3thresh, p5thresh, and p7thresh to 10, 20, 20 instead of 5, 10, 10 significantly improves the outputs of Stage 3 (see attached image where a transit is actually visible). I'm not sure what is driving the rest of the awfulness, but it's possible we need even looser thresholds

[...]

Issue #169, opened 2022-03-07
bright2dn function not edited for NIRSpec or NIRISS


taylorbell57:
The bright2dn function in Stage 3 has not been edited for NIRSpec or NIRISS observations, and so an `ind` variable is never defined, so any NIRISS or NIRSpec file that has had the photom step run on it in Stage 2 will crash. I'm not sure what the code should look like for NIRSpec and NIRISS, so for now I'm just going to add an error message and if others know what to do they should implement those changes

taylorbell57:
@jbrande just pointed out to me that if he ran the NIRSpec files with the photom step in S2 that he didn't get an error message like that mentioned in above which alerted me to the issue that no unit conversion is done at all if the photom step is run. At least for NIRSpec, the BUNIT FITS header ends up being MJy not MJy/sr, so no unit conversion happens (and so the above issue is avoided). This is definitely not the intended action. I have confirmed that MIRI run through S2 with the photom step gives MJy/sr, and I'd assume NIRCam does as well, so this seems unique to NIRSpec and perhaps NIRISS (can you confirm @afeinstein20). If someone that's more familar with these units (e.g. @kevin218) wants to write a slightly tweaked version of bright2dn for just MJy input units, that'd be good!

PR #254, opened 2022-04-01
Fixes for NIRSpec

AarynnCarter:
For certain .ecf parameters it is not possible to perform a background subtraction (see #193) as the median of the error arrays is exactly 0. Adjusting the NIRSpec code to mask these zeroes when calculating the bg_err values. 

This likely only applies for the simulated data, so a warning has been added to the log to remind people when the real data comes around. 

Also removed the manual adjustment from NaN's to 0's as it is no longer needed.


AarynnCarter:
Just adjusted bright2flux.py to be able to digest the NIRSpec photom crds file. Also added a quick option to convert from MJy to DN/s within the existing `bright2dn()` function using the steradian to pixel conversion in the SCI header. 

This addresses Issue #169 partially, I could probably get NIRISS incorporated next week also.

eureka/S3_data_reduction/bright2flux.py
```diff
@@ -118,6 +118,8 @@ def bright2dn(data, meta):
         ind = np.where((foo['filter'] == data.mhdr['FILTER']) * (foo['pupil'] == data.mhdr['PUPIL']) * (foo['order'] == 1))[0][0]
     elif meta.inst == 'miri':
         ind = np.where((foo['filter'] == data.mhdr['FILTER']) * (foo['subarray'] == data.mhdr['SUBARRAY']))[0][0]
+    elif meta.inst == 'nirspec':
+        ind = np.where((foo['filter'] == data.mhdr['FILTER']) * (foo['grating'] == data.mhdr['GRATING']) * (foo['slit'] == data.shdr['SLTNAME']))[0][0] 
     else:
         raise ValueError(f'The bright2dn function has not been edited to handle the instrument {meta.inst}, and can currently only handle miri and nircam observations.')
```
taylorbell57, line 124: You can add NIRSpec to the list in this message now

eureka/S3_data_reduction/s3_reduce.py
```diff
@@ -173,6 +173,7 @@ def reduceJWST(eventlabel, ecf_path='./', s2_meta=None):
                 from . import nircam as inst
             elif meta.inst == 'nirspec':
                 from . import nirspec as inst
+                log.writelog('WARNING: Are you using real JWST data? Take a look at at the flag_bg() function in nirspec.py and Issue #193 on Github!')
```
taylorbell57, line 176: I would instead write:
`log.writelog('WARNING: Are you using real JWST data? If so, you should edit the flag_bg() function in nirspec.py and look at Issue #193 on Github!')`

taylorbell57 (changes requested):

taylorbell57:
Are there any related issues that can be closed with this PR?

AarynnCarter:
Okay I incorporated your tweaks @taylorbell57, and I've also added functionality to do the conversion for NIRISS as well.

However, @afeinstein20, @kevin218, I'd appreciate a few more sets of eyes on the NIRISS line, as I'm not entirely sure whether order=1 is the correct choice, or we need to do something more complicated. 

@taylorbell57 this closes #193, and #169, and maybe we could close #99 at the same time. However, I'm not sure how to formally connect these to the pull request beyond including them in the comment text.

taylorbell57:
> @taylorbell57 this closes #193, and #169, and maybe we could close #99 at the same time. However, I'm not sure how to formally connect these to the pull request beyond including them in the comment text.

I added them for now, but in the future you can either use the Development section on the right of a PR or use "closing keywords" in your comments (described [here](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue))

kevin218 (approved):

PR #255, opened 2022-04-04
Stage specific output documentation

jbrande:
I added an extra page of output plot descriptions independent of the ECF page. I think the output descriptions should be fully up to date with the code, and there are examples of nearly all the plots. The Stage 5 run I did wasn't fitting so well and I haven't sat down to diagnose that yet, but I'll do that today or tomorrow. 

I also added links to the ECF documentation in each demo ECF.


docs/source/outputs.rst
```diff
@@ -0,0 +1,119 @@
...
+
+
+.. _s5-out:
+
+Stage 5 Outputs
+---------------
+
+In Stage 5:
+   - If ``isplots_S5`` = 1: ``Eureka!`` will plot the fitted lightcurve model over the data in each channel.
+   - If ``isplots_S5`` = 2: ``Eureka!`` will plot an RMS deviation plot for each channel to help check for correlated noise.
```
taylorbell57, line 115: This is not correct - the fig5200 name does not mean isplots_S5>=2. It is just the second plot made in S5, and it gets called if isplots_S5>=3. I assume you'll need to update all of these notes about plots accordingly

taylorbell57: It's possible the intended naming convention had been what you assumed, but it is not that way at least for Stage 5 (and I think Stage 4) although I think that is how the plot names are made in Stage 3. @kevin218, we should really settle on one figure naming convention and use it throughout all stages.

taylorbell57 (changes requested):
You should double check that there aren't any duplicates of figures in the docs/media folder and the docs/media/S#_out folders

docs/source/outputs.rst
jbrande: These should be fixed now. I agree the naming conventions should be consistent, and I think the way it works in Stage 3 actually works well. We might want to update the specific numbering eventually, though.a

taylorbell57:
Reminder to add the Allan plot when you get the chance

jbrande:
Alright, @taylorbell57, this should be all set.

docs/source/ecf.rst
```diff
@@ -540,16 +497,48 @@ The path to the directory in which to output the Stage 5 JWST data and plots.
...
+   The Stage 5 fit parameter file has the file extension ``.epf``, not ``.ecf``. These have different formats, and are not interchangeable.
+
+This file describes the transit/eclipse and systematics parameters and their prior distributions. Each line describes a new parameter, with the following basic format:
 
 ``Name    Value    Free    PriorPar1    PriorPar2    PriorType``
 
-The ``PriorType`` can be U (Uniform), LU (Log Uniform), or N (Normal). If U/LU, then ``PriorPar1`` and ``PriorPar2`` are the lower and upper limits of the prior distribution. If N, then ``PriorPar1`` is the mean and ``PriorPar2`` is the stadard deviation of the Gaussian prior.
+``Name`` defines the specific parameter being fit for. Available options are:
+   - Transit and Eclipse Parameters
+      - ``rp`` - planet radius, for the transit models.
```
taylorbell57, line 509: This should be the planet/star radius ratio

docs/source/ecf.rst
```diff
@@ -540,16 +497,48 @@ The path to the directory in which to output the Stage 5 JWST data and plots.
...
+   - Phase Curve Parameters - the phase curve model allows for the addition of up to four sinusoids into a single phase curve
+      - ``AmpCos1`` - Amplitude of the first cosine 
+      - ``AmpSin1`` - Amplitude of the first sine
+      - ``AmpCos2`` - Amplitude of the second cosine
+      - ``AmpSin2`` - Amplitude of the second sine
+   - Limb Darkening Parameters
+      - ``limb_dark`` - The limb darkening model to be used. Options are: ``['uniform', 'linear', 'quadratic', 'kipping2013', 'square-root', 'logarithmic', 'exponential', '4-parameter']``
+      - ``u1`` - The first limb darkening parameter
+      - ``u2`` - The second limb darkening parameter.
+      Further limb darkening parameters can be specified as needed by the specified model.
```
taylorbell57, line 528: I would document the limb darkening parameters differently. I would make it clearer that there are no u parameters for 'uniform', just u1 for 'linear', u1-2 for 'quadratic', u1-2 for 'kipping2013', 'square-root', 'logarithmic', 'exponential', and u1-4 for '4-parameter'

docs/source/ecf.rst
```diff
@@ -540,16 +497,48 @@ The path to the directory in which to output the Stage 5 JWST data and plots.
...
+   - Systematics Parameters - Depending on the model specified in the Stage 5 ECF, set either polynomial model coefficients ``c0--c9`` for 0th to 3rd order polynomials. Polynomial fits should include at least ``c0`` for usable results. Exponential ramp models can fit up to two ramps. ``r0--r2`` for one ramp, ``r3--r5`` for the second.
```
taylorbell57, line 529: I would explain what the coefficients each mean.

For the polynomial, the coefficients are in the reverse order of numpy.poly1d documented [here](https://numpy.org/doc/stable/reference/generated/numpy.poly1d.html) and the x-values for numpy.polyval (documented [here](https://numpy.org/doc/stable/reference/generated/numpy.polyval.html)) are the time with respect to the mean of the time array

For the ramp model, use the definitions below:
https://github.com/kevin218/Eureka/blob/06c6df01a4cf911d537f88af0f1ace97eb5b6974/eureka/S5_lightcurve_fitting/models/ExpRampModel.py#L71

where time_local is the time with respect to the first frame.

docs/source/outputs.rst
```diff
@@ -0,0 +1,145 @@
...
+   - If ``isplots_S5`` = 3: ``Eureka!`` will plot an RMS deviation plot for each channel to help check for correlated noise, plot the normalized residual distribution, and plot the fitting chains for each channel.
+   .. figure:: ../media/S5_out/fig5501.png
+      :alt: Stage 5 RMS deviation plot
+
+      Stage 5 output: RMS Deviation Plot
+   
+   .. figure:: ../media/S5_out/fig5501.png
+      :alt: Stage 5 residual distribution
+
+      Stage 5 output: Residual Distribution
```
taylorbell57, line 126: One of these must be referring to the wrong file

docs/source/outputs.rst
```diff
@@ -0,0 +1,145 @@
...
+
+   .. figure:: ../media/S5_out/fig5403.png
+      :alt: Stage 5 fitting chains
+
+      Stage 5 output: Fitting Chains
```
taylorbell57, line 131: There are two different chain plots - one with the burn-in steps and one without the burn-in steps. This plot is only made for emcee

docs/source/outputs.rst
```diff
@@ -0,0 +1,145 @@
...
+
+   - If ``isplots_S5`` = 5: ``Eureka!`` will plot a corner plot for each channel.
```
taylorbell57, line 133: Only for emcee or dynesty

taylorbell57 (changes requested):

kevin218:
Regarding the figure numbers, I agree with adding another digit to indicate the 'isplots' number in S4 and S5.

docs/source/ecf.rst
```diff
@@ -523,9 +523,7 @@ This file describes the transit/eclipse and systematics parameters and their pri
       - ``AmpSin2`` - Amplitude of the second sine
    - Limb Darkening Parameters
       - ``limb_dark`` - The limb darkening model to be used. Options are: ``['uniform', 'linear', 'quadratic', 'kipping2013', 'square-root', 'logarithmic', 'exponential', '4-parameter']``
-      - ``u1`` - The first limb darkening parameter
-      - ``u2`` - The second limb darkening parameter.
-      Further limb darkening parameters can be specified as needed by the specified model.
+      - ``uniform`` limb-darkening has no parameters, ``linear`` has a single parameter ``u1``, ``quadratic`` and ``kipping2013`` have two ``u1, u2``, ``square-root, logarithmic, exponential, 4-parameter`` all have four parameters ``u1, u2, u3, u4``
```
taylorbell57, line 526: ``square-root, logarithmic, exponential`` also only have ``u1, u2``

docs/source/outputs.rst
jbrande: Text fixed, I'll add both plots later tonight.

docs/source/ecf.rst
jbrande: Done.

docs/source/outputs.rst
jbrande: Fixed.

docs/source/ecf.rst
jbrande: I added more text to describe this.

docs/source/ecf.rst
jbrande: Done.

docs/source/ecf.rst
jbrande: Fixed.

docs/source/installation.rst
```diff
@@ -97,7 +97,7 @@ Issues with installing the jwst dependency
 ------------------------------------------
 
 If you have issues installing the jwst dependency, check out the debugging advice related to the jwst package on our
-`FAQ page <https://eurekadocs.readthedocs.io/en/latest/installation.html#issues-installing-or-importing-jwst>`_.
+`FAQ page <https://eurekadocs.readthedocs.io/en/latest/faq.html#issues-installing-or-importing-jwst>`_.
```
taylorbell57, line 100: Sorry, last request. Can you update this to a reference similar to how we did this elsewhere like:
```
:ref:`Stage 3 Output <s3-out>`
```

taylorbell57:
> Regarding the figure numbers, I agree with adding another digit to indicate the 'isplots' number in S4 and S5.

I've just opened a new issue regarding this (#281) - we can address that with a separate PR

docs/source/installation.rst
jbrande: Done.

jbrande:
@taylorbell57  I think everything's covered now.

docs/source/outputs.rst
```diff
@@ -0,0 +1,148 @@
...
+
+In Stage 4:
+   - If ``isplots_S4`` = 1: ``Eureka!`` will plot the spectral drift per exposure, and the drift-corrected 2-dimensional lightcurve with extracted bins overlaid.
+
+   .. figure:: ../media/S4_out/fig4101.png
+      :alt: Stage 4 spectral drift plot
+
+      Stage 4 output: Spectral Drift Plot
+
+   .. figure:: ../media/S4_out/fig4201.png
```
taylorbell57, line 79: Are you able to get a newer copy of this figure? The formatting has changed since you made this one

taylorbell57: Once that's edited though, I think everything looks great!

taylorbell57 (approved):

PR #257, opened 2022-04-04
Fixing reduced chi-squared calculation

### LINKED ISSUE #253: Stage 5 reduced chi-squared calculation is messed up
For some reason the Stage 5 reduced chi-squared calculation is now messed up. When fitting the noise level using the scatter_mult parameter and using the dynesty sampler, I get a very reasonable looking fit with reasonable looking error bars, and my fig550_res_distri plot looks reasonable, but I get enormous reduced chi-squared values (e.g. 465.03078651787916 for MIRI's longest wavelength bin which used to give 1.0049324562175084 for basically the exact same ending lnprob). Not sure what could be causing that issue

taylorbell57:
Previously the GPeval function would return all ones if a GP was not being used. Because the GP model is additive rather than multiplicative, this messed up the reduced chi-squared calculation but other parts of the code did not end up using the GPeval function so they were fine.

Resolves #253.

### DIFF
--- a/eureka/S5_lightcurve_fitting/models/Model.py
+++ b/eureka/S5_lightcurve_fitting/models/Model.py
@@ -298,9 +298,10 @@ class CompositeModel(Model):
         # Get the time
         if self.time is None:
             self.time = kwargs.get('time')
-            
-        flux = np.ones(len(self.time)*self.nchan)
-        
+
+        # Set the default value
+        flux = np.zeros(len(self.time)*self.nchan)
+
         # Evaluate flux
         for model in self.components:
             if model.modeltype == 'GP':


kevin218 (commented):
Is it possible to have GPeval returns ones and have GPs be multiplicative (instead of additive) to keep in line with all of the other model functions?

taylorbell57:
In my experience I've always done additive GPs, but perhaps they can be multiplicative - what are your thoughts @evamariaa

taylorbell57:
Summarizing our discussion from our Zoom call earlier, we've all only ever seen and used additive GPs but can't say for sure whether they strictly must be additive. I confirmed that we don't need to worry about order of operations - the GP model will always be added as a last step. We decided for now that we should just leave the GP model as additive (with a note in the documentation or something stating that the GP model is additive) and can investigate a multiplicative GP in the future. Given that, I believe this PR should be ready for approval

kevin218 (approved):

PR #259, opened 2022-04-04
Accounting for light-travel time

### LINKED ISSUE #150: Batman model doesn't account for light travel time
The Batman model doesn't account for light travel time, which is likely important for JWST observations. We likely want to consider switching to spiderman for eclipse models and/or phase variations.

taylorbell57:
I have now added some code based on the calculation done in SPIDERMAN using some custom code that I wrote for the Bell_EBM package which offers far faster solutions to Kepler's equation when you are using many time points (based on the algorithm of [Tommasini & Olivieri (2018)](https://arxiv.org/abs/1812.02273) rather than the standard Newtonian iteration method). I see that this paper wasn't published until 2020 ([paper1](https://doi.org/10.3390/math8112017) and [paper2](https://doi.org/10.1016/j.amc.2019.124677)), and I haven't checked to see if there are any differences in their algorithm since I implemented it in 2018. But my tests back in 2018 showed that it performed excellently and for >7 time points it performed substantially faster than the Newton method, so I have an if/else statement in my code to switch between methods based on the number of time points.

I find that for a single fit of the full-orbit MIRI phase curve with dynesty I get a sampling time of 1m35s when accounting for the light-travel time in the eclipse model rather than 1m30s when neglecting it, so I think it is fine to have the adjustment for light-travel time be the default behaviour but I can add the ability to turn it off if desired. The light-travel time isn't accounted for in any of the other models (transit or sinusoidal phase curve), but I can add that if there might be an interest (though I doubt it). For NGTS-10 b, the light travel time difference between transit and eclipse is only ~14 seconds, so I can't really compare the differences in log-likelihood, but I can definitely say it doesn't have any strongly negative impacts or anything.

I also moved the tests folder outside of the eureka folder as I believe that it is supposed to be in the main Eureka directory and VS Code seemed to require it be there in order to let me automatically run all the tests with the click of a button.

Closes #150.

I also merged PR #257 into this branch to properly test this code, so that other PR should be merged first.


eureka/S5_lightcurve_fitting/models/Model.py
```diff
@@ -298,9 +299,10 @@ def GPeval(self,  fit, **kwargs):
         # Get the time
         if self.time is None:
             self.time = kwargs.get('time')
-            
-        flux = np.ones(len(self.time)*self.nchan)
-        
+
+        # Set the default value
+        flux = np.zeros(len(self.time)*self.nchan)
```
kevin218, line 304: Reminder: Open question on what to do here.

kevin218 (commented):
Once #257 is approved, I can approve this one.

eureka/S5_lightcurve_fitting/models/Model.py
taylorbell57: As decided over Zoom, we will leave the GP as additive for now and so this line is fine as is

kevin218 (dismissed):

kevin218 (approved):

PR #275, opened 2022-04-13
Remove jwst package dependency in miri.py by hardcoding wavelengths

sebastian-zieba:
This Pull Request addresses Issue #236.
The old function _wave_MIRI(filename)_ which used the jwst package was removed.
It was replaced with _wave_MIRI_hardcoded()_ which includes the full wavelength array.

The reasoning behind this is that _hdulist['WAVELENGTH', 1].data_ is populated by zeros for the simulated data. 

When the actual data with a proper wavelength array is available, line 55 in miri.py has to be just changed to:

data.wave = hdulist['WAVELENGTH', 1].data


eureka/S3_data_reduction/miri.py
```diff
@@ -57,7 +52,7 @@ def read(filename, data, meta):
     if meta.firstFile:
         print('  WARNING: The wavelength for the simulated MIRI data are currently hardcoded '
               'because they are not in the .fits files themselves')
-    data.wave = np.tile(wave_MIRI(filename),(data.data.shape[2],1))[:,::-1]    # hdulist['WAVELENGTH', 1].data
+    data.wave = np.tile(wave_MIRI_hardcoded(),(data.data.shape[2],1))[:,::-1]    # hdulist['WAVELENGTH', 1].data
```
taylorbell57, line 55: Could you add a check to see if `np.all(hdulist['WAVELENGTH`, 1].data == 0)`, and do this hardcoded method and print the warning if so or just set `data.wave = hdulist['WAVELENGTH`, 1].data` and don't print the warning if it is non-zeros? That would help us in the transition period between simulated and real data.

While you're at it, you could add a similar test for the data.time array below where you check if `len(int_times['int_mid_BJD_TDB']) == 0` (True for simulated data, False for real data).

taylorbell57 (changes requested):

eureka/S3_data_reduction/miri.py
sebastian-zieba: I agree. Sure, I'll do that!

sebastian-zieba:
I added the checks for the wavelength and time arrays. 

If `np.all(hdulist['WAVELENGTH', 1].data == 0)` --> use hardcoded wavelength array and output a warning
else --> use hdulist['WAVELENGTH', 1].data and do swapaxis which is needed for the miri data

if len(int_times['int_mid_BJD_TDB']) == 0 --> use the coded time arrays and output a warning
else --> use int_times['int_mid_BJD_TDB']

taylorbell57 (approved):

PR #282, opened 2022-04-15
Connecting Stages 1 and 2

### LINKED ISSUE #125: Connect Stage 1 to Stage 2
Currently Stage 1 does not connect to Stage 2, so the two cannot be called sequentially and users need to specify the exact output folder from Stage 1. Some small tweaks to Stage 2 (mirroring the code in Stages 3+) will allow a more fluid transition

taylorbell57:
Resolves #125

Stage 2 will now work if Stage 1 and Stage 2 are being run sequentially.
If they are not, the user should also not need to update the inputdir
as the code will try to find the Stage1 outputdir automatically


eureka/S2_calibrations/s2_calibrate.py
```diff
@@ -141,6 +161,160 @@ def calibrateJWST(eventlabel, ecf_path='./', s1_meta=None):
...
+        s1_outputdir += '/'
+    s1_topdir = s1_meta.topdir
+    
+    # Load S2 Eureka! control file and store values in the S1 metadata object
+    ecffile = 'S2_' + meta.eventlabel + '.ecf'
+    meta = s1_meta
+    meta.read(ecf_path, ecffile)
+
+    # Overwrite the inputdir with the exact output directory from S1
+    meta.inputdir = os.path.join(s1_topdir, s1_outputdir)
```
kevin218, line 310: I don't like this implementation, both here and at other stages. The inputdir should use s2_inputdir from the S2 ECF file. Otherwise, there's no way to read in files that were moved or from a renamed directory.

eureka/S2_calibrations/s2_calibrate.py
```diff
@@ -141,6 +161,160 @@ def calibrateJWST(eventlabel, ecf_path='./', s1_meta=None):
...
+    meta.old_datetime = meta.datetime # Capture the date that the
```
kevin218, line 311: Comment is missing some words

eureka/S2_calibrations/s2_calibrate.py
```diff
@@ -65,16 +66,34 @@ def calibrateJWST(eventlabel, ecf_path='./', s1_meta=None):
...
+    if s1_meta != None:
+        meta = load_general_s1_meta_info(meta, ecf_path, s1_meta)
+    else:
+        # Attempt to find subdirectory containing S1 FITS files
+        meta = find_s1_files(meta)
+
+        # Create directories for Stage 2 processing outputs
+        meta.inputdir_raw = meta.inputdir
+        meta.outputdir_raw = meta.outputdir
+        meta.inputdir = os.path.join(meta.topdir, *meta.inputdir_raw.split(os.sep))
```
kevin218, line 82: It would probably be cleaner to never join topdir and inputdir/outputdir, so that they don't need to be split later. Instead, we should combine the two strings only for the actual read/write step.

eureka/lib/readECF.py
```diff
@@ -250,7 +250,11 @@ def copy_ecf(self):
                 else:
                     line_segs = line.strip().split()
                     if line_segs[0]=='inputdir':
-                        new_file.write(line_segs[0]+'\t\t/'+self.inputdir+'\t'+' '.join(line_segs[2:])+'\n')
+                        if self.topdir in self.inputdir:
```
kevin218, line 253: This code becomes unnecessary with my previous suggestion.

eureka/S2_calibrations/s2_calibrate.py
taylorbell57: We can discuss this on Monday, but the issue is that I think we should allow autocompleted paths based on the inputdir. For example, it is far more convenient to specify `inputdir = Stage1` and have Stage 2 find the newest directory in that folder. Before this PR, one would have to run Stage 1, see what foldername the FITS files ended up being put into, copy-paste that foldername into their Stage 2 ECF, and then run Stage 2 - an annoying step for Stage 1->2, but something that would get super tedious at other stages where many folders might be made.

A better example is Stage 3->4, where we wanted to keep the ability for Stage 4 to consider all of the apertures used in Stage 3. So Stage 3 outputs Stage3/S3_2022-03-21_miri_lrs_ap2_bg4_run1 and Stage3/S3_2022-03-21_miri_lrs_ap2_bg14_run1 as an example, and then Stage 4 somehow needs to find those two particular folders (with potentially many other folders inside the Stage3 folder from previous attempts). The code that we have now will look for Meta_Save.dat files in Stage 4, find none, look inside all the folders inside the Stage4 folder and find two Meta_Save.dat files (or many more if there are older runs) and by default will use the most recently created/edited one. But if you set `allapers=True` in Stage4, the code needs to find the other apertures that were used in that same Stage 3 run, so it uses the one Stage 3 Meta_Save file that it does have to figure out where the other apertures would have been saved.

It's all kind of a mess I know, but given the I/O setup we have presently it was the best I could come up with and has been super convenient as long as you're not moving the files to a different drive or sharing them with someone else.  While somewhat uncommon, I agree it's important to support the ability to move the files around at will, and I think we should take the opportunity to rework all the connections between Stages at the same time as we're changing the pipeline's I/O. One option might be putting all the apertures for a Stage3 run into one folder (e.g. Stage3/S3_2022-03-21_miri_lrs_run1/ap2bg4 and Stage3/S3_2022-03-21_miri_lrs_run1/ap2bg14).

Either way, edits to these functions is going to significantly affect all the stages, and I think doing that at around the same time as changing the I/O to Astreus would make for a clean transition. This was just meant as a quick and easy fix to the Stage 1->2 connection to make it consistent with all the other stages for the time being.

eureka/S2_calibrations/s2_calibrate.py
taylorbell57: The issue is that we then need to duplicate the os.path.join code tens of times for every read/write function and plotting function. My idea had been to have inputdir_raw and outputdir_raw be inputdir and outputdir without the topdir joined on so that we still have them if needed.

eureka/lib/readECF.py
taylorbell57: This code could also be unnecessary if inputdir_raw (as described in my previous comment) is correctly defined for all stages - I'm just not sure that it is for all of Stages 1-6

eureka/S2_calibrations/s2_calibrate.py
kevin218: Ok, let's discuss on Monday.  I agree that we need an overhaul, in part because I frequently rename my folders to be more descriptive than simply "runX".  Your proposed solution (Stage3/S3_2022-03-21_miri_lrs_run1/ap2bg4) should work for both our needs, so long as meta.inputdir is derived from the S4 ecf file (e.g., inputdir=Stage3/S3_2022-03-21_miri_lrs_run1).

eureka/S2_calibrations/s2_calibrate.py
kevin218: Ok, I'm fine with going that route but I suggest we rename our variables to avoid confusion.  For example, define meta.s3_topinputdir = os.path.join(meta.s3_topdir, meta.s3_inputdir).  Let me know if I'm wrong here, but I don't think there's any reason to be overwriting meta.inputdir at every stage.  Not sure if having a different topdir for each stage is useful, so that's something we can discuss.

kevin218: Adding to my last comment, this may mean passing `meta.s3_topinputdir` as a variable to `readECF.py`.

taylorbell57: Are you fine with leaving this as I've written it for now and then this will all get significatly overhauled in a few days?

eureka/lib/readECF.py
taylorbell57: Again, are you okay with leaving this as I've written it for now? This currently fixes a bug with Stages 3-6 where the copied ECF's inputdir was previously the topdir+inputdir by accident

kevin218: Yes

kevin218 (approved):

taylorbell57:
Sorry, I realized that this branch hadn't been synced with main in a while, and some of my edits were going to conflict with Aarynn's in a way that GitHub wasn't going to realize. While I was addressing that, I noticed that the Stage 1 test for MIRI was bugged because the small file had too few integrations for jwst's Stage 2 (a known bug with that package). I couldn't fix that without uploading a 2GB uncal file or going through the work of making my own uncal file which I didn't want to do right now, so I just commented out the S1 test for MIRI

kevin218 (approved):

PR #287, opened 2022-04-18
Improve multiprocessing speed during S3, issue #97

### LINKED ISSUE #97: Running S3 with multiple threads seems to be slower than one thread
When testing the new jwst v1.3.3 I noticed that the background subtraction step in S3 ([here](https://github.com/kevin218/Eureka/blob/fc3c4c9cf972b9fc1896133af708349edc54c6b9/eureka/S3_data_reduction/background.py#L65-L73)) was requiring 6 seconds per iteration with 14 cpu threads, but when I tried to run it with only 1 cpu thread it only took 2 seconds per iteration. I can imagine the runtime not scaling linearly with the cpu threads, but running slower seems very unintuitive and is obviously a bad sign. We'll need to see if this is related to my other multithreading issue (#92) or if the there is a larger issue with multithreading here.

AarynnCarter:
See issue #97 for further detail. In short, the data array being passed to the S3 multiprocessing was too large and caused the code the slow down instead of speed up, this PR fixes this by allocating a single integration to each process instead of the full dataset. 

Only thing missing is that this is currently not compatible with NIRISS, but it should be able to be merged as is anyway. 

Will close #97


AarynnCarter:
Okay I don't think I made this pull request correctly... let me resolve a couple of conflicts in the files.

AarynnCarter:
Okay that should be good to go now!

eureka/S3_data_reduction/background.py
```diff
@@ -76,14 +83,27 @@ def writeBG(arg):
...
+                writeBG(inst.fit_bg(data.subdata[n], data.submask[n], n, meta, isplots))
     else:
         # Multiple CPUs
         pool = mp.Pool(meta.ncpu)
         args_list = []
-        for n in range(meta.int_start,meta.n_int):
-            args_list.append((data, meta, n, isplots))
-        jobs = [pool.apply_async(func=inst.fit_bg, args=(*args,), callback=writeBG) for args in args_list]
+
+        # Todo, convert WFC3/NIRISS fit_bg to only accept individual frames (see nircam and below for example)
```
taylorbell57, line 97: Can you edit this comment to only be about NIRISS and not WFC3

eureka/S3_data_reduction/hst_scan.py
```diff
@@ -537,6 +537,7 @@ def correct_slitshift2(data, slitshift, mask=None, isreverse=False):
         return cordata
 
 # Calulate drift2D
+import image_registration as imr
```
taylorbell57, line 540: This is already imported above and can be removed here

eureka/S3_data_reduction/miri.py
```diff
@@ -224,9 +224,10 @@ def flag_bg(data, meta):
     return nircam.flag_bg(data, meta)
 
 
-def fit_bg(data, meta, n, isplots=False):
+def fit_bg(dataim, datamask, n, meta, isplots=False):
     '''Fit for a non-uniform background.
 
     Uses the code written for NIRCam and untested for MIRI, but likely to still work (as long as MIRI data gets rotated)
```
taylorbell57, line 230: While you're at it, do you want to remove the comment saying untested for MIRI, and just say:

> Uses the code written for NIRCam which works for MIRI as long as MIRI data gets rotated

eureka/S3_data_reduction/nirspec.py
```diff
@@ -99,9 +99,9 @@ def flag_bg(data, meta):
 
     return data
 
-def fit_bg(data, meta, n, isplots=False):
+def fit_bg(dataim, datamask, n, meta, isplots=False):
     '''Fit for a non-uniform background.
 
     Uses the code written for NIRCam and untested for NIRSpec, but likely to still work
```
taylorbell57, line 105: Do you also want to edit this comment

taylorbell57 (changes requested):

AarynnCarter:
@taylorbell57 Okay, all your comments are resolved.

taylorbell57 (approved):

PR #299, opened 2022-04-20
Renaming S3-5 main functions

### LINKED ISSUE #256: Renaming of main S1-5 functions
As of right now, the main functions of S1-5 have names that end in JWST, but Stages 3-6 are all compatible with HST/WFC3. I suggest we rename the main S1-5 functions to be more general. Suggestions are below

s1.rampfitJWST -> s1.rampfit
s2.calibrateJWST -> s2.calibrate
s3.reduceJWST -> s3.reduce
s4.lcJWST -> s4.getlc
s5.fitJWST -> s5.fit

taylorbell57:
Resolves #256

### DIFF
--- a/demos/HST/run_eureka_wfc3.py
+++ b/demos/HST/run_eureka_wfc3.py
@@ -9,10 +9,10 @@ eventlabel = 'wfc3'
 ecf_path = './'
 
 if __name__ == '__main__':
-	s3_meta = s3.reduceJWST(eventlabel, ecf_path=ecf_path)
+	s3_meta = s3.reduce(eventlabel, ecf_path=ecf_path)
 
-	s4_meta = s4.lcJWST(eventlabel, ecf_path=ecf_path, s3_meta=s3_meta)
+	s4_meta = s4.getlc(eventlabel, ecf_path=ecf_path, s3_meta=s3_meta)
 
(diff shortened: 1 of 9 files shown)


demos/HST/run_eureka_wfc3.py
```diff
@@ -9,10 +9,10 @@
 ecf_path = './'
 
 if __name__ == '__main__':
-	s3_meta = s3.reduceJWST(eventlabel, ecf_path=ecf_path)
+	s3_meta = s3.reduce(eventlabel, ecf_path=ecf_path)
 
-	s4_meta = s4.lcJWST(eventlabel, ecf_path=ecf_path, s3_meta=s3_meta)
+	s4_meta = s4.getlc(eventlabel, ecf_path=ecf_path, s3_meta=s3_meta)
```
kevin218, line 14: For consistency, maybe we should stick with `genlc`?

taylorbell57: Or we could change the name of the file - whichever you prefer

kevin218: I'd rather not change the directory structure, docs, etc. Let's use genlc.

kevin218 (approved):

PR #302, opened 2022-04-21
Add celerite as a gp package and abillity for shared parameters

evamariaa:
- Some changes to parameters and how they are read in to be able to support shared parameters within the GP. 
- Added the gp package celerite and runs fine on individual channels, but I wasn't able to fully run with shared fits so it might be buggy still. 
- Also fixed that the lc.unc_fit is used within the GP model as the noise instead of lc.unc. 
- (There have also been some changes to the tinygp code but it doesn't work just yet so I need to do this for the next release.)


taylorbell57:
I took a quick look through this but didn't dive too deep into any of it, and things generally seem reasonable to me. I'll try to read through it more thoroughly later unless Kevin beats me to it

kevin218 (approved):

PR #305, opened 2022-04-25
Changing the way inputdir is auto completed

taylorbell57:
Alright, I've now cleaned up the way that stages are connected, so we no longer have weird behaviours if folders get renamed between stages. To do this, I moved the different aperture folders from Stages 3+ into a single folder for that stage (e.g. Stage3/S3_2022-04-25_miri_run1/ which contains the folders ap4_bg5 and ap4_bg10, and similarly for Stages 4-6).

I also wanted to be able to try Eureka! on Windows which uses back slashes instead of forward slashes, so I updated all the slashes to os.sep which figures out whether to use forward or back slashes.

I've run the code through all the tests successfully and manually tested it as well.

### DIFF
--- a/eureka/lib/manageevent.py
+++ b/eureka/lib/manageevent.py
@@ -289,12 +289,16 @@ def mergeevents(new_meta, old_meta):
     - April 25, 2022 Taylor Bell
         Initial version.
     """
-    # Load current MetaClass into old MetaClass
+    # Load current ECF into old MetaClass
+    old_meta.read(new_meta.folder, new_meta.filename)
+
+    # Load any missing parameters from current MetaClass into old MetaClass
     for key in new_meta.__dict__:
-        setattr(old_meta, key, getattr(new_meta, key))
-    new_meta = old_meta
+        if key not in old_meta.__dict__:
+            setattr(old_meta, key, getattr(new_meta, key))
 
-    # Make sure new_meta's inputdir_raw is correct
-    new_meta.inputdir_raw = new_meta.inputdir[len(new_meta.topdir)]
+    # Make sure inputdir is correct
+    old_meta.inputdir = new_meta.inputdir
+    old_meta.inputdir_raw = new_meta.inputdir_raw
 
-    return new_meta
+    return old_meta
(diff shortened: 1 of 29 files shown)


taylorbell57:
Note that this will break backwards compatibility with all Stage 3+ outputs as the folder structure is significatly different, and I didn't bother to add a whole bunch of code to maintain backwards compatibility since the Astraeus edits are going to break backwards compatiblity anyway

eureka/lib/manageevent.py
```diff
@@ -190,3 +192,113 @@ def updateevent(event, filename, add):
...
+    Notes
+    -------
+    History:
+
+    - April 25, 2022 Taylor Bell
+        Initial version.
+    """
+    # Load current ECF into old MetaClass
+    old_meta.read(new_meta.folder, new_meta.filename)
+
```
kevin218, line 294: Should we also check for conflicts between meta parameters to see if they have different values? The function could print a warning message if the same parameter has different values between meta instances. No need to throw an error. This is not high priority, just an idea.

taylorbell57: So the main reason I did things this way is to make sure that parameters from the new ECF overwrite the old ECF parameters (e.g. topdir, inputdir, outputdir, verbose). And for each aperture pair this function will get called again to overwrite the values to those needed for the new aperture (outputdir, spec_hw, bg_hw, and potentially inputdir). I can add log statements (perhaps with mute=True) if you want, but I personally don't think they'd be worthwhile given one of the main uses is specifically to overwrite parameters in the old meta object

AarynnCarter:
@taylorbell57 Will this change break the quickstart instructions? If so I can try to set aside some time to update it before next week.

taylorbell57:
> @taylorbell57 Will this change break the quickstart instructions? If so I can try to set aside some time to update it before next week.

I don't think this should affect the quickstart at all, but Kevin's PR will definitely require there to be at least a few minor tweaks to the quickstart

eureka/lib/manageevent.py
kevin218: Ok, let's leave as is.

kevin218 (approved):

PR #321, opened 2022-05-12
Fixing issues with missing meta.tab_filename

taylorbell57:
With the changes to saving S3 files as Xarray save files, `meta.tab_filename` was removed but not all references to it were. This quick PR fixes that broken reference. I don't know why I didn't notice that before with pytest, but perhaps your branch wasn't up-to-date with main and git did something undesireable when it merged the branches.


kevin218 (approved):
Maybe metadata gets passed to the function during testing instead of restored from the save file.

taylorbell57:
No, I encountered the error during testing this time around, so not sure why I didn't last time

PR #322, opened 2022-05-13
Fixing docstrings and making everything flake8 compliant

taylorbell57:
It seems that flake8 is the standard linting tool used for python to make sure everything is nicely formatted in a standard way, and when I first used flake8 in VS Code basically every line of code had a warning or error on it. I took some time over the past few weeks to clean up the code to make it all flake8 compliant, and now if you run the following command in the Eureka folder you'll get no errors/warnings besides a small number of errors that will take a not-insignificant amount of effort to resolve:
```
flake8 --exclude="./build/**/*,./docs/**/*,.vscode/*,**/site-packages/**/*,*.c,*.pyx,**/*niriss*.py" --per-file-ignores="**/__init__.py:F401,**/manageevent.py:F841,./demos/**/run_*.py:E402,**/test_*.py:E402" .
```
In those flake8 arguments I've ignored the NIRISS files (which I know Adina is going to change significantly) and a few warnings/errors that shouldn't apply to some specific files (e.g. the `__init__.py` files should have unused imports).

While I was at it, I also tried to edit all the docstrings to get them into a uniform format, and I now get no warnings or errors while trying to build the docs (before there were hundreds).

While there are a tonne of edited files, **I have tried my very best to make no changes at all to how the code actually works** and instead just focused on reformatting to make things flake8 compliant, so I'd suggest that you don't need to review every file and can instead just skim some to see what I've done.

Moving forward, I want to add the above flake8 settings to the setup.cfg file that we'll end up making for v0.1, and then everybody can run flake8 in addition to pytest before submitting a PR.


kevin218:
I think I'll follow your suggestion!

eureka/S3_data_reduction/nircam.py
```diff
@@ -41,72 +41,104 @@ def read(filename, data, meta):
...
-    data.attrs['intend']   = data.attrs['mhdr']['INTEND']
+    data.attrs['intend'] = data.attrs['mhdr']['INTEND']
 
-    sci     = hdulist['SCI',1].data
-    err     = hdulist['ERR',1].data
-    dq      = hdulist['DQ',1].data
-    v0      = hdulist['VAR_RNOISE',1].data
-    wave_2d = hdulist['WAVELENGTH',1].data
-    int_times = hdulist['INT_TIMES',1].data[data.attrs['intstart']-1:data.attrs['intend']]
+    sci = hdulist['SCI', 1].data
```
kevin218, line 49: Removing the aligned `=` makes me sad!

eureka/S3_data_reduction/sigrej.py
```diff
@@ -1,63 +1,59 @@
...
-    fmedian:  ndarray
-      (returned) the median of the accepted data.
-    fmedstddev: ndarray
-      (returned) the standard deviation of the accepted data with respect to the median.
+    data : ndarray
+        Array of points to apply sigma rejection to.
+    sigma : ndarray (1D)
+        1D array of sigma values for each iteration of sigma rejection.
+        Number of elements determines number of iterations.
+    mask: byte array; optional
```
kevin218, line 16: So required parameters need a space before the colon and optional input parameters don't (`data :` vs `mask:`)?

eureka/lib/centroid.py
```diff
@@ -1,163 +1,163 @@
 # $Author: patricio $
```
kevin218, line 1: For consistency, maybe we should delete all instances of this outdated method of version control?

eureka/S3_data_reduction/sigrej.py
taylorbell57: No, sorry - this was just a result of me merging in your Astraeus PR after having edited most of the code already and missing the occasional line from your edits. I'll try to read through everything again (or even better, try to find a regex automated way of searching for things)

taylorbell57: Actually there were several places I'd missed, and regex came to save the day.

taylorbell57:
Alright, I addressed your comments, and found some more places where docstrings hadn't been edited properly. I also removed the old MetaClass definitions which were there for backwards compatibility, but we've already thoroughly broken that, so may as well remove those old MetaClass definitions now too. I also broke backwards compatibility with old S5 chains which were saved as csv files, as we have long been saving them as h5 files.

I also moved the lib folder to the bottom of the API page as it is the least well documented and longest section in the API, so I figured let's not put it at the start of the API at least for now. I also noticed that the S5 models folder was missing from the API, so I added those files.

kevin218 (approved):

PR #331, opened 2022-05-23
Added tiny NIRISS dataset

lpcoulombe:
Added stage 1 outputs of a tiny NIRISS dataset (3 integrations) in Eureka/tests/data/JWST-Sim/NIRISS/Stage1 .


kevin218 (approved):

kevin218:
I just realized these are `rampfitstep` files, but we need `rateints` files.

lpcoulombe:
The rampfitstep files are indeed the `rateints` and `rate` files. I have not ran the tiny dataset through stage 2 but I can do it in the next few days. I will change the files' position at the same time.

taylorbell57:
The files are in the correct location, but I am confused about why there are two files and why they have rampfitstep suffixes rather than rateints. Is this just how NIRISS files get processed by the `jwst` pipeline?

kevin218:
Yes, NIRISS is a rebel that doesn't like to follow the rules, thus making our lives more difficult.
The first file appears to contain a single integration, which breaks the pipeline, while the second contains 3 integrations.  
As far as I can tell, there's no reason to have Eureka/eureka/tests, which is where I found the files.

taylorbell57:
Ah indeed, I misread the path in the Files Changed section on the PR

PR #336, opened 2022-06-03
fitters.py bug fix

meganmansfield:
Fixed a typo in calculating the autocorrelation time with emcee that made it never output an autocorrelation time


taylorbell57 (approved):
Aha, good catch!!

PR #338, opened 2022-06-03
Bug fixes and new features

### LINKED ISSUE #320: Allow model zorder to be changed in S6
I was biased the same way, but the problem is that the models will tend to be much higher resolution than the data (at least in my case), and as a result the data gets hidden. I'll open a new issue to allow the data to be plotted on top of or below the model

_Originally posted by @taylorbell57 in https://github.com/kevin218/Eureka/pull/317#discussion_r871673726_

### LINKED ISSUE #337: [Bug]: NaN uncertainties going into Stage 5
### FAQ check

### Instrument

### What happened?

For non-shared fits, we do not currently use np.nanmean to normalize the light curves in Stage 5, so if sigma clipping is done all the uncertainties become NaNs. The required change was already made to the shared fits code, but not the unshared fits code. I'll patch this right now.

### Error traceback output

_No response_

_No response_

_No response_

_No response_

### Code of Conduct

taylorbell57:
This PR resolves #320 (allow setting of S6 model zorder) and resolves #337 (NaN uncertainties going into S5 fits). It also updates now out-of-date Stage 1 code that was broken when files were updated in the lib folder - we didn't catch that earlier since we don't have tests for Stage 1. I've also more generally fixed the `EurekaSpec2Pipeline has no attribute master_background` issue by first checking that such a step exists - this will also resolve the same issue Adina found for NIRISS but do so for all instruments (since I just encountered it on MIRI data). I also fixed a bug in Stage 4 clipping where `lc.wave_mid.attrs['wave_units']` was undefined.

I also now made it so that the bin edges are always plotted in Fig 4101 while they were previously only plotted if there was more than one bin. However, since the bin edges are not necessarily aligned with the edges of the figure, the lines are still useful even if there is only one bin (for a white-light fit).

I also allowed a `meta.time_file` setting in Stage 3 to allow users to pass in a file with the time array they want to use instead of the one inside the FITS files. For the MIRI simulations we've had to directly edit the code for each new simulation, and this will now remove that requirement and also allow users to use their own time array in case there are issues with the values from the FITS headers.

I also moved the NIRISS testing files to the right location.

### DIFF
--- a/demos/JWST/S6_template.ecf
+++ b/demos/JWST/S6_template.ecf
@@ -30,7 +30,7 @@ model_spectrum  None # Path to the model spectrum relative to topdir. Set to Non
 model_x_unit    um # Options include any measurement of light included in astropy.units.spectral (e.g. um, nm, Hz, etc.)
 model_y_unit    Rp/Rs # Options include Rp/Rs, (Rp/Rs)^2, Fp/Fs
 model_y_scalar  1 # Indicate whether model y-values have already been scaled (e.g. write 1e6 if model_spectrum is in ppm)
-model_zorder    0 # The zorder of the model on the plot (0 for beneath the data, np.inf for above the data)
+model_zorder    0 # The zorder of the model on the plot (0 for beneath the data, 1 for above the data)
 model_delimiter None # Delimiter between columns. Typical options: None (for space), ',' for comma
 
 # Directories relative to topdir
--- a/demos/JWST/S6_template.ecf
+++ b/demos/JWST/S6_template.ecf
@@ -30,6 +30,7 @@ model_spectrum  None # Path to the model spectrum relative to topdir. Set to Non
 model_x_unit    um # Options include any measurement of light included in astropy.units.spectral (e.g. um, nm, Hz, etc.)
 model_y_unit    Rp/Rs # Options include Rp/Rs, (Rp/Rs)^2, Fp/Fs
 model_y_scalar  1 # Indicate whether model y-values have already been scaled (e.g. write 1e6 if model_spectrum is in ppm)
+model_zorder    0 # The zorder of the model on the plot (0 for beneath the data, np.inf for above the data)
 model_delimiter None # Delimiter between columns. Typical options: None (for space), ',' for comma
 
 # Directories relative to topdir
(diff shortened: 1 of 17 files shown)


demos/JWST/S6_template.ecf
```diff
@@ -30,6 +30,7 @@ model_spectrum  None # Path to the model spectrum relative to topdir. Set to Non
 model_x_unit    um # Options include any measurement of light included in astropy.units.spectral (e.g. um, nm, Hz, etc.)
 model_y_unit    Rp/Rs # Options include Rp/Rs, (Rp/Rs)^2, Fp/Fs
 model_y_scalar  1 # Indicate whether model y-values have already been scaled (e.g. write 1e6 if model_spectrum is in ppm)
+model_zorder    0 # The zorder of the model on the plot (0 for beneath the data, np.inf for above the data)
```
kevin218, line 33: np.inf seems unnecessary.  What's the zorder of the data?  If 1, for example, the user should be instructed to set model_zorder to >1 for above the data.

docs/source/ecf.rst
```diff
@@ -292,8 +292,9 @@ topdir + outputdir
 ''''''''''''''''''
 The path to the directory in which to output the Stage 3 JWST data and plots.
 
-
-
+topdir + time_file
+''''''''''''''''''
+The path to a file which contains the time array you want to use instead of the one contained in the FITS file.
```
kevin218, line 297: "The path to a file **that** contains the time array"

eureka/S3_data_reduction/nirspec.py
```diff
@@ -59,7 +60,14 @@ def read(filename, data, meta):
                                              data.attrs['intend']]
 
     # Record integration mid-times in BJD_TDB
-    if len(int_times['int_mid_BJD_TDB']) == 0:
+    if (hasattr(meta, 'time_file') and meta.time_file is not None):
```
kevin218, line 63: Rather than repeating the code multiple times (once for each instrument), let's turn this code into a function call.

demos/JWST/S6_template.ecf
taylorbell57: This was based on my past experience where the errorbars end up being at a different zorder than what I'd expected, so I've historically done np.inf just to make sure it works. I agree it's unnecessary though - I'll look into a more elegant solution

kevin218 (approved):
Didn't know non-integer zorders were allowed. Neat!

PR #340, opened 2022-06-06
Changed position of tiny niriss stage 1 output

lpcoulombe:
Changed the position of the tiny niriss dataset from Eureka/eureka/tests to Eureka/tests


taylorbell57:
I see that you've only deleted the files but never added them back in their new location - you might need to use a force flag when adding the files with git

lpcoulombe:
Right sorry, I ended up having to use git add -f for the file to be included in the commit. Should be there now !

jbrande (approved):
Now that the files are properly removed from the other location I think we're good.

PR #348, opened 2022-06-13
Set meta.datetime once at the start of every stage

### LINKED ISSUE #339: Expected behaviour of meta.datetime
### What happened?

As of right now, the value of meta.datetime (e.g. `2022-06-02`) is set when you first run a Eureka! stage, and for all analyses stemming from that first stage you will have the same meta.datetime (e.g. if you rerun Stage 5 on June 3rd, the output folder will still include `2022-06-02` instead of `2022-06-03`). This was an unexpected behaviour of the new `mergeevents` function introduced in PR #305, but I can't decide if it is a bug or an unintended feature, and I wanted to get your opinion. On one hand, it could be confusing if you are restarting from an older April save in June and don't see any June labelled folders, but it could also make it clearer where the inputs to that stage originated (e.g. you rerun an analysis on a run from April but also have a newer run from May, and the folder naming makes it clearer that you are resuming from that April run not the May run).

### Code of Conduct

taylorbell57:
Resolves #339

I thought about setting meta.datetime elsewhere, but I wanted to avoid
setting it multiple times per stage (e.g. once per aperture pair) in
case analyses are being run near midnight

I also renamed the imported package `time` to `time_pkg` as its name
all too often conflicts with a local variable for the observation times

### DIFF
--- a/eureka/S1_detector_processing/s1_process.py
+++ b/eureka/S1_detector_processing/s1_process.py
@@ -1,5 +1,5 @@
 import os
-import time
+import time as time_pkg
 import numpy as np
 from astropy.io import fits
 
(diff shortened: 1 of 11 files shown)


kevin218 (approved):

taylorbell57:
I just realized that the tests failed for this PR - I'll look into that and report back here when that's fixed

taylorbell57:
This PR is ready for your approval now FYI Kevin. It turned out there was a bug with the tests, not with the code itself

kevin218 (approved):

PR #355, opened 2022-06-20
Fixing bug with code coverage

taylorbell57:
I figured out the bug with the code coverage being zero after
Sebastian's PR. pytest was using the installed version of eureka rather
than using the source python files, so the code coverage could not be
tracked. This is now fixed by inserting the path to the source eureka
files at the start of the PATH variable.

Hopefully the codecov bot will comment below in ~10 minutes saying
the test coverage has gone back up to ~60%.


taylorbell57:
I can confirm that this fixes the bug with codecov (https://app.codecov.io/gh/kevin218/Eureka/pull/355), and this PR can be approved

kevin218 (approved):

PR #364, opened 2022-06-28
Update installation docs for stable release v0.1-alpha

jbrande:
I've updated the installation page on the docs to reflect that we want users to install the v0.1-alpha stable release. This was pretty simple, you just pass the release tag to the git repo address as if it's a specific branch.

I've also fixed some minor wording issues and rearranged the installation instructions to more clearly distinguish between setting up a Conda environment, then using git/pip to download and install Eureka! from downloading Eureka! and then using the environment.yml file to set up the environment.

### DIFF
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -52,14 +52,6 @@ Other specific branches can be installed using:
 	
 	pip install -e git+https://github.com/kevin218/Eureka.git@mybranchname#egg=eureka[jwst]
 
-To update your ``Eureka!`` installation to the most recent version, you can then run the following command:
-
-.. code-block:: bash
-
-	pip install --upgrade git+https://github.com/kevin218/Eureka.git#egg=eureka[jwst]
-
-
-
 In order to use any of the demo ECF files, follow the instructions in the :ref:`Demos <demos>` section of the :ref:`Quickstart <quickstart>` page.
 
 
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -2,8 +2,14 @@
 Installation
 ============
 
+Installation methods
+--------------------
+
+In order to have consistent, repeatable results across the ``Eureka!`` user community, we recommend that all general users install the most recent stable release of ``Eureka!``, v0.1-alpha. The following installation instructions are written with this in mind, and the most recent stable release is also available as a zipped archive here: https://github.com/kevin218/Eureka/releases/tag/v0.1-alpha
+
+
 Initial environment preparation
--------------------------------
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 It is **strongly** recommended that you install ``Eureka!`` in a new ``conda`` environment as other packages you've previously
 installed could have conflicting requirements with ``Eureka!``. You can install a lightweight version of conda at `this link <https://docs.conda.io/en/latest/miniconda.html>`_. Once conda is installed, you can create a
 new environment by doing:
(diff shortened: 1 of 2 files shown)


docs/source/installation.rst
```diff
@@ -13,74 +19,69 @@ new environment by doing:
...
+GitHub following the method described below.
 
+To update your ``Eureka!`` installation to the most recent version, you can then run the following command:
 
-c) With ``pip``
-~~~~~~~~~~~~~~~
+.. code-block:: bash
 
-Once in your new conda environment, you can install the ``Eureka!`` package with ``pip`` with the following command:
+	pip install --upgrade git+https://github.com/kevin218/Eureka.git#egg=eureka[jwst]
```
taylorbell57, line 62: This should be:
```
pip install --upgrade git+https://github.com/kevin218/Eureka.git@mybranchname#egg=eureka[jwst]
```

docs/source/installation.rst
```diff
@@ -13,74 +19,69 @@ new environment by doing:
...
+Other specific branches can be installed using:
 
 .. code-block:: bash
+	
+	pip install -e git+https://github.com/kevin218/Eureka.git@mybranchname#egg=eureka[jwst]
 
-	git pull
-	conda env update --file environment.yml --prune
-	pip install --no-deps --upgrade .
+If you desire any of the files in the `demos folder <https://github.com/kevin218/Eureka/tree/main/demos>`_, you will have to download these from
```
taylorbell57, line 55: Can you update this to use the instructions Aarynn wrote for the quickstart guide, as there's now a link people can click to download the demos files

docs/source/installation.rst
taylorbell57: Or use v0.1-alpha instead of mybranchname

docs/source/installation.rst
```diff
@@ -13,74 +19,69 @@ new environment by doing:
...
-~~~~~~~~~~~~~~~
+.. code-block:: bash
 
-Once in your new conda environment, you can install the ``Eureka!`` package with ``pip`` with the following command:
+	pip install --upgrade git+https://github.com/kevin218/Eureka.git#egg=eureka[jwst]
 
-.. code-block:: bash
 
-	pip install -e git+https://github.com/kevin218/Eureka.git#egg=eureka[jwst]
+Setting up a ``conda`` environment directly from the ``git`` repository
```
taylorbell57, line 65: I would name this something like "Installing with a conda environment.yml file" instead

taylorbell57 (changes requested):

docs/source/installation.rst
jbrande: Maybe we delete this bit entirely? If we don't want most people working with the potentially stable "whatever we've pushed to main most recently" version, then they should just follow the instructions in the previous sections.

docs/source/installation.rst
jbrande: I just pointed to those instructions entirely now. Actually, with the quickstart page in as good shape as it is, we might not even need the sentence here to prompt the reader to go there.

docs/source/installation.rst
taylorbell57: Sure, that probably is best

docs/source/installation.rst
taylorbell57: I think the more pointers to the Quickstart the better honestly

docs/source/installation.rst
jbrande: Done.

PR #369, opened 2022-07-06
Add ability to download HST data using MAST API

kevin218:
This PR adds two files: 

- mastDownload.py is a wrapper for astroquery.mast.Observations that is geared towards downloading and organizing exoplanet data acquired by HST.
- download_data_HST_template.py provides a demo of how to download HST data.  It is a script that can be edited and run locally.

### DIFF
--- a/demos/HST/download_data_HST_template.py
+++ b/demos/HST/download_data_HST_template.py
@@ -30,7 +30,7 @@ md.login(mast_token)
 for vis in visits:
     # Download data from MAST Archive
     result = md.download(proposal_id, vis, inst, download_dir, subgroup)
-    if result != None:
+    if result is not None:
         # Consolodate and move data into new directory
         md.consolidate(result, final_dir)
         # Sort data into science and calibration folders (scan vs direct image)
--- a/demos/HST/download_data_HST_template.py
+++ b/demos/HST/download_data_HST_template.py
@@ -1,0 +1,41 @@
+#!/usr/bin/env python3
+
+import eureka.lib.mastDownload as md
+
+# Proposal/Program ID, can be string or int
+proposal_id = '13467'
+
+# List of one or more visit numbers
+visits = [60]
+# visits = np.arange(60,85)
+
+# Instrument name, can be upper or lower case.
+# Supported options include: WFC3, STIS, COS, or FGS.
+inst = 'WFC3'
+
+# Temporary download directory will be 'download_dir'/mastDownload/...
+download_dir = '.'
+
+# FITS file type (usually IMA, sometimes FLT)
+subgroup = 'IMA'
+
+# MAST API token for accessing data with exclusive access
+mast_token = None
+
+# Final destination of files after calling mast.consolidate
+final_dir = './HD209458/ima'
+
+# If data are public, no need to call md.login() or md.logout()
+md.login(mast_token)
+for vis in visits:
+    # Download data from MAST Archive
+    result = md.download(proposal_id, vis, inst, download_dir, subgroup)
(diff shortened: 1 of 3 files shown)


demos/HST/download_data_HST_template.py
```diff
@@ -0,0 +1,41 @@
...
+
+# Final destination of files after calling mast.consolidate
+final_dir = './HD209458/ima'
+
+# If data are public, no need to call md.login() or md.logout()
+md.login(mast_token)
+for vis in visits:
+    # Download data from MAST Archive
+    result = md.download(proposal_id, vis, inst, download_dir, subgroup)
+    if result != None:
```
taylorbell57, line 33: flake8 will want this to be `if result is not None:`. You should check for the other changes flake8 requests for the non-NIRISS files (once that NIRISS PR gets merged, it'll be so much easier to tell what flake8 issues were introduced by a new PR).

demos/HST/download_data_HST_template.py
```diff
@@ -0,0 +1,41 @@
...
+
+# MAST API token for accessing data with exclusive access
+mast_token = None
+
+# Final destination of files after calling mast.consolidate
+final_dir = './HD209458/ima'
+
+# If data are public, no need to call md.login() or md.logout()
+md.login(mast_token)
+for vis in visits:
```
taylorbell57, line 30: I can't tell at a glance - are the different visits put into the same folder in the end or different folders?

demos/HST/download_data_HST_template.py
taylorbell57: Running this locally also gives me the following error traceback:
```
  File "/Users/tjbell1/Research/JWST/ERS/Eureka2/demos/HST/download_data_HST_temp.ecf", line 34, in <module>
    if result != None:
  File "/Users/tjbell1/miniconda3/envs/eureka/lib/python3.9/site-packages/astropy/table/table.py", line 3429, in __ne__
    return ~self.__eq__(other)
  File "/Users/tjbell1/miniconda3/envs/eureka/lib/python3.9/site-packages/astropy/table/table.py", line 3426, in __eq__
    return self._rows_equal(other)
  File "/Users/tjbell1/miniconda3/envs/eureka/lib/python3.9/site-packages/astropy/table/table.py", line 3474, in _rows_equal
    result = self.as_array() == other
TypeError: Cannot compare structured or void to non-void arrays.
```

taylorbell57: But that is addressed by changing the line to `is result is not None:`

taylorbell57 (changes requested):
Alright, with one tweak this was functional for me, but flake8 also has some complaints and we should add `astroquery` as an explicit requirement

demos/HST/download_data_HST_template.py
taylorbell57: I can see now that it puts them into the same folder. I was going to suggest that they end up in different folders, but I know my phase curve and likely some other observations had to be broken into multiple visits, so perhaps it's fine to leave this as is.

taylorbell57 (approved):
Alright, as long as putting different visits into the same folder is the intended behaviour, this PR is good to go, and I can add the sci/cal folder edit in my WFC3 PR. You might want to add something about this file to the documentation in the future too, but I don't think that's critical enough to hold this PR back for it

kevin218:
Thanks for making the changes for me.  It's been a rough week for me with way too many deadlines.

If users want to put visits into their own directories, they can make `final_dir` a list and iterate over the items within the for loop.

PR #370, opened 2022-07-06
Bug patches and log tweaks

taylorbell57:
This PR:

1. Fixes the allapers setting in Stages 4+ which was broken by my previous fix for datetime.
2. Fixes Stage 6 for lsq runs from Stage 5 (lack of error bars previously broke this)
3. Fixes broken allanplot in Stage 5 if sigma clipping was done in Stage 4
4. Does instart-1 once per file, rather than everywhere instart is used
5. Slightly tweaks S3+ log formatting to look a bit nicer

### DIFF
--- a/src/eureka/S5_lightcurve_fitting/plots_s5.py
+++ b/src/eureka/S5_lightcurve_fitting/plots_s5.py
@@ -188,8 +188,8 @@ def plot_corner(samples, lc, meta, freenames, fitter):
     - December 29, 2021 Taylor Bell
         Moved plotting code to a separate function.
     """
-    samples = np.copy(samples)
-    freenames = np.copy(freenames)
+    samples = samples
+    freenames = freenames
     ndim = len(freenames)+1  # One extra for the 1D histogram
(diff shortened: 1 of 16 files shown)


src/eureka/S5_lightcurve_fitting/plots_s5.py
```diff
@@ -188,8 +188,8 @@ def plot_corner(samples, lc, meta, freenames, fitter):
     - December 29, 2021 Taylor Bell
         Moved plotting code to a separate function.
     """
-    samples = np.copy(samples)
-    freenames = np.copy(freenames)
+    samples = samples
```
kevin218, line 191: What is the point of these two lines (191+192)?

taylorbell57: Hah, indeed - I'd meant to remove the np.copy which was unnecessary and RAM intensive, but had not meant to leave those lines behind

kevin218 (approved):

Issue #165, opened 2022-03-04
SRCYPOS Estimated Incorrectly (jwst==1.4.2)


AarynnCarter:
Okay I've spent the better part of the day looking at the error I had on Tuesday and I'm a little stumped, but have made some progress. 

```
  File "/Users/acarter/Documents/TRANSITS/EUREKA/DC_NIRSPEC/../Eureka_dev/eureka/S3_data_reduction/background.py", line 171, in fitbg
    print(np.where(mask[j,xvals]))

IndexError: index 30 is out of bounds for axis 1 with size 30 
```

The crash occurs during the background subtraction, and I've realised this is because a dodgy SRCYPOS value of ~221 is being used to create the xvals array, when it should probably be in the range of 0-31. The weird thing is that I cannot find out where this value is being set. I've grepped the string in both Eureka and my site-packages folders and I can only find instances where it's called.

When I made the Stage 2 data for the data challenge, I used the default pipeline process and the SRCXPOS/SRCYPOS values are *not* in the SCI header. So we're probably doing something in Eureka slightly differently? That or my .ecf file isn't set up right. 

@taylorbell57 I would greatly appreciate if you could quickly run Stage 2/3 with the .ecf files I've attached and let me know whether you get a similar error / the SRCYPOS value ends up in the header of your calints file. 

I'm going to reach out to Erin and see if she'll share her Stage 2 .ecf file with me too.  

[ecfs.zip](https://github.com/kevin218/Eureka/files/8182201/ecfs.zip)

taylorbell57:
Huh, so you're saying that eureka's S2 pipeline is making a SRCYPOS value in the FITS header that your separate jwst stage 2 pipeline attempts did not add? Have you looked at the actual output S2 FITS files to confirm that SRCYPOS is actually in there and not something that is going wrong with the methods we wrote to get the source position in Stage 3? I definitely didn't intentionally change anything in the S2 code that would determine the SRCYPOS, but I suppose it's possible that something went wrong. Or perhaps it is related to the slit_y_high, slit_y_low, etc. parameters that you wrote code to edit?

AarynnCarter:
Okay, sorry to leave this for so untouched for so long. I ran things with the slit_y / wavelength parameters set to the default (None) and still get this issue...

@taylorbell57 or @erinmmay, could you send over a copy of your anaconda environment? Maybe it's an issue with my package versions(conda env export > environment.yml). Thanks!

taylorbell57:
[...]

PR #373, opened 2022-07-12
Updating jwst package version to 1.6.0

### LINKED ISSUE #335: Updating jwst package version to >= 1.4.0
As of right now our code needs to use `jwst` version 1.3.3 for at least some of the simulations (see #165), but I know that real MIRI observations are going to need `jwst` >= 1.4.0 at least to run Stage 2 (version 1.5.2 is the current up-to-date version). At some point we're going to need to break compatibility with the simulations to handle the real observations, but I don't know when we should plan on doing that.

### Code of Conduct

taylorbell57:
This PR updates the `jwst` package version from 1.3.3 to the latest 1.6.0 (resolves #335). In addition, for newer versions of `jwst` than 1.3.3 we don't need to specify an `asdf` version number (that was just a bug with `jwst` 1.3.3), so I removed that version number.

The main reason we couldn't do this before is that newer versions of the `jwst` pipeline broke for the NIRSpec simulations (see Issue #165). I found that this could be quickly resolved by allowing the value in the FITS header to be overridden using the source_pos function which already exists. Previously, if the FITS header had a SRCYPOS value, you had to use it, but now that behaviour is only the case if `meta.source_pos_type='header'`. Moving forward, I think if we want to use the SRCYPOS value with subarray observations it's possible we might need to account for the subarray offset, but I'm not sure if that's the case or if this is just an artifact from the simulated data. I'll note this in Issue #165 for now, and if we don't see this SRCYPOS issue with `meta.source_pos_type='header'` for real data then we can close that issue, or if we do see this issue then we can look into things like subarray offsets.

I have confirmed that all the tests pass now, but we should check S1-3 offline for the full MIRI, NIRCam, and NIRSpec simulations to make sure that they all still work correctly.

### DIFF
--- a/demos/HST/S3_wfc3_template.ecf
+++ b/demos/HST/S3_wfc3_template.ecf
@@ -18,7 +18,7 @@ flatfile        flats/WFC3.IR.G102.flat.2.fits         # The calibration file to
 # Subarray region of interest
 ywindow         [80,250]    # Vertical axis as seen in DS9
 xwindow         [80,250]    # Horizontal axis as seen in DS9
-src_pos_type    hst         # Determine source position when not given in header (Options: gaussian, weighted, max, or hst)
+src_pos_type    hst         # Determine source position when not given in header (Options: header, gaussian, weighted, max, or hst)
 
 # Subarray region for direct images used for wavelength calibration
 centroidtrim    5           # The box width to cut around the centroid guess to perform centroiding on the direct images. This should be an integer.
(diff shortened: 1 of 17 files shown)


kevin218 (dismissed):

taylorbell57:
So I just tried running this locally for Stage 1, and it appears that the `integ_info` returned from the `stcal.ramp_fitting.utils.output_integ` function changed from version 0.7.3 to the latest version 1.0.0 that came out 18 days ago, and that ends up impacting `eureka.S1_detector_processing.ramp_fitting.create_integration_model` as the `int_times` are no longer included in `integ_info`. I'll update `eureka.S1_detector_processing.ramp_fitting.create_integration_model` and set a requirement that `stcal >= 1.0.0`

kevin218 (approved):

PR #376, opened 2022-07-13
Adding S4 white-light output, S5 white-light constrained fits

taylorbell57:
This PR outputs a white-light light curve from Stage 4 if requested, and allows users to constrain parameters in S5 as `'white_free'` to use a prior based on the posterior from an initial white-light fit or `'white_fixed'` to freeze the parameter based on the white-light fit. If no parameters in the EPF are `'white_free'` or `'white_fixed'`, then no white-light fit is performed. If any parameters are to be constrained by the white-light fit, S5 first performs a white-light fit and then updates the Parameters object based on that fit before starting the spectroscopic fits. This feature is particularly useful for things like orbital parameters which should not vary with wavelength.

This PR is relevant to issue #225, but different


src/eureka/S5_lightcurve_fitting/fitters.py
```diff
@@ -1262,7 +1276,9 @@ def save_fit(meta, lc, model, fitter, results_table, freenames, samples=[]):
 
     # Save the S5 outputs in a human readable ecsv file
     event_ap_bg = meta.eventlabel+"_ap"+str(meta.spec_hw)+'_bg'+str(meta.bg_hw)
-    if lc.share:
+    if lc.white:
```
kevin218, line 1279: We should define channel_tag sooner and then make fname simpler: `fname = f'S5_{fitter}_samples{channel_tag}'.

kevin218 (approved):
One small comment that is totally not important.

kevin218 (approved):

PR #379, opened 2022-07-13
Download JWST data

kevin218:
This PR allows users to download JWST using the template provided in the demos directory.

### DIFF
--- a/demos/JWST/download_data_JWST_template.py
+++ b/demos/JWST/download_data_JWST_template.py
@@ -3,38 +3,43 @@
 import eureka.lib.mastDownload as md
 
 # Proposal/Program ID, can be string or int
-proposal_id = '13467'
+proposal_id = '02734'
 
 # List of one or more visit numbers
-visits = [60]
-# visits = np.arange(60,85)
+visits = [2]
+# visits = np.arange(1,4)
 
-# Instrument name, can be upper or lower case.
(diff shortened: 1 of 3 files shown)


demos/JWST/download_data_JWST_template.py
```diff
@@ -0,0 +1,47 @@
...
+# If data are public, no need to call md.login() or md.logout()
+md.login(mast_token)
+for vis in visits:
+    # Download data from MAST Archive
+    result = md.downloadJWST(proposal_id, vis, calib_level, subgroup, download_dir)
+    if result is not None:
+        # Consolodate and move data into new directory
+        md.consolidate(result, final_dir)
+        # Sort data into science and calibration folders (scan vs direct image)
+        # md.sort(final_dir)
```
taylorbell57, line 43: Do we want these lines to still be there for JWST?

taylorbell57 (approved):
Alright, I made some very minor tweaks to the docstrings and put the link to the website to generate the API tokens in the download.py files for easier access. Otherwise this looks good to me (see my above comment though)

demos/JWST/download_data_JWST_template.py
kevin218: Maybe... I need a better understanding of what files are needed.  Maybe some of the FG files will be helpful, at which point we'll want to sort them into their own folder.

PR #384, opened 2022-07-14
Fixing overly RAM intensive S3 NaN masking

taylorbell57:
Need to do masking on numpy arrays rather than xarrays to avoid extreme
RAM usage


taylorbell57:
This PR's edits ended up getting encorporated into PR #385, so closing here

PR #390, opened 2022-07-15
Fixing time arrays for real data

taylorbell57:
We incorrectly guessed that each segment would have the whole time
array. Each segment only has the time values for that segment, so no
need to crop the arrays


kevin218 (dismissed):

taylorbell57:
Hmm, just saw that this fails the pytest - I'll look into how that could be

kevin218 (approved):

PR #392, opened 2022-07-15
More useful error message if no FITS files found

taylorbell57:
Because of an update I made to util.readfiles, an unhelpful error
message would be raised if no files were found. This patch reduces
duplicated code between S1-3 and raises a better message if no
files are found


kevin218 (approved):
I ran into this problem earlier today, as are many people I'm sure.  Timely fix!

PR #393, opened 2022-07-15
Trying out automatic version numbering

taylorbell57:
Using setuptools_scm to allow automatic version numbering

Let's not merge this quite yet, I'm just using the automated tests to make sure that the tweak behaves as expected


taylorbell57:
Alright, this PR takes advantage of https://github.com/pypa/setuptools_scm/ to allow us to remove all hard coded version numbers in Eureka! and instead uses the git tags to automatically set version numbers. So if the latest git tag is v0.2.1 and you install code that is 20 commits past that tag (which this branch is), then setuptools_scm says that version of the pipeline is version 0.2.2.dev20+gbd6c45d (increase patch number by one, add "dev", add the number of commits past the latest patch, and then give the condensed form of the git commit hash). This basically eliminates any difficulties or headaches with setting version numbers - we just add git tags whenever we feel there has been a significant change and want a new fixed point

taylorbell57:
And while I was at it, I fixed some bugs with the documentation FYI

kevin218 (approved):

Issue #394, opened 2022-07-15
[Bug]: S3: NIRSpec PRISM gain file is not full frame, S3 expects it to be full frame


erinmmay:
### FAQ check

- [X] Yes, I checked the FAQ and my question has not been addressed.

### Instrument

NIRSpec (Stages 1-3)

### What happened?

The gain file is not full frame for NIRSpec PRISM, and the code expects a full frame gain file. When the gain file is then trimmed to the subarray pixels at line 87 in bright2flux.py it results in an empty array in one dimension.

The result is a "ValueError: operands could not be broadcast together with shapes..." when attempting to multiply the data by the subgain

@cpiaulet and @AarynnCarter and I solved this by reading in the header information for the gain file to find the pixel values that the gain file is trimmed to and adjusting the slices appropriately.  see pending PR...

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #395, opened 2022-07-15
fix for gain files that aren't full frame

erinmmay:
for issue #394

### DIFF
--- a/src/eureka/S3_data_reduction/bright2flux.py
+++ b/src/eureka/S3_data_reduction/bright2flux.py
@@ -84,7 +84,14 @@ def dn2electrons(data, meta):
     ny = data.attrs['mhdr']['SUBSIZE2']
 
     # Load gain array in units of e-/ADU
-    gain = fits.getdata(meta.gainfile)[ystart:ystart+ny, xstart:xstart+nx]
+    gain_header = fits.getheader(meta.gainfile)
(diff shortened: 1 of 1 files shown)


src/eureka/S3_data_reduction/bright2flux.py
```diff
@@ -84,7 +84,14 @@ def dn2electrons(data, meta):
...
     # Load gain array in units of e-/ADU
-    gain = fits.getdata(meta.gainfile)[ystart:ystart+ny, xstart:xstart+nx]
+    gain_header = fits.getheader(meta.gainfile)
+    xstart_gain = gain_header['SUBSTRT1']
+    ystart_gain = gain_header['SUBSTRT2']
+
+    ystart_trim = ystart-ystart_gain + 1  # 1 indexed, NOT zero                                                         
+    xstart_trim = xstart-xstart_gain + 1
+
+    gain = fits.getdata(meta.gainfile)[ystart_trim:ystart_trim+ny, xstart_trim:xstart_trim+nx]
```
taylorbell57, line 94: Oh my, this PR just made me realize that a few lines below we rotate the MIRI gain file but don't then do `[:, ::-1]` like we do with the data, so our gain file is definitely being applied wrong for MIRI!

kevin218 (approved):

Issue #396, opened 2022-07-15
[Bug]: Incorrect time arrays in S3 traced back to S1 custom ramp fitting stripping int_times from header


erinmmay:
### FAQ check

- [X] Yes, I checked the FAQ and my question has not been addressed.

### Instrument

NIRSpec (Stages 1-3)

### What happened?

After stage 1 all headers contain no data in the int_times keyword. 

As a result in stage 3 it assumes it is simulated data (i.e with no int_times values) and creates time arrays that have length=intend where intend is the number of the last integration in that segment. For a single segment this works, but for multiple segments, when running segment 2+, this results in a time array that is too long and you get the error below.

Best @cpiaulet, @AarynnCarter, and I can trace it back to is the custom ramp fitting in stage 1 which does set the int_times array to empty in multiple places. Assuming this is a holdover from the simulated data? We need some assistance from someone more familiar with this step!

### Error traceback output

Traceback (most recent call last):
 File “/home/mayem1/Analysis/JWST/wasp39b/nirspec_prism/run_eureka.py”, line 28, in <module>
  s3_spec, s3_meta = s3.reduce(eventlabel, ecf_path=ecf_path)
 File “/home/mayem1/anaconda3/envs/eureka/lib/python3.9/site-packages/eureka/S3_data_reduction/s3_reduce.py”, line 236, in reduce
  data, meta, log = inst.read(meta.segment_list[i], data,
 File “/home/mayem1/anaconda3/envs/eureka/lib/python3.9/site-packages/eureka/S3_data_reduction/nirspec.py”, line 85, in read
  data[‘flux’] = xrio.makeFluxLikeDA(sci, time, flux_units, time_units,
 File “/home/mayem1/anaconda3/envs/eureka/lib/python3.9/site-packages/astraeus/xarrayIO.py”, line 114, in makeFluxLikeDA
  da = xr.DataArray(
 File “/home/mayem1/anaconda3/envs/eureka/lib/python3.9/site-packages/xarray/core/dataarray.py”, line 402, in __init__
  coords, dims = _infer_coords_and_dims(data.shape, coords, dims)
 File “/home/mayem1/anaconda3/envs/eureka/lib/python3.9/site-packages/xarray/core/dataarray.py”, line 152, in _infer_coords_and_dims
  raise ValueError(
ValueError: conflicting sizes for dimension ‘time’: length 6100 on the data but length 12200 on coordinate ‘time’

### Code of Conduct

[...]

PR #397, opened 2022-07-16
Resolve issue with blank int_times + fix NIRspec disp axis trimming

AarynnCarter:
Quick fix for issue #396

Haven't had the time to *fully* check this through Stage 3 yet, but I can see that the INT_TIMES is correctly populated after Stage 1. Will take a proper look tomorrow but wanted to put the PR in anyway.

### DIFF
--- a/src/eureka/S1_detector_processing/ramp_fitting.py
+++ b/src/eureka/S1_detector_processing/ramp_fitting.py
@@ -712,7 +712,7 @@ def create_image_model(input_model, image_info):
     out_model.var_poisson = var_poisson
     out_model.var_rnoise = var_rnoise
     out_model.err = err
-    int_model.int_times = input_model.int_times
+    out_model.int_times = input_model.int_times
 
     return out_model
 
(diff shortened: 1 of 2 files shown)


jbrande:
I'm not sure this has resolved the issue, I'm still getting the same error as Erin after pulling your dev branch. For what it's worth, I'm trying this on Everett's minified NIRSpec data.

AarynnCarter:
@jbrande, I'm checking things over now, hopefully I can report back at some point today. I don't think Everett's minified data will work as the INT_TIMES for that file gets populated with 3200 values (i.e. the size of the untrimmed seg004 file)

AarynnCarter:
Okay I can confirm that with this PR, in combination with Erin's #395, Eureka! will run from Stage 1 through Stage 3 on one segment of the PRISM data.

jbrande:
> @jbrande, I'm checking things over now, hopefully I can report back at some point today. I don't think Everett's minified data will work as the INT_TIMES for that file gets populated with 3200 values (i.e. the size of the untrimmed seg004 file)

Ah, yeah. That would do it.

AarynnCarter:
Finally found a fix for the really annoying dispersion axis trimming that the jwst pipeline does to the NIRSpec PRISM data -  now we have the full 512x32 array. It works similarly to the custom ramp fitting routines, by temporarily swapping out one of the pipeline functions for a duplicate with the desired wavelength range trimming. 

@taylorbell57, can't figure out the Flake8 indenting rules, leaving it for now...

kevin218 (approved):

PR #400, opened 2022-07-17
Basic functionality for group level background subtraction + Stage 3 DQ masking

AarynnCarter:
I've added some extra functionality to the earlier stages of Eureka! to help with the NIRSpec PRISM reduction, and maybe others too.  

Firstly, you can now perform a background subtraction on groups in Stage 1 (instead of integrations in Stage 3), which I find greatly reduces the scatter in the light curves and removes "striping" visible in the 2D MAD plot (see images below). Because of this, I also added a toggle to skip/perform the background subtraction in Stage 3, although I've found better results if I leave it on (i.e. double background subtraction). 

Additionally, I added a very quick toggle to mask pixels during Stage 3 where the corresponding DQ array is non-zero. 

I anticipate we'll need to build on these routines further / investigate things more deeply, but this should be a good start. 

fig3101-2D_LC
fig3101-2D_LC


demos/JWST/S1_miri_template.ecf
```diff
@@ -26,6 +26,16 @@ skip_gain_scale     False
 #Pipeline stages parameters
 jump_rejection_threshold  4.0 #float, default is 4.0, CR sigma rejection threshold
 
+#Background subtraction
+grouplevel_bg 		True
+ncpu				8
+bg_y1 				6
+bg_y2 				26
```
taylorbell57, line 33: I assume these are just a copy of what you used for NIRSpec? We'll need to choose values relevant to MIRI (and test whether they actually work for MIRI) before setting this step to run by default in the template

demos/JWST/S1_nirx_template.ecf
```diff
@@ -25,6 +25,16 @@ skip_gain_scale     False
 #Pipeline stages parameters
 jump_rejection_threshold  4.0 #float, default is 4.0, CR sigma rejection threshold
 
+#Background subtraction
+grouplevel_bg 		True
+ncpu				8
+bg_y1 				6
+bg_y2 				26
```
taylorbell57, line 32: Also, I assume this is only for NIRSpec and may not work well for NIRCam (and especially not NIRISS)? We'll likely need to break the templates for S1 into NIRCam, NIRSpec, and NIRISS if it does end up being instrument dependent

demos/JWST/S1_nirx_template.ecf
```diff
@@ -25,6 +25,16 @@ skip_gain_scale     False
...
+bg_deg 				0
+p3thresh			3
+verbose				True
+isplots 			1
```
taylorbell57, line 36: Please use `isplots_S1` to keep the format similar to the other stages

demos/JWST/S1_miri_template.ecf
```diff
@@ -26,6 +26,16 @@ skip_gain_scale     False
 #Pipeline stages parameters
 jump_rejection_threshold  4.0 #float, default is 4.0, CR sigma rejection threshold
 
+#Background subtraction
+grouplevel_bg 		True
```
taylorbell57, line 30: Please don't use tabs in the ECF/EPF templates as it messes up the formatting for people depending on what file viewer they're using

demos/JWST/S3_miri_lrs_template.ecf
```diff
@@ -11,8 +11,10 @@ suffix          calints     # Data file suffix
 ywindow         [150,380]   # Vertical axis as seen in DS9
 xwindow         [7,70]      # Horizontal axis as seen in DS9
 src_pos_type    gaussian    # Determine source position when not given in header (Options: header, gaussian, weighted, max, or hst)
+dqmask			True        # Mask non-zero pixels from the DQ array
```
taylorbell57, line 14: Great thinking!

demos/JWST/S1_miri_template.ecf
```diff
@@ -26,6 +26,16 @@ skip_gain_scale     False
 #Pipeline stages parameters
 jump_rejection_threshold  4.0 #float, default is 4.0, CR sigma rejection threshold
 
+#Background subtraction
```
taylorbell57, line 29: Please document all new ECF/EPF settings in docs/source/ecf.rst and also update the ECF/EPF files in docs/media/

src/eureka/S1_detector_processing/s1_process.py
```diff
@@ -90,9 +90,9 @@ def rampfitJWST(eventlabel, ecf_path=None):
             if hdulist[0].header['INSTRUME'] == 'NIRCAM':
                 hdulist[0].header['NDITHPTS'] = 1
                 hdulist[0].header['NRIMDTPT'] = 1
-
+            
             EurekaS1Pipeline().run_eurekaS1(filename, meta, log)
-
+
```
taylorbell57, line 95: If you want to add blank lines here, please make sure they have no tabs or spaces on them

src/eureka/S3_data_reduction/plots_s3.py
```diff
@@ -28,8 +28,8 @@ def lc_nodriftcorr(meta, wave_1d, optspec, optmask=None):
     None
     '''
     normspec = util.normalize_spectrum(meta, optspec, optmask=optmask)
-    wmin = np.ma.min(wave_1d)
-    wmax = np.ma.max(wave_1d)
+    wmin = np.nanmin(wave_1d)
+    wmax = np.nanmax(wave_1d)
```
taylorbell57, line 32: Did you mean to do this? I guess wave_1d isn't a masked array so you needed to use the np.nan versions to handle those strange values you mentioned in your issue?

src/eureka/S1_detector_processing/ramp_fitting.py
```diff
@@ -87,6 +86,33 @@ def process(self, input):
...
+                    dqmask = np.where(dq[:,ngrp,:,:]!= 0)
+                    grp_mask[dqmask] = 0
+
+                    xrdata = (['time', 'y', 'x'], grp_data)
+                    xrmask = (['time', 'y', 'x'], grp_mask)                
+                    xrdict = dict(flux=xrdata, mask=xrmask)
+                    data = xrio.makeDataset(dictionary=xrdict)
+                    data['flux'].attrs['flux_units']='n/a'
+
+                    data = bkg.BGsubtraction(data, self.s1_meta, self.s1_log, self.s1_meta.isplots)
```
taylorbell57, line 112: I assume this line is probably too long for flake8. The following should work fine (and gives you some idea of how flake8 wants things formatted)
```
                    data = bkg.BGsubtraction(data, self.s1_meta, self.s1_log,
                                             self.s1_meta.isplots)
```

src/eureka/S1_detector_processing/ramp_fitting.py
```diff
@@ -87,6 +86,33 @@ def process(self, input):
...
+                import astraeus.xarrayIO as xrio
+
+                all_data = input_model.data
+                dq = input_model.groupdq
+                self.s1_meta.inst = input_model.meta.instrument.name.lower()
+                self.s1_meta.int_start = 0
+                self.s1_meta.n_int = all_data.shape[0]
+
+                for ngrp in range(all_data.shape[1]):
+                    grp_data = all_data[:,ngrp,:,:]
```
taylorbell57, line 101: flake8 will want spaces after every comma

taylorbell57 (changes requested):

src/eureka/S1_detector_processing/ramp_fitting.py
```diff
@@ -87,6 +86,33 @@ def process(self, input):
             Updated for JWST version 1.3.3, code restructure
         '''
         with datamodels.RampModel(input) as input_model:
+
+            if self.s1_meta.grouplevel_bg:
```
taylorbell57, line 90: The simplest way to resolve most of my comments is likely to just replace this line with the following and then remove the ECF settings from the other instruments. That way we're not leading people astray in the templates but get this feature merged as soon as possible:
```
if hasattr(self.s1_meta, 'grouplevel_bg') and self.s1_meta.grouplevel_bg:
```
The above change makes is so that you don't need to add the grouplevel_bg and related settings to the ECF.

taylorbell57:
This will later be merged by a different, larger PR, so closing this for now

PR #403, opened 2022-07-18
Update documentation to v0.3

kevin218:
Looks like this PR has changes that should have already been merged.  Changes were only supposed to be in installation.rst.  The other files have already been reviewed.

### DIFF
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -5,15 +5,15 @@ Installation
 Installation methods
 --------------------
 
-In order to have consistent, repeatable results across the ``Eureka!`` user community, we recommend that all general users install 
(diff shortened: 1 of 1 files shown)


taylorbell57:
@kevin218, a reminder that you should probably retag the merge of this PR as v0.3

kevin218:
Doesn't look like I can make edits to a current release. When I delete and recreate it, it goes back to the version from 2 days ago.  Ultimately, I think we're ok.  People will look at the installation instructions online, not their local copy.

taylorbell57:
I was able to work it out in the end - I had to delete the tag from the remote and local copies and then could retag

PR #414, opened 2022-07-20
Bug in medstddev calc, no axis specified

kevin218:


src/eureka/lib/medstddev.py
```diff
@@ -106,7 +106,7 @@ def medstddev(data, mask=None, medi=False, axis=0):
     residuals = data - median
     # calculate standar deviation:
     with np.errstate(divide='ignore', invalid='ignore'):
-        std = np.ma.sqrt(np.ma.sum(residuals**2.0) / (ngood - 1.0))
+        std = np.ma.std(residuals, axis=axis)
```
taylorbell57, line 109: This meaningfully changes the std value returned from this function, and so the test for this function currently fails. Are you sure you mean to do this, or do you just want to change it to the following which shouldn't change the return values but should fix whatever bug you encountered
```
std = np.ma.sqrt(np.ma.sum(residuals**2.0, axis=axis) / (ngood - 1.0))
```

kevin218: I originally just added the axis keyword, as you suggested, then decided to test it against np.ma.std, which is cleaner. Both produced similar results, difference likely being in using N-1 vs N.

I'll have to look into why it's failing.

FYI, without the axis keyword, nothing was getting flagged in this routine.

taylorbell57: At present, it is failing because I wrote a unit test based on the test cases written in the docstring. You'll just need to update the docstring and tests if you want to use std

kevin218: I've updated the std calc to include ddof=1 (default is 0).  The pytest results were closer, but not quite identical, so I also updated the expected output.  Since I need this code for the NIRCam data, I'm going to merge as Admin.

PR #417, opened 2022-07-23
Update S2_nircam_wfss_template.ecf

meganmansfield:
skip_bkg_subtract was set to False by default


kevin218 (changes requested):
Also need to update the file in docs/media

kevin218 (approved):

PR #420, opened 2022-07-25
Allowing manual clipping and unbinned spectra from S4

taylorbell57:
This PR allows the manual removal of data in S4 (e.g. the first several integrations, the last several integrations, or any integrations in the middle). This PR also allows unbinned spectra out of S4 by setting nspecchan=None. I also addressed some minor bugs I found while testing on WFC3 data including not updating meta.n_int after summing the NDRs.


taylorbell57:
While I was at it, I made some minor tweaks to S5 to use the new util.normalize_spectrum (needed for WFC3) and fixed a bug with the physeval function if interp=True (also needed for WFC3).

src/eureka/S4_generate_lightcurves/plots_s4.py
```diff
@@ -201,7 +201,7 @@ def lc_driftcorr(meta, wave_1d, optspec, optmask=None):
         plt.xlabel(r'Wavelength ($\mu m$)')
         plt.colorbar(label='Normalized Flux')
 
-        if len(meta.wave_low) > 1:
+        if len(meta.wave) > 1 and len(wave_1d) != meta.nspecchan:
```
kevin218, line 204: Did you mean to change the first part from `wave_low` to `wave`?

src/eureka/S4_generate_lightcurves/s4_genLC.py
```diff
@@ -294,6 +310,10 @@ def genlc(eventlabel, ecf_path=None, s3_meta=None):
                     plots_s4.driftxpos(meta, lc)
                     plots_s4.driftxwidth(meta, lc)
 
+            if hasattr(meta, 'manual_clip') and meta.manual_clip is not None:
```
kevin218, line 313: Manual clipping should go into Stage 5.  When fitting the LCs, people will want to try different clip sizes without having to rerun S4.

src/eureka/lib/util.py
```diff
@@ -92,6 +92,56 @@ def trim(data, meta):
     return subdata, meta
 
 
+def manual_clip(spec, lc, meta, log):
```
kevin218, line 95: When moved to S5, only `lc` should be modified, thus `spec` won't be needed as an input parameter.

kevin218 (changes requested):

src/eureka/S4_generate_lightcurves/plots_s4.py
taylorbell57: Yeah, the length of both arrays must be the same, and `meta.wave` is shorter to write than `meta.wave_low`

kevin218 (approved):

taylorbell57:
Let's not merge this yet - I just noticed that the tests are currently failing

taylorbell57:
Alright, this is ready for approval now @kevin218

kevin218 (approved):

Issue #308, opened 2022-05-09
Eureka! Code Profiling


jbrande:
### What happened?

We need to profile Eureka's performance in order to assess major bottlenecks and hopefully speed things up. This is especially important for Stage 3 and Stage 5. I think I can do this with Python's cProfile utility, so that's where I'll be starting.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

jbrande:
Looking at S5 emcee runs now. Most of the time spent is spent in likelihood.py, but a lot of this actually looks to be built-in python/numpy stuff we probably don't have much control over. However, about 50 seconds of the 135 seconds that we're in likelihood.py is spent in Model.py either evaluating or updating the models. In update(), a good chunk of this is spent reading EPFs, and parsing the coefficients.

![s5_profile](https://user-images.githubusercontent.com/5041139/168348058-ab3cfbbc-ce1f-42a0-9512-0c5635bbfa91.jpg)

In eval(), the biggest time sink seems to be evaluating the transit and systematics models. The Batman transit model takes more of this time than the polynomial systematics model, so perhaps when starry gets implemented, we might be able to speed up the transit lightcurves a bit.

taylorbell57:
This is super interesting Yoni, thanks so much for looking into this! I think improving the update function will help a lot and should be pretty doable. I've been thinking for a while that we're really not doing things quickly in that function, but it's nice to see just how big an effect this has.

For S3 I definitely agree that a tonne of time is spent on plotting. The plot resolution size might have some effect on this, but I wonder if that's the main issue or if there's something else we can do. I guess it is the savefig lines that are the slowest though, so maybe resolution really is the bigger concern. I wonder if PDF vs PNG significantly changes things and if there's any other optimizations like that we can do (like different backends as you'd mentioned).

taylorbell57:
I quite strongly suspect that a significant part of issue #223 is caused by the inefficient update functions in Stage 5, but it's possible dynesty just scales very poorly with the number of data and/or fitted parameters

jbrande:
Yeah, getting into the dynesty profile it's similar to the emcee profile, except that the nested sampling prior transforms also take a good bit of time. 
![s5_dynesty_profile](https://user-images.githubusercontent.com/5041139/168491451-1b621acc-c4b2-42eb-af8a-56cd7954278b.jpg)

Reading EPFs, setting attributes, transit and systematics models all still seem to be where we have room to improve.

[...]

PR #434, opened 2022-07-29
Speeding Up S5

taylorbell57:
As pointed out by Yoni in #308, a significant amount of time was being spent in S5 on just updating the parameters on each step. The issue is that we have tried to make things very robust, object oriented, and whatnot, but doing so can have significant runtime costs if special care isn't taken. This PR is a step in that direction where I've done speed testing in ipython to figure out the fastest way I could think of to do Model.update() without changing the format of anything. Using cProfile as Yoni did and fitting a real MIRI/LRS transit with lsqfitter, I find a ~15% improvement in fitting runtime from these fairly minor edits. In the figures below, I've highlighted the Model.update function's runtime in pink where most of the runtime difference comes from. For these fits, I fitted the white-light lightcurve and then a single wavelength (by setting `testing_S5` to True). With a parallelized emcee fit it is harder to profile the code, but running them side-by-side with 4 threads each on an 8 core machine, the updated code finishes 24 seconds earlier (3% faster). For whatever reason, dynesty doesn't work properly for the spectral data for this dataset with either set of code (takes 10+ minutes so I terminated it).

Code from the main branch:
Screen Shot 2022-07-29 at 10 48 01

Code from this PR:
Screen Shot 2022-07-29 at 10 48 15


kevin218 (approved):
Nice speed improvement!

Issue #313, opened 2022-05-09
[Enhancement]: Photometry Routines


taylorbell57:
As we've been discussing for a while, we need photometry methods for the simultaneous photometry from NIRCam and from specific photometry observations (e.g. Laura and I (through Tom Greene) each have TRAPPIST-1 photometry observations coming from MIRI). I believe @sebastian-zieba has already done most of the work of repurposing previous photometry code (POET based code I believe).

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

taylorbell57:
This has been partially addressed by PR #437 for NIRCam photometry, but photometry is still yet to be supported for other instruments

PR #437, opened 2022-08-01
NIRCam photometry routine

sebastian-zieba:
I've added a photometry routine to Eureka! It has been successfully tested with commissioning data of HAT-P-14 b and ERS data of WASP-39 b. 
(Related to Issue #313)

Most changes affect S3. Only minor changes had to be made to S4. S5 has been untouched.

Workflow for S3:
- When reading in the data in nircam.py, the code checks if the "CHANNEL" entry in the header is "SHORT". If so it will set meta.photometry = True. This entry in the metadata will be then used all throughout S3 and S4.
- After doing things that already exist in s3_reduce.py (like trimming the array, checking if nans are in the array, ...), the newly added code will flag bad pixels using the Data Quality array (the user can skip that if they want).
- Next, an outlier flagging along the time axis will be performed for every single pixel in the segment (the user can skip that if they want).
- Next, we correct for the one over f noise if the user wants to do this in S3 and not in S1 or S2 (the user can skip that if they want).
- Next, we determine the centroid position (code taken from POET). This is being done twice in the code. First, a coarse centroid determination where we fit a 2D gaussian to the whole frame. Then we cut out a small area around this centroid and do another 2D gaussian fit. The user doesn't need to specify a guess position. This double-fitting approach has been more robust than a center of light method if the target is partially off the detector (like in the case of HAT-P-14 b).
- Before performing aperture photometry, the user can decide if they want to interpolate where bad pixel were found. This uses scipy.interpolate.griddata
- Next, for the aperture photometry, the user chooses an aperture, and two radii for the sky background (code taken from POET).
- The remaining part uses already existing code from S3.

Several diagnostic plots are being saved during an S3 run. 

isplots_S3 = 1:
the light curve:
fig3103-1D_LC

the movement of the centroid:
fig3104-Centroid

isplots_S3 = 3:

a background light curve:
fig3305-1D_BG_LC

2D frame:
fig3307_file0_int000_2D_frame

aperture size over time:
fig3309_aperture_size

one over f correction in S3:
fig3310_file0_int000_2D_frame

isplots_S3 = 5:

zoom into aperture region:
fig3308_file3_int8_2D_frame

2D gaussian fit to determine centroid:
fig3502_file0_int000_centroid_fgc

differences between consecutive 2D frames (might be interesting to track down mirror movements):
fig3503_int0_2D_frame_diff

Two bugs in S5 were found and also fixed in this PR:
- manual_clip in S5 doesn't work. 

in s5_fit.py: _util.manual_clip(lc, meta, log)_ had to be changed into _meta, lc, log = util.manual_clip(lc, meta, log)_ otherwise no trimming will actually happen.

- using "square-root" as a limb darkening law will return an error. batman calls the law "squareroot". All "square-root" entries were therefore changed to "squareroot".


sebastian-zieba:
Okay, I see that flake8 and pytest didn't go through. Looks like flake8 is mostly complaining about the POET code (tabs are not a multiple of 4) and I also know why the tests didn't work. I'm gonna work on this tomorrow.

taylorbell57:
Thanks for working out the issues with flake8 and the already existing pytest tests. Can also please write a test for the photometry code you've added so that we can all make sure that we don't break things for photometry when we change things for spectroscopy?

sebastian-zieba:
Sure, no worries.
Yeah sounds good, I'm gonna work on some tests tomorrow.

sebastian-zieba:
Oh, should I also add a little bit about the photometry routine to the Documentation already? Like an explanation for the new S3 ecf flags I've added?

taylorbell57:
> Oh, should I also add a little bit about the photometry routine to the Documentation already? Like an explanation for the new S3 ecf flags I've added?

Yes, please do so

sebastian-zieba:
Update:
- Fixed bug I introduced into the photometry routine when working out the flake8 issues
- Added explanation to the new ecf options to the documentation

ToDo:
- Documentation:
  -  Add photometry plots to Eureka! Outputs page
- Add pytest tests

kevin218:
In the interest of time (and the need to analyze the NIRCam SW channel this week), I suggest we accept this PR as long as it is free of any known bugs and have a separate PR next week with documentation and tests.

sebastian-zieba:
I am just about to push a last commit fixing some minor bugs. I'll push it after the meeting. The tests are also finished, only problem is that the fits file is pretty big.

sebastian-zieba:
I will push the final fixes today and the tests (which need a small fits file based on the simulated photometry dataset) can be pushed later.

sebastian-zieba:
The latest push removed the odd-DQ-flagging written for the photometry routine and now uses the already existing code in s3_reduce.py. If the user wants to do odd DQ flagging then they'll have to have "dqmask          True " in their ecf. I added this entry to every ecf and wrote a short description for the ecf page on ReadTheDocs.

src/eureka/lib/apphot.py
```diff
@@ -0,0 +1,621 @@
...
+
+    ret[stat] = status
+    # ttotal =  time.time() - tini
+
+    return ret[np.where(retidx)]
+
+
+def apphot_status(data):
+    """
+    Prints a warning if aperture step had errors.
```
taylorbell57, line 595: Docstring format likely needs tweaking to format properly

src/eureka/lib/apphot.py
```diff
@@ -0,0 +1,621 @@
+# $Author$
+# $Revision$
+# $Date$
+# $HeadURL$
+# $Id$
```
taylorbell57, line 5: Can delete these lines

src/eureka/lib/centerdriver.py
```diff
@@ -0,0 +1,106 @@
+# $Author: patricio $
+# $Revision: 304 $
+# $Date: 2010-07-13 11:36:20 -0400 (Tue, 13 Jul 2010) $
+# $HeadURL: file:///home/esp01/svn/code/python/branches/patricio/photpipe/lib/centerdriver.py $
+# $Id: centerdriver.py 304 2010-07-13 15:36:20Z patricio $
```
taylorbell57, line 5: Can delete these lines

src/eureka/lib/centerdriver.py
```diff
@@ -0,0 +1,106 @@
...
+
+def centerdriver(method, data, guess, trim, radius, size,
+                 mask=None, uncd=None, fitbg=1, maskstar=True,
+                 expand=5.0, psf=None, psfctr=None, i=None, m=None, meta=None):
+    """
+    Use the center method to find the center of a star in data, starting
+    from position guess.
+
+    Parameters:
+    -----------
```
taylorbell57, line 21: No colons after Parameters and Returns anywhere (and that also means shortening the underline below)

src/eureka/lib/centerdriver.py
```diff
@@ -0,0 +1,106 @@
...
+    --------
+    A y,x tuple (scalars) with the coordinates of center of the target
+    in data.
+
+    Example:
+    --------
+    nica
+
+    Modification History:
+    ---------------------
```
taylorbell57, line 50: Docs won't like this heading either

src/eureka/lib/chunkbad.py
```diff
@@ -0,0 +1,186 @@
+# $Author: patricio $
+# $Revision: 288 $
+# $Date: 2010-06-24 03:33:44 -0400 (Thu, 24 Jun 2010) $
+# $HeadURL: file:///home/esp01/svn/code/python/branches/patricio/photpipe/lib/poet_chunkbad.py $
+# $Id: poet_chunkbad.py 288 2010-06-24 07:33:44Z patricio $
+
+
```
taylorbell57, line 7: Can delete these lines

src/eureka/lib/chunkbad.py
```diff
@@ -0,0 +1,186 @@
...
+    column of FP.  It replaces flagged pixels with the median for
+    that location within the chunk, adding back on the original
+    frame median.  If all pixels are flagged in that location
+    within the chunk, it uses just the frame median.
+
+    Mean images: For each pixel position in Data (i.e., Data[i,j,*,pos])
+    it calculates the mean of pixels that are not flagged as bad.
+
+    Revisions
+    ---------
```
taylorbell57, line 73: Docs won't like this heading

src/eureka/lib/imageedit.py
```diff
@@ -0,0 +1,261 @@
+# $Author: patricio $
+# $Revision: 301 $
+# $Date: 2010-07-10 03:33:44 -0400 (Sat, 10 Jul 2010) $
+# $HeadURL: file:///home/esp01/svn/code/python/branches/patricio/photpipe/lib/imageedit.py $
+# $Id: imageedit.py 301 2010-07-10 07:33:44Z patricio $
+
```
taylorbell57, line 6: Can delete these lines

src/eureka/lib/imageedit.py
```diff
@@ -0,0 +1,261 @@
...
+pasteimage(data, subim, (dyc,dxc), (syc,sxc)=(None,None)):
+    Inserts subim array into data, matching the specified coordinates.
+
+Examples:
+---------
+
+Revisions
+---------
+2010-07-11  patricio  Added documentation.      pcubillos@fulbrightmail.org
+"""
```
taylorbell57, line 37: Can also delete these lines

src/eureka/lib/imageedit.py
```diff
@@ -0,0 +1,261 @@
...
+
+
+def trimimage(data, c, r, mask=None, uncd=None, oob=0):
+    (yc, xc) = c
+    (yr, xr) = r
+
+    """
+    Extracts a rectangular area of an image masking out of bound pixels.
```
taylorbell57, line 45: There should not be code before the docstring starts

src/eureka/lib/imageedit.py
```diff
@@ -0,0 +1,261 @@
...
+
+    >>> # Set out of bound pixels in the mask to -1:
+    >>> subim, mask = trimimage(data, (dyc,dxc), (1,2), mask=msk, oob=-1)
+    >>> print(mask)
+    [[-1.  1.  1.  0.  1.]
+     [-1.  1.  1.  0.  1.]
+     [-1.  1.  1.  1.  1.]]
+
+    Revisions
+    ---------
```
taylorbell57, line 107: Docs won't like this header

src/eureka/lib/imageedit.py
```diff
@@ -0,0 +1,261 @@
...
+    return ret
+
+
+def pasteimage(data, subim, dy_, syx=(None, None)):
+    (dyc, dxc) = dy_
+    (syc, sxc) = syx
+    # (None,None) = Nne
+
+    """
+    Inserts the subim array into data, the data coordinates (dyc,dxc)
```
taylorbell57, line 151: Again, there should be no code before the docstring

src/eureka/lib/imageedit.py
```diff
@@ -0,0 +1,261 @@
...
+    >>> dyc,dxc = 2,5
+    >>> pasteimage(data, subim, (dyc,dxc), (syc,sxc))
+    [[0, 0, 0, 0, 0],
+     [0, 0, 0, 0, 1],
+     [0, 0, 0, 0, 1],
+     [0, 0, 0, 0, 1],
+     [0, 0, 0, 0, 0]]
+
+    Revisions
+    ---------
```
taylorbell57, line 215: Docs won't like this section

src/eureka/lib/interp2d.py
```diff
@@ -0,0 +1,33 @@
...
+# $Revision: 304 $
+# $Date: 2010-07-13 11:36:20 -0400 (Tue, 13 Jul 2010) $
+# $HeadURL: file:///home/esp01/svn/code/python/branches/patricio/photpipe/lib/interp2d.py $
+# $Id: interp2d.py 304 2010-07-13 15:36:20Z patricio $
+
+# 2d interpolation using 1D interp twice
+
+# must include or all or none of y,x,yi,xi
+
+# iexpand*(np.array(sz)-1)+1 = np.array(sz)+(np.array(sz)-1)*(iexpand-1)
```
taylorbell57, line 11: Can delete all these lines

src/eureka/lib/interp2d.py
```diff
@@ -0,0 +1,33 @@
...
+
+import numpy as np
+
+
+def interp2d(image, expand=5, y=None, x=None, yi=None, xi=None):
```
taylorbell57, line 16: If you understand this function, can you document it?

src/eureka/lib/meanerr.py
```diff
@@ -0,0 +1,114 @@
+# $Author: carthik $
+# $Revision: 267 $
+# $Date: 2010-06-08 22:33:22 -0400 (Tue, 08 Jun 2010) $
+# $HeadURL: file:///home/esp01/svn/code/python/branches/patricio/photpipe/lib/meanerr.py $
+# $Id: meanerr.py 267 2010-06-09 02:33:22Z carthik $
```
taylorbell57, line 5: Can delete these lines

src/eureka/lib/meanerr.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+def meanerr(data, derr, mask=None, err=False, status=False):
+    """
+    Calculate the error-weighted mean and the error in the
+    error-weighted mean of the input data, omitting masked data, NaN
+    data or errors, and data whose errors are zero.
+
+    Parameters:
+    -----------
```
taylorbell57, line 17: No colon after Parameters and Returns or other section titles (and then shorten the underlines by one)

src/eureka/lib/meanerr.py
```diff
@@ -0,0 +1,114 @@
...
+
+    Notes:
+    ------
+    Follows maximum likelihood method (see, e.g., Bevington and
+    Robinson 2003, Data Reduction and Error Analysis for the
+    Physical Sciences, 3rd ed, McGraw Hill, Ch. 4.).
+
+
+    EXAMPLE:
+    --------
```
taylorbell57, line 47: All-caps might break docs

src/eureka/lib/meanerr.py
```diff
@@ -0,0 +1,114 @@
...
+    >>> print(men.meanerr(data, derr,mask=mask, err=True, status=True))
+    (6.279069767441861, 1.4467284665112363, 5)
+
+    >>> derr[4] = 0.0
+    >>> print(men.meanerr(data, derr,mask=mask, err=True, status=True))
+    (5.4545454545454541, 1.6514456476895409, 7)
+
+
+    MODIFICATION HISTORY:
+    ---------------------
```
taylorbell57, line 71: Docs won't like this section title

src/eureka/lib/util.py
```diff
@@ -553,3 +554,62 @@ def manmask(data, meta, log):
...
+
+
+# PHOTOMETRY
+def interp_masked(data, meta, i):
+    """
+    Interpolates masked pixels.
+    Based on the example here:
+    https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.griddata.html
+    """
+    print('Interpolating masked values...')
```
taylorbell57, line 566: This should be a log not a print

src/eureka/lib/util.py
```diff
@@ -553,3 +554,62 @@ def manmask(data, meta, log):
...
     return data
+
+
+# PHOTOMETRY
+def interp_masked(data, meta, i):
+    """
+    Interpolates masked pixels.
+    Based on the example here:
+    https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.griddata.html
+    """
```
taylorbell57, line 565: Please document the Parameters and Returns

src/eureka/lib/util.py
```diff
@@ -553,3 +554,62 @@ def manmask(data, meta, log):
...
+    data.flux.values[i] = grid_z
+
+    return data
+
+
+def phot_arrays(data):
+    """
+    Setting up arrays for the photometry routine.
+    These arrays will be populated by the returns coming from centerdriver.py and apphot.py
+    """
```
taylorbell57, line 596: Please document the parameters and returns

src/eureka/S3_data_reduction/plots_s3.py
```diff
@@ -565,3 +568,307 @@ def median_frame(data, meta):
     plt.savefig(meta.outputdir+fname, dpi=300)
     if not meta.hide_plots:
         plt.pause(0.1)
+
+
+# Photometry
+def phot_lc(meta, data):
+    """
+    Plots the flux as a function of time.
+    """
```
taylorbell57, line 577: Please document the parameters for all these new functions - should just be a matter of copy-pasting things repeatedly

taylorbell57: Can you also put the figure number in the first line of the docstring like the other plots so it's easy to see from the docs pages which functions make which figures

src/eureka/S3_data_reduction/plots_s3.py
```diff
@@ -565,3 +568,307 @@ def median_frame(data, meta):
...
+    plt.tight_layout()
+    fname = (f'figs{os.sep}fig3108-Centroid' + figure_filetype)
+    plt.savefig(meta.outputdir + fname, dpi=250)
+    if not meta.hide_plots:
+        plt.pause(0.2)
+
+
+def phot_npix(meta, data):
+    """
+    Plots the (x, y) centroids and (sx, sy) the Gaussian 1-sigma half-widths as a function of time.
```
taylorbell57, line 650: This docstring is incorrect

src/eureka/S3_data_reduction/plots_s3.py
```diff
@@ -565,3 +568,307 @@ def median_frame(data, meta):
...
+    fname = (f'figs{os.sep}fig3306_file{file_number}_int{int_number}_2D_Frame'
+             + figure_filetype)
+    plt.savefig(meta.outputdir + fname, dpi=250, tight_layout=True)
+    if not meta.hide_plots:
+        plt.pause(0.2)
+
+
+def phot_2d_frame_zoom(meta, m, i, data):
+    """
+    Plots the 2D frame together with the centroid position and apertures.
```
taylorbell57, line 771: Please explain here how this function differs from the one above

taylorbell57: Also, I feel like this function could be merged with the one above with a small number of if/else statements

src/eureka/S3_data_reduction/plots_s3.py
```diff
@@ -565,3 +568,307 @@ def median_frame(data, meta):
...
+    fname = (f'figs{os.sep}fig3504_file{file_number}_int{int_number}_2D_Frame_Zoom'
+             + figure_filetype)
+    plt.savefig(meta.outputdir + fname, dpi=250, tight_layout=True)
+    if not meta.hide_plots:
+        plt.pause(0.2)
+
+
+def phot_2d_frame_oof(meta, m, i, data, flux_w_oof):
+    """
+    Plots the 2D frame together with the centroid position and apertures.
```
taylorbell57, line 823: Please avoid duplicated docstrings like this one - what does this specific function make?

src/eureka/S3_data_reduction/plots_s3.py
```diff
@@ -565,3 +568,307 @@ def median_frame(data, meta):
...
+    fname = (f'figs{os.sep}fig3307_file{file_number}_int{int_number}_2D_Frame_OOF' +
+             figure_filetype)
+    plt.savefig(meta.outputdir + fname, dpi=250, tight_layout=True)
+    if not meta.hide_plots:
+        plt.pause(0.2)
+
+
+def phot_2d_frame_diff(meta, data):
+    """
+    Plots the 2D frame together with the centroid position and apertures.
```
taylorbell57, line 854: Please avoid duplicated docstrings like this one - what does this specific function make?

src/eureka/S3_data_reduction/plots_s3.py
```diff
@@ -565,3 +568,307 @@ def median_frame(data, meta):
...
+    plt.savefig(meta.outputdir + fname, dpi=250, tight_layout=True)
+    if not meta.hide_plots:
+        plt.pause(0.2)
+
+
+def phot_2d_frame_oof(meta, m, i, data, flux_w_oof):
+    """
+    Plots the 2D frame together with the centroid position and apertures.
+    """
+    fig, ax = plt.subplots(2, 1, figsize=(8.2, 4.2))
```
taylorbell57, line 825: You'll also want to set the figure number to 3307 here as well so that the figure gets overridden with each new copy. And then also add a clf after that to make sure the figure gets cleared out before each plot

taylorbell57:
@sebastian-zieba, can you mark the comments you've addressed as resolved, and once you're ready for another review click the "Re-request review" button so that I know when this is ready for one last look-over?

src/eureka/S3_data_reduction/nircam.py
```diff
@@ -209,3 +229,159 @@ def cut_aperture(data, meta, log):
...
+            if edges3[0] <= kk < edges3[1]:
+                continue
+            else:
+                use_cols3[kk] = False
+        flux3 = data.flux.values[i][:, use_cols3]
+        err3 = data.err.values[i][:, use_cols3]
+        mask3 = data.mask.values[i][:, use_cols3]
+
+    if meta.oof_corr == 'meanerr':
+        for j in range(128):
```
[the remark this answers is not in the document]
sebastian-zieba: meanerr is actually already using numpy arrays to calculate the mean (weighted by the uncertainties) of a row.

src/eureka/S3_data_reduction/nircam.py
```diff
@@ -209,3 +229,159 @@ def cut_aperture(data, meta, log):
...
+    star_pos_x_untrim = int(star_pos_x) + meta.xwindow[0]  # position of star before trimming
+    star_exclusion_area_untrim = \
+        np.array([star_pos_x_untrim - meta.oof_bg_dist, star_pos_x_untrim + meta.oof_bg_dist])
+
+    use_cols = np.ones(2048, dtype=bool)
+    for k in range(2048):
+        if star_exclusion_area_untrim[0] <= k < star_exclusion_area_untrim[1]:
+            use_cols[k] = False
+    use_cols = use_cols[meta.xwindow[0]:meta.xwindow[1]]
+    # Array with bools checking if column should be used for background subtraction
```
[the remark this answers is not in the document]
sebastian-zieba: I'm not exactly sure if I understand your point, sorry.
The reason I'm not doing this step in 2D is because the wings of the target on the detector are actually pretty big. So I have found that you shouldn't include anything closer than like 300 pixels from the centroid when you do the 1/f correction. Otherwise, you'd subtract stellar light in this step.
And because ny is way smaller than this area you should avoid, I'm only doing this in 1D.

src/eureka/S3_data_reduction/nircam.py
```diff
@@ -209,3 +229,159 @@ def cut_aperture(data, meta, log):
...
+    # Need to test performance with and without estsig on real data.
+    if hasattr(meta, 'use_estsig') and meta.use_estsig:
+        bgerr = np.median(data.err)
+        estsig = [bgerr for j in range(len(meta.bg_thresh))]
+    else:
+        estsig = None
+
+    nbadpix_total = 0
+    for i in tqdm(range(flux.shape[1]), desc='Looping over Rows for outlier removal'):
+        for j in range(flux.shape[2]):  # Loops over Columns
```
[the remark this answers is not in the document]
sebastian-zieba: I profiled my code and this double loop doesn't seem to be a bottleneck. I also can't think of a better way to write that without making changes to sigrej.sigrej.

sebastian-zieba:
Hey Taylor, I incorporated the majority of the feedback. Thanks!
I added a comment to three little things I didn't change but I still clicked on resolve conversation, so that I can do a re-request.

I also added tests now with the most recent commits. They only run through S3 and S4 because I didn't change S5 at all. 
The dataset for the tests contains just 4 integrations of WASP39b/ERS/NIRCam and has a size of 34 MB. Turns out, that the simulated dataset from the data challenge didn't properly simulate the 1/f noise for NIRCam, so it doesn't represent the actual data. If people think, we shouldn't use actual ERS data for testing here, then I can switch the small calints ERS fits file with a small a fits file based on the data from the data challenge.

I also created a completely new S3 ecf just for the photometry. Earlier all of that was in the S3 NIRCam ecf. There are just so many spectroscopic S3 ecf entries which arent used by the photometry, so having a separate file for the photometry seemed logical to me.

Also, now when doing a run, the directories will get names like how spectroscopy does. E.g., if photap = 65 and skyin = 90 and skyout = 150, then the directory will be called "ap65bg90150".

sebastian-zieba:
Hey @taylorbell57 , do you know if the tests here are not successfully running because of the same reason as in #451 ? The tests are running fine for me locally. 
Seems like they aren't passing for MIRI and NIRSpec. Both of these start at S2. The other tests are going through.

taylorbell57:
I'll get to reviewing this PR later today, but this PR's tests are definitely not failing in the same way as #451 was previously. It's possible that there's some newly released ASDF package version that breaks things at Stage 2 or something - I can try to look into it some.

sebastian-zieba:
Thanks Taylor! Yeah, looking at the errors I also thought about asdf but I checked the asdf version from this failed run with a successful run (Pytest Testing 323) from a few days ago and both seem to use asdf-2.12.0.

src/eureka/lib/util.py
```diff
@@ -481,7 +482,7 @@ def get_mad(meta, log, wave_1d, optspec, optmask=None,
     return np.ma.mean(ediff)
 
 
-def get_mad_1d(data, ind_min=0, ind_max=-1):
```
sebastian-zieba, line 484: I also noticed that **ind_max** is set to -1 here by default. This would exclude the very last integration though because it's used like this: data[ind_min:ind_max]. So I changed this to **None**.

taylorbell57:
> Thanks Taylor! Yeah, looking at the errors I also thought about asdf but I checked the asdf version from this failed run and compared it with a successful run (Pytest Testing 323) from a few days ago and both seem to use asdf-2.12.0.

So I can confirm that it is some new package version that is breaking things - with my normal enviroment the tests all pass, but with a fresh enviroment I get the same ASDF error messages. I've yet to isolate which package is the one causing the issue though

src/eureka/lib/util.py
taylorbell57: Great catch!

taylorbell57:
Alright, I isolated the issue to jsonschema==4.10.0 which for some reason causes the Stage 2 code to break. I've added an upper limit to the setup.cfg which should resolve that issue now

taylorbell57:
> I also added **tests** now with the most recent commits. They only run through S3 and S4 because I didn't change S5 at all. The dataset for the tests contains just 4 integrations of WASP39b/ERS/NIRCam and has a size of 34 MB. Turns out, that the simulated dataset from the data challenge didn't properly simulate the 1/f noise for NIRCam, so it doesn't represent the actual data. If people think, we shouldn't use actual ERS data for testing here, then I can switch the small calints ERS fits file with a small a fits file based on the data from the data challenge.

I think this is probably adequate for our purposes - the simulated data for the other instruments isn't the most representative either.

> I also created a **completely new S3 ecf just for the photometry**. Earlier all of that was in the S3 NIRCam ecf. There are just so many spectroscopic S3 ecf entries which arent used by the photometry, so having a separate file for the photometry seemed logical to me.

I think this was a great choice. Better to keep each template as focused as possible.

> Also, now when doing a run, the **directories will get names like how spectroscopy does**. E.g., if photap = 65 and skyin = 90 and skyout = 150, then the directory will be called "**ap65bg90150**".

I would tweak that a bit for clarity to "ap65bg90_150" as 90150 is hard to read/understand and "-" is not supported by Windows computers

src/eureka/lib/imageedit.py
```diff
@@ -0,0 +1,227 @@
...
+            Value for out of bound pixels in the mask. Default is 0.
+
+    Returns
+    -------
+    Tuple, up to 3 2D-ndarray images containing the extracted image,
+    its mask (if specified) and a uncd array (if specified). The shape
+    of each array is (2*yr+1, 2*xr+1).
+
+    Example
+    -------
```
taylorbell57, line 33: Needs to be "Examples" to be put in the docs

src/eureka/lib/imageedit.py
```diff
@@ -0,0 +1,227 @@
...
+              Semi-length of the extracted image. if not specified,
+              (syc,sxc) will be the center of subim.
+
+    Returns
+    -------
+    The data array with the subim array inserted, according to the
+    given coordinates.
+
+    Example
+    -------
```
taylorbell57, line 135: Needs to be "Examples" to be put in the docs

src/eureka/lib/imageedit.py
```diff
@@ -0,0 +1,227 @@
...
+     [0, 0, 0, 0, 1],
+     [0, 0, 0, 0, 1],
+     [0, 0, 0, 0, 0]]
+
+    Notes
+    -----
+    History:
+
+    2010-07-11  patricio  Added documentation.      pcubillos@fulbrightmail.org
+    2022-08-03  szieba    changed (syc, sxc) is (None, None) to (syc, sxc) == (None, None)
```
taylorbell57, line 180: Not formatted correctly for the docs pages

src/eureka/lib/interp2d.py
```diff
@@ -0,0 +1,46 @@
...
+    ----------
+    image : 2D numpy array
+        Contains the 2D frame which will be oversampled in pixels
+    expand : int
+        The factor by which a pixel should be oversampled.
+        E.g., if set to 5, a pixel will turn into 25 subpixels.
+    y : 1D numpy array
+        np.arange(ny), with ny being the number of pixels in the y direction
+    x : 1D numpy array
+        np.arange(nx), with ny being the number of pixels in the x direction
```
taylorbell57, line 20: "ny" -> "nx"

src/eureka/lib/meanerr.py
```diff
@@ -0,0 +1,106 @@
...
+            Set to True to return a bit flag.  If value=0, result is good.
+            Bits: 0 = NaN(s) in data.
+            1 = some errors equal zero.
+            2 = masked pixel(s) in data.
+
+    Return
+    ------
+    This function returns the error-weighted mean of the unmasked
+    elements of Data. If err or status were set to True, return a
+    tuple.
```
taylorbell57, line 31: Needs to be "Returns" to be put in the docs. Also please format in the same way as the rest of the package (an example is below):

```
Returns
-------
dtype1
    Description of variable1
dtype2
    Description of variable2
```

src/eureka/lib/meanerr.py
```diff
@@ -0,0 +1,106 @@
...
+    Follows maximum likelihood method (see, e.g., Bevington and
+    Robinson 2003, Data Reduction and Error Analysis for the
+    Physical Sciences, 3rd ed, McGraw Hill, Ch. 4.).
+
+    History:
+
+    2005-11-15 jh        Written by: Joseph Harrington, Cornell.
+    jh@oobleck.astro.cornell.edu
+    2010-11-18 patricio  Wrote in python, docstring added.
+    pcubillos@fulbrightmail.org
```
taylorbell57, line 44: Also not formatted correctly for docs

src/eureka/lib/meanerr.py
```diff
@@ -0,0 +1,106 @@
...
+
+    Example
+    -------
```
taylorbell57, line 47: Needs to be Examples

taylorbell57 (changes requested):

docs/media/S3_template.ecf
```diff
@@ -6,12 +6,14 @@ ncpu            1           # Number of CPUs
 nfiles          1           # The number of data files to analyze simultaneously
 max_memory      0.5         # The maximum fraction of memory you want utilized by read-in frames (this will reduce nfiles if need be)
 suffix          calints     # Data file suffix
+photometry      True        # Set to True if the user wants to analyze a photometric dataset
```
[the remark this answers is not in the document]
taylorbell57: I removed these as they do not belong in this docs/media template

src/eureka/lib/chunkbad.py
```diff
@@ -0,0 +1,179 @@
...
+    Written by:	Joseph Harrington, Cornell. 24-11-2005
+    jh@oobleck.astro.cornell.edu
+    29-11-2005 jh       Filled in header.
+    29-05-2007 jh       Handle data with non-N*64 size.
+    29-05-2007 khorning Handle data with only 3 dimensions
+    14-06-2007 khorning Moved chuck size to within the loop
+    16-07-2007 jh       Ensured all loop variables are longs, not ints.
+    13-04-2008 ks       Changed message to print (line 134)
+    13-04-2008 ks       Converted long to string in print (line 134)
+    20-07-2010 patricio Converted to python.
```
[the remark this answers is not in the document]
taylorbell57: This file is also not used anywhere - why add it?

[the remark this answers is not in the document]
taylorbell57: I resolve the issues with the docs, but leaving this open to ask why this file was added

taylorbell57:
Alright, I tried to take care of most of the edits I'd requested and most of the flake8 issues, but there are two files I don't have time to get to today.

taylorbell57:
Also, I tried building the docs on the readthedocs website and it fails for some reason. If you have time to understand why that could have failed, that'd be awesome. Otherwise I'll try to troubleshoot that later

docs/source/ecf.rst
```diff
@@ -168,6 +168,10 @@ As we want to do our own spectral extraction, we set this variable to ``calints`
 .. note::
 	Note that other Instruments might used different suffixes!
 
+photometry
+''''''''''
+Only used for photometry analyses. Set to True if the user wants to analyze a photometric dataset.
+
```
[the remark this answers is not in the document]
sebastian-zieba: The reason why I added that is because of the directory-naming update (like the ap65bg90150 I mentioned above). The directories which use the ap and bg sizes are being created **before** the fits files are being read in. 
I guess, I could instead check if the ecf contains photap, skyin and skyout and then assume that the person is doing photometry... but I am not sure if that's a good solution.

sebastian-zieba:
Oh! I saw your "Tweaking flake8 maxlinelength" commit. Are we doing 80 characters per line as the max instead of 100 now?

src/eureka/S3_data_reduction/nircam.py
```diff
@@ -355,15 +370,19 @@ def corr_oneoverf(data, meta, i, star_pos_x, log):
...
+                    edges_temp = edges_all[k]
+                    data.flux.values[i][j, edges_temp[0]:edges_temp[1]] -= \
+                        me.meanerr(flux_all[k][j], err_all[k][j],
+                                   mask=mask_all[k][j], err=False)
     elif meta.oneoverf_corr == 'median':
         for k in range(4):
             if ampl_used_bool[0]:
-                data.flux.values[i][:, edges_all[k][0]:edges_all[k][1]] -= \
+                edges_temp = edges_all[k]
+                data.flux.values[i][:, edges_temp[0]:edges_temp[1]] -= \
```
sebastian-zieba, line 381: Hey @taylorbell57 , out of curiosity, any reason you changed 
data.flux.values[i][:, edges_all[k][0]:edges_all[k][1]] to 
edges_temp = edges_all[k] and then 
data.flux.values[i][:, edges_temp[0]:edges_temp[1]] ?

sebastian-zieba:
> Also, I tried building the docs on the readthedocs website and it fails for some reason. If you have time to understand why that could have failed, that'd be awesome. Otherwise I'll try to troubleshoot that later

I actually just compiled it (Build # 17820310) and it ran successfully. I'm not sure why the ones from yesterday were failing apparently.

sebastian-zieba:
> Alright, I tried to take care of most of the edits I'd requested and most of the flake8 issues, but there are two files I don't have time to get to today.

I fixed the flake8 issues in s3_reduce.py and plots_s3.py now. They were all line length issues

src/eureka/S3_data_reduction/nircam.py
```diff
@@ -48,16 +50,31 @@ def read(filename, data, meta, log):
...
     err = hdulist['ERR', 1].data
     dq = hdulist['DQ', 1].data
     v0 = hdulist['VAR_RNOISE', 1].data
-    wave_2d = hdulist['WAVELENGTH', 1].data
     int_times = hdulist['INT_TIMES', 1].data
 
+    if hdulist[0].header['CHANNEL'] == 'LONG':  # Spectroscopy will have "LONG" as CHANNEL
+        meta.photometry = False
+        wave_2d = hdulist['WAVELENGTH', 1].data
+    elif hdulist[0].header['CHANNEL'] == 'SHORT':  # Photometry will have "SHORT" as CHANNEL
```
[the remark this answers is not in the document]
sebastian-zieba: > Is there no way to get spectropscopy in the short channel using some different observing setup? If there is, then this would break for those data. There must be some other setting in the FITS header which says whether you're looking at spectra or photometry.

That's a great point. As far as I know, NIRCam's Grisms are only in the 2.4–5.0 µm area (so Long Wavelength) using either F277W, F322W2, F356W or F444W. So ['CHANNEL'] == 'SHORT' (i.e. wavelength <= 2.3 micron) will **always** be photometry. But I guess someone could use **just** use the LW light to do photometry and don't do spectroscopy at all? 

So I think the most robust solution would be to check hdulist[0].header['FILTER'] and if that's one of the following: F277W, F322W2, F356W or F444W it **has** to be spectroscopy. In any other case, it's photometry.

Here are some helpful websites:
- https://jwst-docs.stsci.edu/jwst-near-infrared-camera/nircam-instrumentation/nircam-grisms
- https://jwst-docs.stsci.edu/jwst-near-infrared-camera/nircam-instrumentation/nircam-filters

sebastian-zieba:
> > Also, now when doing a run, the **directories will get names like how spectroscopy does**. E.g., if photap = 65 and skyin = 90 and skyout = 150, then the directory will be called "**ap65bg90150**".
> 
> I would tweak that a bit for clarity to "ap65bg90_150" as "90150" is hard to read/understand and "90-150" is not supported by Windows computers

I was considering this but I was afraid the code was expecting that the entries in meta.bg_hw_range have to be integers. 

But thankfully, it looks like using a string as meta.bg_hw does not create problems. The photometry code isn't using meta.spec_hw and meta.bg_hw either way but meta.photap, meta.skyin and meta.skyout instead. So meta.spec_hw_range and meta.bg_hw_range are only really used for directory stuff in the photometry routine.

src/eureka/lib/chunkbad.py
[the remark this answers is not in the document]
sebastian-zieba: I just removed the file, it's not used currently. Thanks!

docs/source/ecf.rst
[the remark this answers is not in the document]
taylorbell57: Hmm, I see your point. What you have works for now, so let's just stick with that for the time being

taylorbell57:
> Oh! I saw your "Tweaking flake8 maxlinelength" commit. Are we doing 80 characters per line as the max instead of 100 now?

Yeah, the flake8 default is 80 and that was always the intended value, but their docs misled me to think the flake8_linting.yml maxlinelength setting was controlling the length of the log prints from the automated linting, not the maximum line length of the code it was linting.

src/eureka/S3_data_reduction/nircam.py
taylorbell57: Yeah, was just to make things fit better with the 80 character line limit. Without that, it was going to get really awkward to get it all compliant.

taylorbell57 (approved):
Alright, while non-ideal in my mind to have the big `if not photometry: ...  else: ...` blocks in the main S3 function, it works for now and I know Kevin wants this merged sooner than later

Issue #436, opened 2022-08-01
[Bug]: Issue with finding source position if it is close to the edge of the detector


lauraflagg:
### FAQ check

- [X] Yes, I checked the FAQ and my question has not been addressed.

### Instrument

NIRSpec (Stages 1-3)

### What happened?

For G395H, because the trace is curved, sometimes the initial source position is fairly close to the edge of the detector and the spec_hw puts it over the edge of the detector.  So, when pos_max-meta.spec_hw is calculated, it returns a negative number, obviously results in the code breaking when the negative number is used to index the array.  

I was able to fix this fairly easily by checking if pos_max-meta.spec_hw is less than 0  (or meta.ywindow[0]) and if so, setting it to 0 (or meta.ywindow[0]).

### Error traceback output

Traceback (most recent call last):
  File "/Users/lflagg/OneDrive - Cornell University/Eureka/demos/WASP39b_G395/run_eureka_nrs1.py", line 29, in <module>
    s3_spec, s3_meta = s3.reduce(eventlabel, ecf_path=ecf_path)
  File "/Users/lflagg/anaconda3/envs/eureka/lib/python3.9/site-packages/eureka/S3_data_reduction/s3_reduce.py", line 287, in reduce
    data, meta, log = source_pos.source_pos_wrapper(data, meta,
  File "/Users/lflagg/anaconda3/envs/eureka/lib/python3.9/site-packages/eureka/S3_data_reduction/source_pos.py", line 99, in source_pos_wrapper
    meta.src_ypos = source_pos(flux[integ], meta, data.attrs['shdr'],
  File "/Users/lflagg/anaconda3/envs/eureka/lib/python3.9/site-packages/eureka/S3_data_reduction/source_pos.py", line 160, in source_pos
    src_ypos, src_ywidth = source_pos_gauss(flux, meta, m, n, plot)
  File "/Users/lflagg/anaconda3/envs/eureka/lib/python3.9/site-packages/eureka/S3_data_reduction/source_pos.py", line 359, in source_pos_gauss
    print(np.ma.max(sum_row))
  File "/Users/lflagg/anaconda3/envs/eureka/lib/python3.9/site-packages/numpy/ma/core.py", line 6716, in max
    return obj.max(axis=axis, fill_value=fill_value, out=out, **kwargs)
  File "/Users/lflagg/anaconda3/envs/eureka/lib/python3.9/site-packages/numpy/ma/core.py", line 5837, in max
[...]

PR #440, opened 2022-08-02
Bug fix for bug #436, plus smoothing trace option for G395 

lauraflagg:
This is based on working with the G395H data.

The bug fix for bug #436 is confined to source_pos.py

The remaining edits are to force a smooth trace using a spline.  Currently, a hot pixel may result in the wrong center of mass when straightening the spectrum.  This is a robust way to avoid that.  

The below files are cuts taken from two versions of the flux files from S3.

Without the smooth trace:
image

With the smooth trace:
image

### DIFF
--- a/demos/JWST/S3_nirspec_fs_template.ecf
+++ b/demos/JWST/S3_nirspec_fs_template.ecf
@@ -29,6 +29,7 @@ p7thresh        60          # X-sigma threshold for outlier rejection during opt
 
 # G395H curvature treatment
 curvature       None        # How to manage the curved trace on the detector (Options: None, correct). Use correct for NIRSpec/G395.
+smooth_trace    False       # Force a smooth trace
 
 # Diagnostics
 isplots_S3      3           # Generate few (1), some (3), or many (5) figures (Options: 1 - 5)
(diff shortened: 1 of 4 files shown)


kevin218:
Great minds think alike! Just submitted a PR (#441) that solves the same problem using similar techniques (median filter vs. spline).  Both should work just fine.  I also added a figure (3106).

I don't have a strong preference on which method we use, but let's not trip over ourselves with these two PRs.  Maybe the user should be allowed to specify whether to smooth or interpolate in the ECF.

taylorbell57:
How about we merge one of these PRs into the other to allow either smoothing or interpolation depending on some meta parameter. And then once that's done, we can merge into main?

kevin218:
I needed to compute a "clean" median frame (interpolated over bad pixels) for optimal spectral extraction and ended up making smoothing during curvature correction unnecessary.  It still happens in my PR (for now), but the final result is the same as just computing the shifts on the clean median frame.

kevin218:
I'd like to merge the bug fix to #436 without the code for smoothing. Can you resubmit with the requested changes?

taylorbell57:
Alright, closing this PR as the issue was resolved in a similar manner by PR #441

PR #449, opened 2022-08-07
fix manual clip bug in S5

lkreidberg:
this PR fixes a bug in S5 for manually clipping data points. The lc and meta arrays are now updated when the manual_clip function is called in s5_fit.py

### DIFF
--- a/src/eureka/S5_lightcurve_fitting/s5_fit.py
+++ b/src/eureka/S5_lightcurve_fitting/s5_fit.py
@@ -135,7 +135,7 @@ def fitlc(eventlabel, ecf_path=None, s4_meta=None):
 
             if hasattr(meta, 'manual_clip') and meta.manual_clip is not None:
                 # Remove requested data points
-                util.manual_clip(lc, meta, log)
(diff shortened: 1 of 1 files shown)


taylorbell57:
Out of curiosity, did this actually not work as is, or were you worried it wouldn't based on reading the code? I would've expected it to work just fine since python passes variables by reference into functions, so the functions are editing the original copies of the variables and we shouldn't need to return meta, lc, or log (but I have been just to help communicate to the reader that they've been changed). If this actually doesn't work as it is in the main branch, then we will need to do some careful checking elsewhere to make sure a similar bug isn't happening in other places

lkreidberg:
it didn't work! I was surprised, because I also thought the variables should have been updated. Not sure what happened here, but after the explicit assignment everything was fine!

sebastian-zieba:
Yeah, I had exactly the same problem like Laura

taylorbell57:
Huh, that's super strange! Thanks for the patch!

taylorbell57 (approved):

PR #441, opened 2022-08-02
Smooth curvature, compute clean median flux

kevin218:
Curvature plot was sensitive to bad pixels, so I smoothed the results first.  Also added a new plot (3106) to demonstrate the measured curvature.  Also, fixed a few loose ends from previous PR.

PR #450, opened 2022-08-09
Add Fig 3106, 3304, 3401 to docs/media + minor changes

kevin218:
In addition to adding/replacing figures in docs/median, this PR includes minor updates based on feedback from #441.


taylorbell57:
I'm working on Sebastian's PR right now so it'll be a little bit (~1 day maybe) before I can review this, but some other things I've noticed while trying to use the new `clean_median_flux` function:

1. I want to be able to set in the median_thresh parameter in the ECF
2. I want to be able to try without smoothing, so I suggest skipping the smoothing related steps if window_len <= 1 or window_len is None
3. I find it is important to smooth over masked pixels from the DQ array before smoothing and interpolating because some part of the smoothing/interpolating code does not appear to be aware of masked arrays and ignores the mask.

The local version of the code that I have come up with so far is:
```
def clean_median_flux(data, meta, log):
    """Computes a median flux frame that is free of bad pixels.

    Parameters
    ----------
    data : Xarray Dataset
        The Dataset object.
    meta : eureka.lib.readECF.MetaClass
        The metadata object.
    log : logedit.Logedit
        The current log.

    Returns
    -------
    data : Xarray Dataset
        The updated Dataset object.

    Notes
    -----
    History:

    - 2022-08-03, KBS
        Inital version
    """
    log.writelog('  Computing clean median frame...', mute=(not meta.verbose))

    # Create mask using all data quality flags
    mask = np.copy(data.mask.values)
    mask[data.dq.values > 0] = 0

    # Compute median flux using masked arrays
    flux_ma = np.ma.masked_where(mask == 0, data.flux.values)
    medflux = np.ma.median(flux_ma, axis=0)
    ny, nx = medflux.shape

    # Interpolate over masked pixels
    clean_med = np.zeros((ny, nx))
    xx = np.arange(nx)
    for j in range(ny):
        x1 = xx[~medflux.mask[j]]
        goodmed = medflux[j][~medflux.mask[j]]
        f = spi.interp1d(x1, goodmed, 'linear', fill_value='extrapolate')
        # f = spi.UnivariateSpline(x1, goodmed, k=1, s=None)
        clean_med[j] = f(xx)
    medflux = np.ma.masked_array(clean_med)

    # Apply smoothing filter
    if meta.window_len > 1:
        smoothflux = np.zeros((ny, nx))
        for j in range(ny):
            smoothflux[j] = smooth.medfilt(medflux.data[j], meta.window_len)

        # Compute residuals
        residuals = medflux - smoothflux

        # Flag outliers
        if not hasattr(meta, 'median_thresh'):
            meta.median_thresh = 5
        outliers = sigma_clip(residuals, sigma=meta.median_thresh, maxiters=5,
                              axis=1, cenfunc=np.ma.median, 
                              stdfunc=np.ma.std).mask

        # Interpolate over bad pixels
        clean_med = np.zeros((ny, nx))
        xx = np.arange(nx)
        for j in range(ny):
            x1 = xx[~outliers[j]]
            goodmed = medflux[j][~outliers[j]]
            f = spi.interp1d(x1, goodmed, 'linear', fill_value='extrapolate')
            # f = spi.UnivariateSpline(x1, goodmed, k=1, s=None)
            clean_med[j] = f(xx)

    data['medflux'] = (['y', 'x'], clean_med)
    data['medflux'].attrs['flux_units'] = data.flux.attrs['flux_units']

    if meta.isplots_S3 >= 4:
        plots_s3.median_frame(data, meta)

    return data
```

src/eureka/S3_data_reduction/optspex.py
```diff
@@ -480,7 +480,7 @@ def profile_gauss(subdata, mask, threshold=10, guess=None, isplots=0):
     return profile
 
 
-def clean_median_flux(data, meta, log, median_thresh=5):
+def clean_median_flux(data, meta, log, m, median_thresh=5):
```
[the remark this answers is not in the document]
taylorbell57: See also my other comments about this function

taylorbell57 (changes requested):

taylorbell57 (changes requested):

src/eureka/S3_data_reduction/optspex.py
```diff
@@ -511,40 +512,57 @@ def clean_median_flux(data, meta, log, median_thresh=5):
 
     # Create mask using all data quality flags
     mask = np.copy(data['mask'].values)
-    mask[np.where(data.dq > 0)] = 1
+    if meta.dqmask:
+        mask[np.where(data.dq % 2 == 1)] = 0
```
taylorbell57, line 516: No need to do this since it is already done in the main s3_reduce function

src/eureka/S3_data_reduction/optspex.py
```diff
@@ -511,40 +512,57 @@ def clean_median_flux(data, meta, log, median_thresh=5):
...
+        # Interpolate over masked regions
+        interp_med = np.zeros((ny, nx))
+        xx = np.arange(nx)
+        for j in range(ny):
+            x1 = xx[~np.ma.getmaskarray(medflux[j])]
+            goodrow = medflux[j][~np.ma.getmaskarray(medflux[j])]
+            f = spi.interp1d(x1, goodrow, 'linear',
+                             fill_value='extrapolate')
+            # f = spi.UnivariateSpline(x1, goodmed, k=1, s=None)
+            interp_med[j] = f(xx)
```
taylorbell57, line 533: This needs to be done even if `meta.window_len == 1` (my MIRI case). If you found that there was no issue in doing this for your data, then we can just move it before the if statement. If it is detrimental for your data, then it could be moved to the else part of this if/else statement

src/eureka/S3_data_reduction/optspex.py
```diff
@@ -511,40 +512,57 @@ def clean_median_flux(data, meta, log, median_thresh=5):
...
+                              axis=1, cenfunc='median')
+
+        # Interpolate over bad pixels
+        clean_med = np.zeros((ny, nx))
+        xx = np.arange(nx)
+        for j in range(ny):
+            x1 = xx[~np.ma.getmaskarray(outliers[j]) *
+                    ~np.ma.getmaskarray(medflux[j])]
+            goodrow = medflux[j][~np.ma.getmaskarray(outliers[j]) *
+                                 ~np.ma.getmaskarray(medflux[j])]
```
taylorbell57, line 552: Multiplying booleans like this gives the result of an `and` statment - is that your intention? I'd think you'd want an `or`

taylorbell57 (changes requested):

src/eureka/S3_data_reduction/optspex.py
kevin218: I had the same thought, and yet it made a difference with/without it when dqmask = True.  Quite unexpected.

taylorbell57: Oh yikes, that's worth investigating then as that should definitely not be happening

src/eureka/S3_data_reduction/optspex.py
kevin218: Let's see if I've got my logic straight.  `getmaskarray` returns zeros for good pixels and `~` converts those to ones.  So now we're multiplying two arrays where 1=good and 0=bad.  In this case, we want multiplicate because 1*0 should be 0, meaning if either of the masks is flagged as bad then the data are bad.

taylorbell57: Aha yes, you're right - I had tried to double-check my thinking in ipython, but it turns out the behaviour of `~False` is not the same as `~np.array([False])`. The former gives `-1` while the latter gives `np.array([True])`. Confusing! But you're all good

src/eureka/S3_data_reduction/optspex.py
kevin218: I ran some more tests and everything appears to work as expected, so I've now removed the unnecessary code.

taylorbell57 (approved):

PR #464, opened 2022-09-12
Minor S5-6 improvements

taylorbell57:
This PR adds minor improvements to Stages 5 and 6.

- Fixed Eureka version number on docs page
- Saving Stage 5 emcee/dynesty chains using Astraeus instead of hdf5 for easier use (safely backwards compatible)
- Allow setting dynesty's run_nlive to 'min' as an easy shortcut for the minimum reasonable live points
- Added a new Stage 5 figure to help show phase variations
- Added first order phase curve amplitude and offset as plotted/tabulated parameters in Stage 6.
- Added second order phase curve amplitude and offset as plotted/tabulated parameters in Stage 6.
- Added nightside flux as plotted/tabulated parameter in Stage 6


src/eureka/lib/util.py
```diff
@@ -364,6 +364,18 @@ def find_fits(meta):
     return meta
 
 
+def binData(data, nbin=60, err=False):
```
kevin218, line 367: Lacks documentation

kevin218 (changes requested):

src/eureka/lib/util.py
taylorbell57: Sorry, this is a draft PR and still in prep - I was going to submit it and then decided to add a plot showing the phase variations in this same PR so I just left this as a draft. I'll mark this as ready for you in a day or two when I'm finished and everything is tested and documented

src/eureka/S5_lightcurve_fitting/fitters.py
```diff
@@ -790,7 +805,9 @@ def dynestyfitter(lc, model, meta, log, **kwargs):
     log.writelog('Running dynesty...')
 
     min_nlive = int(np.ceil(ndims*(ndims+1)//2))
-    if nlive < min_nlive:
+    if nlive == 'min':
+        nlive = min_nlive
+    elif nlive < min_nlive:
```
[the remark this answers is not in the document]
taylorbell57: I don't think it truly is a strict limit, but a recommended threshold from dynesty.

taylorbell57:
Alright @kevin218, this is ready for your review now

src/eureka/S3_data_reduction/straighten.py
```diff
@@ -154,6 +154,7 @@ def straighten_trace(data, meta, log, m):
 
     # apply the shifts to the data
     data.flux.values = roll_columns(data.flux.values, shifts)
+    data.mask.values = roll_columns(data.mask.values, shifts)
```
kevin218, line 157: Wow, how did I miss this?!?!

src/eureka/S3_data_reduction/s3_reduce.py
```diff
@@ -472,19 +472,16 @@ def reduce(eventlabel, ecf_path=None, s2_meta=None):
                             data['nskyideal'][i], data['status'][i],
                             data['betaper'][i]) = aphot
 
+                # Delete unsaveable FITS data
+                del (data.attrs['filename'], data.attrs['mhdr'],
```
kevin218, line 476: I think we solved this problem, so there should no longer be any need to delete these attributes woth that latest version.

taylorbell57: The code before this commit worked fine unless you don't save the FluxData and then try to save the SpecData. I need to do that because the GO data I'm working on ended up consuming ~1 TB with FluxData save files while I was doing a parameter exploration...

kevin218: Is there a better solution that doesn't always delete these parameters? I'm on my phone, so I can't look at it easily.

taylorbell57: Sure, I had just figured those FITS header details weren't worth saving, but I'll just make another quick commit to try to save them when possible and delete them if not

taylorbell57: Alright, done

kevin218 (approved):

taylorbell57:
The MIRI S6 tests kept occasionally failing on every PR (depending on how the mini S5 run went), so I just added a small tweak to protect against that. Hopefully that'll fix the failing tests

kevin218 (approved):

taylorbell57:
Well the test still failed, but it passes locally. Looking at the test log, it looks like S5 started on the 16th and S6 started on the 17th (must be 6 hours east of Pacific). It ran all the way through but then the test failed just because it didn't know to look for the S6 folder under a new date. I'm just going to merge this PR as is since that was a pretty flukey bug.

PR #469, opened 2022-10-06
Minor convenience improvements

taylorbell57:
This PR does the following:

1. Fills the MIRI wavelength array in Stage 3 if the extract_1d step was skipped in Stage 2 (which greatly speeds up Stage 2). With our current version of the jwst pipeline, the wavelength array only gets created if extract_1d is run and even then it just gets filled with zeros. This Stage 3 tweak takes care of both of those situations and will end up doing nothing once we upgrade to a jwst version that properly fills the wavelength array in Stage 2.
2. Adds a few more wavelengths to the NIRCam photometery code to allow different observing parameters
3. Reduces the number of repeated log statements for photometry (now only logged for the first integration per segment which keeps the noise down).
4. Allows the `nplots` ECF parameter to work for the photometry figures 3307 and 3505 which greatly speeds up S3 photometry and reduces the number of uninteresting figures generated. Updated the ECF documentation page accordingly.
5. Fixes a bug where changing the path to an S3 folder would break the loading of the S3 save file.
6. Adds a feature which I've needed for a while now. I store all my analyses on an external hard drive, so if I accidentally unplug that drive or need to unplug it to transport the computer while an analysis is paused, then the entire analysis ends up crashing because the log file gets forcefully closed. I added a small tweak where a failure to print to the log is attempted to be fixed by reconnecting to the log file (assuming I've now reconnected my drive) and if that fails it just prints a warning to the terminal instead and continues on without crashing the code just for the log file (and will try to reconnect to the log file each time we log something). This has already saved me once which was a useful but unintentional test of the tweak.
7. flake8 tweaks

### DIFF
--- a/docs/source/ecf.rst
+++ b/docs/source/ecf.rst
@@ -367,7 +367,7 @@ Sets how many plots should be saved when running Stage 3. A full description of 
 
 nplots
 ''''''
-Sets how many integrations will be used for per-integration figures (Figs 3301, 3302, 3303, 3501). Useful for in-depth diagnoses of a few integrations without making thousands of figures. If set to None, a plot will be made for every integration.
+Sets how many integrations will be used for per-integration figures (Figs 3301, 3302, 3303, 3307, 3501, 3505). Useful for in-depth diagnoses of a few integrations without making thousands of figures. If set to None, a plot will be made for every integration.
 
 vmin
 ''''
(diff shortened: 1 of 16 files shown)


taylorbell57:
Of course the test passes when I run it locally and fails when it runs on GitHub... It looks like yet another issue with the S6 code that rounds to the correct number of significant figures for pretty LaTeX tables. The code works fine when you get a good fit but often fails during these tests when we're doing just a super fast and terrible fit to make sure things don't break. I'll still work on a quick fix for this bug though

kevin218 (approved):

PR #474, opened 2022-10-13
JOSS review edits and NIRCam wavelength

### LINKED ISSUE #455: [Enhancement]: Wavelength solution for NIRCam
### Instrument

NIRCam (Stages 1-3)

### What is your suggestion?

Adding the NIRCam wavelength solution that has been used by the NIRCam team as the JWST CRDS one is not accurate at the blue end of the spectrum.

### Error traceback output

_No response_

_No response_

_No response_

_No response_

### Code of Conduct

### LINKED ISSUE #472: Questions and suggestions wrt Quickstart tutorial
### Instrument

_No response_

### What happened?

As part of https://github.com/openjournals/joss-reviews/issues/4503, I'm going through the "Quickstart" tutorial, and here I'm including comments and suggestions.

- I think it would be useful to include a high-level overview of how Eureka works either at the beginning of this document, or on the index page, or both. In particular, I mean that this tutorial suggests that the standard workflow will be to write configuration files and then run a command line script, rather than interfacing with the Python API directly. For example, is it expected that all users would start by copying the demo project and then edit the `.ecf` files for their own purposes, or is the demo specific to these example data? I think it would be very useful to provide more context here, rather than just jumping in.
- The demo `run_eureka.py` script has some path manipulation (`sys.path.insert(0, f'..{os.sep}..{os.sep}')`) which is not required given that the tutorial expects that Eureka is installed and that we're supposed to be working in a separate directory when running the quickstart. This could also cause issues with import paths etc. depending on where the working directory is located. I think it should be removed.
- It would be useful to be clear within the tutorial that the `inputdir` and `outputdir` are both relative to `topdir` (I know that the demo files say this, but it should be in the tutorial text too) because `/Stage2` (for example) is not a great path to see! I'd say that overall this interface feels clunky: is there a reason why `topdir` can't be defined once and then reused in the other stages with the subdirectory names can be set automatically?
- When running `run_eureka.py`, there were a handful of warnings (deprecation warnings, sqrt of negative number, etc.) buried in the logs (the full log is included below). Are these anything to be worried about? The tutorial should mention what to expect here and when not to worry.
- When comparing to the results figures, I don't get exactly the same results. All of the numbers (e.g. MAD) in the titles are somewhat different, and the axis limits in some of the figures vary significantly. If differences are expected (why?) when running with the same datasets and following the steps exactly as given, the tutorial should make it clear that we expect these differences, and how much difference is too much. Overall we just want to be able to check that things ran ok when following the tutorial on a local machine.

Full output log from `run_eureka.py`: https://gist.github.com/dfm/266b66102b0ed57c360ec3da17762c08

### Error traceback output

_No response_

macOS 12.6 (M1)

3.9.7

_No response_

### Code of Conduct

taylorbell57:
This PR does the following:

Resolves #472, resolves #455

Specifically, this PR:
1. Updates quickstart figures using current code version so that users don't worry about differences between their results and those on the quickstart docs.
2. Reduces the number of warnings and adds note about expected warnings in quickstart tutorial.
3. Carries the centroid_sx from S3 photometry through to S4 (previously was forgotten, so it was unavailable in Stage 5).
4. Implements the NIRCam longwave wavelength solutions using a polynomial fit computed by Everett Schlawin (@eas342) for F322W2 and F444W. These can be used by setting meta.poly_wavelength=True, otherwise the values in the FITS files will be used. This new meta parameter has been documented in the ECF docs and is assumed to be False if not put in the ECF.


dfm:
This looks great to me! I can confirm that the updated figures are consistent with those that I got when running the Quickstart locally. 🎉

kevin218 (approved):
Looks good!

PR #475, opened 2022-10-14
Write S5 LC files to a single Xarray file in Stage 6

kevin218:
The single HDF5 file makes it easier to share spectroscopic LCs and fits with colleagues or make available online.


taylorbell57:
There was just a minor flake8 issue, but the pytest also failed. Your new code assumes that there will be `meta.n_int` integrations, but this is not the case if the new `manual_clip` feature is used in Stage 5 (as is the case in the MIRI test). You're likely better off just using the astropy table to tell you how many integrations there were rather than relying on `meta.n_int`. The other option is that we edit `meta.n_int` in Stage 5 to reflect the reduced number of integrations after manual clipping. We could also do both of those - up to you

kevin218:
@taylorbell57 I took the easy route and simply queried the data.  In the end, it's the most robust solution.  I am also having trouble with test_MIRI.py.  The problem seems to be that Eureka! isn't using the latest version of Astraeus.  That should be an easy fix...

taylorbell57:
Hah, my own PR is having isssues with the MIRI test, and that issue is being driven by a NaN/inf issue in S6 that should be quick to resolve

taylorbell57:
Ah no, your issue with the MIRI test is the same as mine

taylorbell57:
I think I just fixed the MIRI test issue on my PR (https://github.com/kevin218/Eureka/pull/474) - the test runs successfully there at least. You can merge main into this branch once that PR gets approved and merged, and hopefully that'll fix the issue with your test. As of right now, the test here on GitHub crashed before it ran your new edits, so I don't want to approve this yet without this PR's test going through successfully. That PR of mine was the one that had the updated NIRCam wavelength solution FYI

kevin218:
Yeah, agreed on making sure the tests pass first.

taylorbell57:
Gah, there was some CRDS error in the Stage 2 MIRI test. This is most likely a sporadic issue with their file serving and will probably be resolved soon (if not already). I'll try to rerun this test in a bit

kevin218:
Pytest passed yesterday after I updated the Astraeus version number, so it should pass again once the CRDS error goes away

taylorbell57 (approved):

PR #482, opened 2022-11-03
Quick bokeh bug fix

meganmansfield:
I found that upon doing a clean install with v0.6 I was getting the following error about bokeh not having some required packages set up the right way:

ImportError: cannot import name 'Panel' from 'bokeh.models.widgets' 

It looks like this problem is because the default install now installs bokeh v3.0.1, while previously Eureka! had been automatically installing v2.4.3. I'm not sure if it definitely needs to be v2.4.3 or just <v3.0, but I added a strict requirement for the version. We can change that to an upper limit if you think it would be better though!

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -7,7 +7,7 @@ dependencies:
     - asdf
     - astropy
     - batman-package
-    - bokeh==2.4.3
+    - bokeh=2.4.3
     - ccdproc
     - celerite
     - corner
--- a/environment.yml
+++ b/environment.yml
@@ -7,7 +7,7 @@ dependencies:
     - asdf
     - astropy
     - batman-package
-    - bokeh=2.4.3
+    - bokeh[version='<3.0']
     - ccdproc
     - celerite
     - corner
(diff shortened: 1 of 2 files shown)


kevin218:
Pytest still seems to be failing, despite the fix. Any ideas why?

taylorbell57:
Strange that this didn't get caught before - I guess they must've just released a new version that breaks things for us. Sigh...

meganmansfield:
From the history in pytest, it looks like it's still installing version 3.0.1. Is there another place besides the environments.yml file that I need to make a change to get it to pay attention to the version number? Did I use the wrong syntax somewhere? Sorry, I'm not too familiar with environments files.

meganmansfield:
Thanks! That fixed the bokeh issue, but it's now throwing an issue that is completely unrelated to the package install I just added. Not sure what's going on there...

taylorbell57:
Ah, that's most likely caused by temporary CRDS outages. On occasion it just fails to download the files it needs from CRDS and gives error messages like that. Let's try rerunning these tests in a bit when CRDS is probably back online and hopefully they'll run through successfully

kevin218 (approved):
Assuming pytest passes this time...

PR #484, opened 2022-11-07
Enable GLBS for MIRI, fix GLBS plotting

kevin218:
This PR enables group-level background subtraction for MIRI.  However, the spectra are cleaner when GLBS is **NOT** used, so I wouldn't recommend people apply this step with MIRI data.  

Fortunately, I was able to get plots_s3.image_and_background() working when using GLBS, so that's a small win.

### DIFF
--- a/demos/JWST/S4_template.ecf
+++ b/demos/JWST/S4_template.ecf
@@ -9,6 +9,9 @@ wave_min        1.5         # Minimum wavelength. Set to None to use the shortes
 wave_max        4.5         # Maximum wavelength. Set to None to use the longest extracted wavelength from Stage 3.
 allapers        False       # Run S4 on all of the apertures considered in S3? Otherwise will use newest output in the inputdir
 
+# Mannualy mask pixel columns by index number
+# mask_columns    []
+
 # Parameters for drift correction of 1D spectra
 recordDrift     False    # Set True to record drift/jitter in 1D spectra (always recorded if correctDrift is True)
 correctDrift    False   # Set True to correct drift/jitter in 1D spectra (not recommended for simulated data)
--- a/environment.yml
+++ b/environment.yml
@@ -34,7 +34,7 @@ dependencies:
         - crds
         - exotic-ld
         - image_registration @ git+https://github.com/keflavich/image_registration.git@master#egg=image_registration
-        - jwst==1.6.0
+        - jwst==1.8.0
         - myst-parser # Needed for documentation
         - sphinx-rtd-theme # Needed for documentation
         - stcal
--- a/src/eureka/S1_detector_processing/group_level.py
+++ b/src/eureka/S1_detector_processing/group_level.py
@@ -38,8 +38,17 @@ def GLBS(input_model, log, meta):
     dq = input_model.groupdq
 
     meta.inst = input_model.meta.instrument.name.lower()
-    meta.int_start = 0
     meta.n_int = all_data.shape[0]
+    meta.int_start = 0
+    if not hasattr(meta, 'nplots') or meta.nplots is None:
+        meta.int_end = meta.n_int
+    elif meta.int_start+meta.nplots > meta.n_int:
+        # Too many figures requested, so reduce it
+        meta.int_end = meta.n_int
+    else:
+        meta.int_end = meta.int_start+meta.nplots
+    if meta.inst == 'miri':
+        meta.isrotate = False
 
     for ngrp in range(all_data.shape[1]):
         grp_data = all_data[:, ngrp, :, :]
@@ -61,10 +70,11 @@ def GLBS(input_model, log, meta):
         xrdict = dict(flux=xrdata, mask=xrmask)
         data = xrio.makeDataset(dictionary=xrdict)
         data['flux'].attrs['flux_units'] = 'n/a'
+        data.attrs['intstart'] = meta.intstart
         if ngrp == all_data.shape[1]-1:
             data = bkg.BGsubtraction(data, meta, 
                                      log, 0, 
-                                     meta.isplots)
+                                     meta.isplots_S1)
         else:
             # do not show plots except for the last group
             data = bkg.BGsubtraction(data, meta, 
--- a/src/eureka/S3_data_reduction/background.py
+++ b/src/eureka/S3_data_reduction/background.py
@@ -288,7 +288,7 @@ def fitbg(dataim, meta, mask, x1, x2, deg=1, threshold=5, isrotate=False,
                     plt.plot(goodxvals, dataslice, 'bo')
                     plt.plot(range(nx), bg[j], 'g-')
                     fname = ('figs'+os.sep+'Fig3601_BG_'+str(j) +
-                             plots.plots.figure_filetype)
+                             plots.figure_filetype)
                     plt.savefig(meta.outputdir + fname, dpi=300)
                     plt.pause(0.01)
 
--- a/src/eureka/S3_data_reduction/background.py
+++ b/src/eureka/S3_data_reduction/background.py
@@ -261,11 +261,11 @@ def fitbg(dataim, meta, mask, x1, x2, deg=1, threshold=5, isrotate=False,
                     residuals = dataslice - model
                     # Simple standard deviation (faster but prone to missing
                     # scanned background stars)
-                    # stdres = np.std(residuals)
+                    stdres = np.std(residuals)
                     # Median Absolute Deviation (slower but more robust)
                     # stdres  = np.median(np.abs(np.ediff1d(residuals)))
                     # Mean Absolute Deviation (good compromise)
-                    stdres = np.mean(np.abs(np.ediff1d(residuals)))
+                    # stdres = np.mean(np.abs(np.ediff1d(residuals)))
                     if stdres == 0:
                         stdres = np.inf
                     stdevs = np.abs(residuals) / stdres
@@ -281,7 +281,7 @@ def fitbg(dataim, meta, mask, x1, x2, deg=1, threshold=5, isrotate=False,
             # background image
             if len(goodxvals) != 0:
                 bg[j] = np.polyval(coeffs, range(nx))
-                if isplots >= 6:
+                if isplots == 6:
                     plt.figure(3601)
                     plt.clf()
                     plt.title(str(j))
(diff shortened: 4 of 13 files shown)


environment.yml
```diff
@@ -34,7 +34,7 @@ dependencies:
         - crds
         - exotic-ld
         - image_registration @ git+https://github.com/keflavich/image_registration.git@master#egg=image_registration
-        - jwst==1.6.0
+        - jwst==1.8.0
```
taylorbell57, line 37: This same change should be made to environment_pymc3.yml

src/eureka/S3_data_reduction/background.py
```diff
@@ -261,11 +261,11 @@ def fitbg(dataim, meta, mask, x1, x2, deg=1, threshold=5, isrotate=False,
                     residuals = dataslice - model
                     # Simple standard deviation (faster but prone to missing
                     # scanned background stars)
-                    # stdres = np.std(residuals)
+                    stdres = np.std(residuals)
                     # Median Absolute Deviation (slower but more robust)
                     # stdres  = np.median(np.abs(np.ediff1d(residuals)))
                     # Mean Absolute Deviation (good compromise)
-                    stdres = np.mean(np.abs(np.ediff1d(residuals)))
+                    # stdres = np.mean(np.abs(np.ediff1d(residuals)))
```
taylorbell57, line 268: Does this change have a postive, negative, or neutral effect on the Stage 3 background subtraction?

src/eureka/S3_data_reduction/background.py
```diff
@@ -281,14 +281,14 @@ def fitbg(dataim, meta, mask, x1, x2, deg=1, threshold=5, isrotate=False,
             # background image
             if len(goodxvals) != 0:
                 bg[j] = np.polyval(coeffs, range(nx))
-                if isplots >= 6:
+                if isplots == 6:
```
taylorbell57, line 284: Why not `>=`? That's how all the other isplots are handled, right?

src/eureka/S3_data_reduction/miri.py
```diff
@@ -284,7 +285,16 @@ def fit_bg(dataim, datamask, n, meta, isplots=0):
...
+    if hasattr(meta, 'isrotate'):
+        isrotate = meta.isrotate
+    else:
+        isrotate = 2
+    bg, mask = background.fitbg(dataim, meta, datamask, meta.bg_y1,
+                                meta.bg_y2, deg=meta.bg_deg,
+                                threshold=meta.p3thresh, isrotate=isrotate,
+                                isplots=isplots)
+    return bg, mask, n
+    # return nircam.fit_bg(dataim, datamask, n, meta, isplots=isplots, **kwargs)
```
taylorbell57, line 297: Can delete this commended out line

src/eureka/S3_data_reduction/optspex.py
```diff
@@ -784,7 +784,9 @@ def optimize(meta, subdata, mask, bg, spectrum, Q, v0, p5thresh=10,
                             plt.figure(3803)
                             plt.clf()
                             plt.suptitle(str(i) + "/" + str(nx))
-                            plt.plot(subdata[:, i], 'bo')
+                            plt.errorbar(np.arange(ny), subdata[:, i],
+                                         np.sqrt(variance[:, i]),
```
taylorbell57, line 788: Does this end up being xerr or yerr? I believe the ordering of the keywords is xerr first, but I might be wrong. I always specify `yerr=` since I can't remember which error it assumes you've input if you only input one error

src/eureka/S3_data_reduction/optspex.py
```diff
@@ -534,17 +534,27 @@ def clean_median_flux(data, meta, log, m):
...
+        outliers = sigma_clip(residuals, sigma=meta.median_thresh,
+                              maxiters=None, axis=1, cenfunc='median')
+
+        #FINDME: Need to move this figure to plots_s3.py
+        # plt.figure(2, figsize=(8, 4))
+        # plt.clf()
+        # plt.title("Outliers")
+        # plt.imshow(outliers, origin='lower', aspect='auto')
+        # plt.colorbar()
+        # plt.pause(1)
```
taylorbell57, line 557: Is this commended out code still needed?

src/eureka/S1_detector_processing/group_level.py
```diff
@@ -38,8 +38,17 @@ def GLBS(input_model, log, meta):
...
+    meta.int_start = 0
+    if not hasattr(meta, 'nplots') or meta.nplots is None:
+        meta.int_end = meta.n_int
+    elif meta.int_start+meta.nplots > meta.n_int:
+        # Too many figures requested, so reduce it
+        meta.int_end = meta.n_int
+    else:
+        meta.int_end = meta.int_start+meta.nplots
+    if meta.inst == 'miri':
+        meta.isrotate = False
```
taylorbell57, line 51: Why not rotate the MIRI frames?

src/eureka/S3_data_reduction/miri.py
```diff
@@ -284,7 +285,16 @@ def fit_bg(dataim, datamask, n, meta, isplots=0):
     n : int
         The current integration number.
     """
-    return nircam.fit_bg(dataim, datamask, n, meta, isplots=isplots)
+    if hasattr(meta, 'isrotate'):
+        isrotate = meta.isrotate
+    else:
+        isrotate = 2
```
taylorbell57, line 291: The data is already rotated by this point in Stage 3, so setting isrotate to 2 by default isn't the correct behaviour, right?

src/eureka/S3_data_reduction/plots_s3.py
```diff
@@ -369,7 +375,7 @@ def profile(meta, profile, submask, n, m):
         plt.pause(0.2)
 
 
-def subdata(meta, i, n, m, subdata, submask, expected, loc):
+def subdata(meta, i, n, m, subdata, submask, expected, loc, variance):
```
taylorbell57, line 378: This new parameter needs to be added to the docstring

src/eureka/S3_data_reduction/plots_s3.py
```diff
@@ -1100,3 +1108,46 @@ def phot_2d_frame_diff(data, meta):
...
+               interpolation='nearest')
+    plt.ylabel('Relative Pixel Position')
+    plt.xlabel('Relative Pixel Position')
+    plt.colorbar()
+    plt.tight_layout()
+    file_number = str(m).zfill(int(np.floor(np.log10(meta.num_data_files))+1))
+    int_number = str(n).zfill(int(np.floor(np.log10(meta.n_int))+1))
+    fname = (f'figs{os.sep}fig3506_file{file_number}_int{int_number}' +
+             f'_Std_Dev_Profile'+plots.figure_filetype)
+    plt.savefig(meta.outputdir + fname, dpi=200)
```
taylorbell57, line 1153: Do you want to add a `plt.pause(0.2)` if `not meta.hideplots` like with the other plots?

src/eureka/S3_data_reduction/straighten.py
```diff
@@ -35,7 +35,9 @@ def find_column_median_shifts(data, meta, m):
                    np.sum(data, axis=0))
 
     # Smooth CoM values to get rid of outliers
-    smooth_coms = smooth.medfilt(column_coms, meta.window_len)
+    smooth_coms = smooth.medfilt(column_coms, 11)
```
taylorbell57, line 38: Why force a window length of 11?

taylorbell57 (changes requested):
In addition to the specific comments below, make sure that any flake8 errors are addressed

src/eureka/S3_data_reduction/background.py
kevin218: It makes the BG calculation faster; the effect is neutral.

src/eureka/S3_data_reduction/background.py
kevin218: I was testing parts of the code that required isplots > 6 and didn't want to sit and wait for the code to create lots of plots that I didn't care about.  In general, it makes sense to use `==` for values of 6+ because people won't be investigating different parts of the code at the same time.

src/eureka/S1_detector_processing/group_level.py
kevin218: `isrotate` is true for the NIR instruments, so it needs to be false for MIRI.  For some background info on the matter... I originally wrote `background.fitbg()` for an instrument that dispersed light along the y axis (like MIRI) and then modified it using the `isrotate` command for instruments that disperse light along x.

src/eureka/S3_data_reduction/miri.py
kevin218: `miri.fit_bg()` is now called in Stages 1 and 3.  When called in Stage 3, the default behavior is to use `isrotate = 2` for the reasons mentioned in the conversation above.  When called in Stage 1, the data do not need to be rotated, so the use should be using `meta.isrotate = False`.

src/eureka/S3_data_reduction/straighten.py
kevin218: We were using `meta.window_len` for multiple purposes and it became a problem.  In order to test `optspex.clean_median_flux()` with window_len = 1, I could not have `smooth.medfilt()` in straighten.py also using a window length of 1.  After some thought, I concluded that we will never need to change the smoothing window length, so I fixed it to 11.

demos/JWST/S4_template.ecf
```diff
@@ -9,6 +9,9 @@ wave_min        1.5         # Minimum wavelength. Set to None to use the shortes
 wave_max        4.5         # Maximum wavelength. Set to None to use the longest extracted wavelength from Stage 3.
 allapers        False       # Run S4 on all of the apertures considered in S3? Otherwise will use newest output in the inputdir
 
+# Mannualy mask pixel columns by index number
+# mask_columns    []
```
taylorbell57, line 13: When we update the demos ECFs, we should also update the ECFs in docs/media. I see that the S1 file in that folder also didn't get updated when the GLBS was first added, so if you could update that ECF at the same time that'd be great.

Also, please document this change in the docs/source/ecf.rst file

taylorbell57 (changes requested):

demos/JWST/S4_template.ecf
kevin218: I'm updating the docs/media ECFs and ecf.rst file as part of the Bias Correction PR (#507).

taylorbell57 (approved):

Issue #314, opened 2022-05-09
[Enhancement]: Add more astrophysical models


taylorbell57:
### Instrument

Light curve fitting (Stages 4-6)

### What is your suggestion?

The astrophysical models that we mentioned during our meeting today that we'd like to at least investigate adding to Eureka are:

- [x] catwoman (for asymmetric transits)
- [ ] spiderman (for phase curves and eclipse mapping); may not be worthwhile given how extremely time-consuming it is. If we do add it, we should sparsely compute it and interpolate in time
- [x] starry (for transits, eclipse mapping, and phase curves. This depends on pymc3; I have already started adding starry while working on Issue #265)
- [ ] eigencurves/eigenmapping (Megan's code for eclipse mapping)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

taylorbell57:
FYI, I now have well functioning and modular code for starry+pymc3 that just needs tests and documentation. I'll try to have a PR submitted for that before the end of the month in prep for the MIRI ERS observations

PR #492, opened 2022-11-29
New PyMC3 Fitter, New Starry Astrophysical Model, and Differentiable Systematics Models

### LINKED ISSUE #265: Add PyMC3 as a S5 sampling method
Related to Issue #192 and #223, I have been told that PyMC3 and other Hamiltonian Monte Carlo sampling methods are among the best/only methods of fitting shared fits with potentially hundreds of fitted parameters. PyMC3 requires that models be written in a very different way than for lsq, emcee, or dynesty, so adding it will require a non-trivial amount of work and will require that each astrophysical/detector model be implemented separately for PyMC3. If this is indeed the best/only option for doing shared fits though, then that seems worthwhile for me. I'll be making a first attempt at adding in a pymc3fitter over the next few days and will see how well it can handle shared fits.

taylorbell57:
This PR:

Resolves #265 with the addition of pymc3 as an *optional* dependency
Works on #314 with the addition of starry as an *optional* dependency

In addition, this PR:

- Packages pymc3, starry, and related packages together under the `pymc3` optional dependency collection
- Improves the formatting of the ecf.rst documentation page
- Adds clarity on how to install collections of optional dependencies
- Copies the input `input_meta` and `s#_meta` variables to ensure they are not changed by the next stage
- Fixes some issues with WFC3's drift_2D function and separate_direct function
- Fixes a bug with the calculation of meta.wave_low and meta.wave_hi
- Makes sure filename_S3_SpecData and filename_S4_LCData aren't written over during mergeevent
- Duplicates all non-differentiable systematic models (aside from the GP model) as differentiable models that can be used with pymc3. In the future I'd like to merge these implementations, but this was non-trivial and I didn't want to delay the addition of this important improvement
- Adds starry as the only differentiable astrophysical model. This one model handles transits, eclipses, eclipse mapping, and phase curves.
- Adds the PyMC3 No U-Turns Sampler (NUTS - called nuts in the ECF) and the differentiable gradient descent optimization as implemented in the exoplanet package (called exoplanet in the ECF). These are added in gradient_fitters.py while the old fitters remain in fitters.py. This also comes with a new Figure 5305 which shows the NUTS trace as 1D histograms and temporal walker evolution.
- Adds a new force_positivity ECF setting to Stage 5 to allow users to force positive sinusoid_pc phase curves.
- Adds a test for the exoplanet and nuts fitters using a partial phase curve fit to the synthetic MIRI eclipse observations.

Future improvements will need to include:
- Ideally merging differentiable and non-differentiable code to avoid duplicated code
- Better handling shared fits to not exceed the maximum number of fitted parameters in PyMC3. PyMC3 accepts parameters that are arrays so this shouldn't be too hard, but it'll take some work to rework the formatting from things like `rp_0` to `rp[0]`
- Add an example figure for Figure 5305 to the documentation page

In my experience, PyMC3 NUTS fits have been as fast or significantly faster than dynesty fits, especially when not fitting for eclipse mapping signals or phase variations which are more computationally intensive. I have encountered the occasional issue where quite large datasets can't have two chains fitted in parallel and instead need to have the chains run in series, but I don't believe this could be an issue with our code.

I have successfully tested the code with and without pymc3 and related packages installed on my computer. I see no flake8 issues on my computer. All new functions have docstrings and new ECF and EPF parameters have been documented.


kevin218 (changes requested):
Clearly a lot of work went into this PR. Nicely done!

src/eureka/S5_lightcurve_fitting/__init__.py
```diff
@@ -1,7 +1,23 @@
...
+    import starry
+    import pymc3_ext
+    import arviz
+    success = True
+except ImportError:
+    success = False
+    print("Could not import starry and/or pymc3 related packages. "
+          "Functionality may be limited.")
+if success:
+    from . import differentiable_models
```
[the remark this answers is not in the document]
taylorbell57: I hesitated to catch any and all exceptions with importing this as it might hide some other issue. Looking more carefully at what throws `ImportError`s though, it seems it only gets thrown if "Import can't find module, or can't find name in module." which should only catch cases when one of the dependencies isn't installed (and not some other issue with our code). So in summary, I will do as you suggested.

taylorbell57:
Alright, your comments have been addressed, but in trying to use the code on what data we currently have for WASP-43b I found that compute_fn in Stage 6 didn't work for starry fits and then noticed that my simple tweak doesn't actually calculate the nightside hemisphere's flux but instead the anti-stellar point's flux, so I'll need to edit that function a bit more before this PR can be merged.

taylorbell57:
Alright, as long as the tests pass here (already checked on my laptop), I think I've properly taken care of the compute_fn function for starry fits (and the compute_fp was already right because of how I setup the starry model).

It's not really clear what pc_amp and pc_offset should be for a generic spherical harmonic map, but I think I'll stick to just Y10 and Y11 for the pc_amp and pc_offset term and Y20 and Y22 for the pc_amp2 and pc_offset2 terms. I'll need to tweak the compute_amp terms as they're again calculated for the map and not the phase curve. Since I'm only using those Y10, Y11, Y20, and Y22 terms though, I'm just going to use Cowan+2008's simple but exact equation 7. I'll try to get this done tomorrow

kevin218:
I have a more robust solution in mind for the amplitude and offset that doesn't depend on the functional form. It involves generating a series of phase curves from the parameter distributions and solving for the max.

taylorbell57:
Yeah, so that's what I did to find the nightside flux. I guess there's two different things that we're talking about though. First, I and others might want to know the amplitude and offset of the first and second order terms separately (for easier comparison with other datasets or analyses where higher order terms weren't fitted). Second, one might want to know the peak-to-trough amplitude as well as the maximum and minimum offsets of the overall combined phase curve with all the sinusoid or Ylm terms together. I guess we should probably support both - we just need clear names for all the possible options

taylorbell57:
The other headache is that this currently only takes into account the terms that were fitted and ignores any terms set to white_fixed or fixed...

taylorbell57:
That last commit I just made makes the behaviour of pc_amp from starry models consistent with that for batman models. I'd suggest leaving changes to both models' pc_amp calculations to a future PR. I need to locally test for a potential bug with higher order models I encountered while trying to fit WASP-43b, and after that is addressed or unreproducible I'll ping you here for your updated review.

taylorbell57:
Alright, I figured out that I myself was just doing something wrong and that the code is fine as is. This PR is ready for your review again, @kevin218

src/eureka/S6_planet_spectra/s6_spectra.py
```diff
@@ -801,20 +907,103 @@ def compute_fn(meta, log, fit_methods, nsamp=1e4):
...
+    planet_map = starry.Map(ydeg=meta.ydeg, nw=nsamp)
+    planet_map2 = starry.Map(ydeg=meta.ydeg, nw=nsamp)
     for i in range(meta.nspecchan):
-        fluxes = fp[i]*(1-2*ampcos[i])
+        inds = np.random.randint(0, len(fp[i]), nsamp)
+        for ell in range(1, meta.ydeg+1):
+            for m in range(-ell, ell+1):
+                if hasattr(temp, f'Y{ell}{m}'):
+                    planet_map[ell, m, :] = getattr(temp, f'Y{ell}{m}')[i]
+                    planet_map2[ell, m, :] = getattr(temp, f'Y{ell}{m}')[i]
```
kevin218, line 1003: I don't see the need for planet_map2, but I trust you've thought this through.

kevin218 (approved):

src/eureka/S6_planet_spectra/s6_spectra.py
taylorbell57: This snippet is copied from Stage 5, so I'll explain the motivation for writing it that way for Stage 5. Because the code gets compiled (aka this for-loop isn't run at every step in the fit - only once before the fit is run), I was worried about circular references with setting the map's amplitude to a value that depends on it's amplitude. Specifically, `planet_map.flux()` depends on the value of `planet_map.amp`, so setting `planet_map.amp` based on `planet_map.flux()` with this compiled theano code had me worried about what would happen in practice as it would likely depend on the order in which the variables are updated. I just tried removing `planet_map2` though and it doesn't seem to fundamentally break things, but comparing two tests with `planet_map2` and two tests without it, the fit without `planet_map2` seemed to have a much harder time converging (many more divergences and large discrepancies between independent chains) and took longer to complete. Since there's some intrinsic variability in the path taken from one run to the next, it's maybe possible that the fit without `planet_map2` just followed an especially unlucky path twice, but I think that unlikely and either way it seems clear that there's no significant disadvantage to leaving things the way they currently are

PR #502, opened 2023-02-14
Add full frame outlier rejection capabilities

kevin218:
Instead of only performing outlier rejection along the time axis within the BG region, this added functionality allows the user to perform outlier rejection on the entire frame.  This step (when enabled) takes place before computing the median frame, so has the potential to yield a more reliable median frame.  The risk is that, for deep transits, this routine could flag in-transit points as bad. So, this method should be used with caution.

I have updated the demos and test directories, but not the html file.

### DIFF
--- a/demos/JWST/S3_miri_lrs_template.ecf
+++ b/demos/JWST/S3_miri_lrs_template.ecf
@@ -15,9 +15,9 @@ record_ypos     True        # Option to record the y position and width for each
 dqmask          True        # Mask pixels with an odd entry in the DQ array
 
 # Outlier rejection along time axis
-ff_outlier      False       # Set True to use full frame; set False to use only background region
-ff_thresh       [5,5]       # Full Frame double-iteration X-sigma threshold for outlier rejection along time axis
-bg_thresh       [5,5]       # Background region double-iteration X-sigma threshold for outlier rejection along time axis
+ff_outlier      False       # Set False to use only background region (recommended for deep transits)
+                            # Set True to use full frame (works well for shallow transits/eclipses)
+bg_thresh       [4,4]       # Double-iteration X-sigma threshold for outlier rejection along time axis
 
 # Background parameters
 bg_hw           10          # Half-width of exclusion region for BG subtraction (relative to source position)
--- a/demos/JWST/S3_miri_lrs_template.ecf
+++ b/demos/JWST/S3_miri_lrs_template.ecf
@@ -14,9 +14,13 @@ src_pos_type    gaussian    # Determine source position when not given in header
 record_ypos     True        # Option to record the y position and width for each integration (only records if src_pos_type is gaussian)
 dqmask          True        # Mask pixels with an odd entry in the DQ array
 
+# Outlier rejection along time axis
+ff_outlier      False       # Set True to use full frame; set False to use only background region
+ff_thresh       [5,5]       # Full Frame double-iteration X-sigma threshold for outlier rejection along time axis
+bg_thresh       [5,5]       # Background region double-iteration X-sigma threshold for outlier rejection along time axis
+
 # Background parameters
 bg_hw           10          # Half-width of exclusion region for BG subtraction (relative to source position)
-bg_thresh       [5,5]       # Double-iteration X-sigma threshold for outlier rejection along time axis
 bg_deg          0           # Polynomial order for column-by-column background subtraction, -1 for median of entire frame
 p3thresh        5           # X-sigma threshold for outlier rejection during background subtraction
 
--- a/demos/JWST/S3_nircam_wfss_template.ecf
+++ b/demos/JWST/S3_nircam_wfss_template.ecf
@@ -12,11 +12,16 @@ ywindow         [5,64]      # Vertical axis as seen in DS9
 xwindow         [100,1700]  # Horizontal axis as seen in DS9
 src_pos_type    gaussian    # Determine source position when not given in header (Options: header, gaussian, weighted, max, or hst)
 record_ypos     True        # Option to record the y position and width for each integration (only records if src_pos_type is gaussian)
+dqmask          False       # Mask pixels with an odd entry in the DQ array
+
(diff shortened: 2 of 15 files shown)


taylorbell57:
Is this PR ready for my review?

kevin218:
> 

Yeah, I think now would be a good time to review it.

taylorbell57:
Just merging in the now-updated main branch so I can tell what was added in this PR vs your earlier PR

tests/MIRI_ecfs/S3_MIRI.ecf
```diff
@@ -12,9 +12,13 @@ src_pos_type gaussian   # Determine source position when not given in header (Op
 record_ypos  True        # Option to record the y position and width for each integration (only records if src_pos_type is gaussian)
 dqmask       True        # Mask pixels with an odd entry in the DQ array
 
+# Outlier rejection along time axis
+# ff_outlier      False # Set True to use full frame; set False to use only background region
```
taylorbell57, line 16: Why is this commented out?

demos/JWST/S3_nircam_wfss_template.ecf
```diff
@@ -12,11 +12,16 @@ ywindow         [5,64]      # Vertical axis as seen in DS9
 xwindow         [100,1700]  # Horizontal axis as seen in DS9
 src_pos_type    gaussian    # Determine source position when not given in header (Options: header, gaussian, weighted, max, or hst)
 record_ypos     True        # Option to record the y position and width for each integration (only records if src_pos_type is gaussian)
+dqmask          False       # Mask pixels with an odd entry in the DQ array
```
taylorbell57, line 15: Shouldn't this be set to True?

demos/JWST/S3_nirspec_fs_template.ecf
```diff
@@ -14,10 +14,14 @@ src_pos_type    gaussian    # Determine source position when not given in header
 record_ypos     True        # Option to record the y position and width for each integration (only records if src_pos_type is gaussian)
 dqmask          True        # Mask pixels with an odd entry in the DQ array
 
+# Outlier rejection along time axis
+ff_outlier      True        # Set True to use full frame; set False to use only background region
```
taylorbell57, line 18: Is this intentionally set to True only for the NIRSpec template? Given the risk of sigma-clipping away a deep transit, I feel like it might be better to have this be False by default. It's probably also worth putting a terse word of caution in all these ECFs as well.

demos/JWST/S3_miri_lrs_template.ecf
```diff
@@ -14,9 +14,13 @@ src_pos_type    gaussian    # Determine source position when not given in header
 record_ypos     True        # Option to record the y position and width for each integration (only records if src_pos_type is gaussian)
 dqmask          True        # Mask pixels with an odd entry in the DQ array
 
+# Outlier rejection along time axis
+ff_outlier      False       # Set True to use full frame; set False to use only background region
+ff_thresh       [5,5]       # Full Frame double-iteration X-sigma threshold for outlier rejection along time axis
```
taylorbell57, line 19: These two new options need to be documented in the ECF template in docs/media/ as well as the docs/source/ecf.rst file. You should also make it clear in those docs that bg_thresh isn't used if ff_outlier is set to True and ff_thresh is used instead. Actually, since the two thresholds have the same purpose and only one of them is ever used at the same time, why have two different names? Why not just use meta.bg_thresh inside the new ff_outlier algorithm?

src/eureka/S3_data_reduction/nirspec.py
```diff
@@ -143,6 +138,44 @@ def flag_bg(data, meta, log):
     return data
 
 
+def flag_ff(data, meta, log):
+    '''Outlier rejection of full frame along time axis.
+    For data with deep transits, there is a risk of masking good transit data.
+    Proceed with caution.
```
taylorbell57, line 144: This is currently an exact clone of the nircam.py function, so you should just call that function like you've done in miri.py

taylorbell57 (changes requested):

src/eureka/S3_data_reduction/s3_reduce.py
```diff
@@ -342,7 +347,11 @@ def reduce(eventlabel, ecf_path=None, s2_meta=None, input_meta=None):
 
                     # Perform outlier rejection of
                     # sky background along time axis
-                    data = inst.flag_bg(data, meta, log)
+                    meta.bg_y2 = meta.src_ypos + meta.bg_hw
+                    meta.bg_y1 = meta.src_ypos - meta.bg_hw
+                    if (not hasattr(meta, 'ff_outlier')
+                            or meta.ff_outlier is False):
```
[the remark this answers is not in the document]
taylorbell57: If meta.ff_outlier was set to None, then the `if meta.ff_outlier` above wouldn't be triggered and this `if meta.ff_outlier is False` would also not be triggered. But `if not meta.ff_outlier` would be triggered (annoyingly complicated, I know...)

demos/JWST/S3_nircam_wfss_template.ecf
kevin218: No, NIRCam seems to perform better when `dqmask` is set to False.

demos/JWST/S3_nirspec_fs_template.ecf
kevin218: Agreed

demos/JWST/S3_miri_lrs_template.ecf
kevin218: Good point.  I removed ``ff_thresh`` completely and added descriptions to the ecf.rst file.

demos/JWST/S3_nircam_wfss_template.ecf
taylorbell57: Huh, surprising but good to know

taylorbell57 (approved):

PR #505, opened 2023-02-17
Allowing custom gain value, sinusoid_pc with PyMC3 models in S5, and bug fixes

taylorbell57:
This PR:
- Allows users to provide their own gain value to Stage 3 (useful for MIRI where the value on CRDS is/was very wrong)
- Allows the use of the sinusoid_pc model with the theano-based differentiable versions fit with PyMC3. This allows for faster fitting of phase variations compared to the starry spherical harmonics as it ignores eclipse mapping signals. This can be useful when you expect no eclipse mapping signals or want to compare fits with and without accounting for eclipse mapping signals to ensure they're significant
- Updates the image_registration package to 0.2.6 (now that it's on PyPI)
- Fixes an issue with determining the maximum order for spherical harmonics that caused a crash when using odd-m Ylm terms and also excessively slowed the code when there were only even-m Ylm terms
- Fixes issues with the differentiable models related to shared fits
- Forces the eccentricity to be between [0,1] for PyMC3 models (otherwise the code crashes)
- Fixes an issue with the batman eclipse model that's only present when there is a non-zero eccentricity
- Fixes a bug with saving S5 outputs in the savetable_S5 function
- Fixes the calculation of the offsets and amplitudes in Stage 6 for PyMC3 models

There are no currently open issues that this PR closes.

### DIFF
--- a/demos/JWST/S1_miri_template.ecf
+++ b/demos/JWST/S1_miri_template.ecf
@@ -28,7 +28,7 @@ jump_rejection_threshold  4.0 #float, default is 4.0, CR sigma rejection thresho
 
 # Custom linearity reference file
 custom_linearity    False
-superbias_file      /path/to/custom/linearity/fits/file
+linearity_file      /path/to/custom/linearity/fits/file
 
 # Custom bias frames
 custom_bias         False
--- a/demos/JWST/S1_miri_template.ecf
+++ b/demos/JWST/S1_miri_template.ecf
@@ -26,6 +26,21 @@ skip_gain_scale     False
 #Pipeline stages parameters
 jump_rejection_threshold  4.0 #float, default is 4.0, CR sigma rejection threshold
 
+# Custom linearity reference file
+custom_linearity    False
+superbias_file      /path/to/custom/linearity/fits/file
+
+# Custom bias frames
+custom_bias         False
+superbias_file      /path/to/custom/superbias/fits/file
+
+# Saturation
+update_sat_flags    False   # Wheter to update the saturation flags more aggressively
+expand_prev_group   False   # Expand saturation flags to previous group
+dq_sat_mode         percentile # Options: [percentile, min, defined]
+dq_sat_percentile   50      # Percentile of the entire time series to use to define the saturation mask (50=median)
+dq_sat_columns      [[0, 0], [0,0], [0,0], [0,0], [0,0]]  #for dq_sat_mode = defined, user defined saturated columns
+
 # Project directory
 topdir              /home/User/Data/JWST-Sim/MIRI/
 
--- a/demos/JWST/S1_nirx_template.ecf
+++ b/demos/JWST/S1_nirx_template.ecf
@@ -29,9 +29,12 @@ jump_rejection_threshold  4.0 #float, default is 4.0, CR sigma rejection thresho
 custom_linearity    False
 linearity_file      /path/to/custom/linearity/fits/file
 
-# Custom bias frames
+# Custom bias
+bias_correction     None    # Bias correction options: [mean, group_level, smooth, None]
+bias_group          1       # Group number options: [1, 2, ..., each]
+bias_smooth_length  201     # Window length when using 'smooth' bias correction
 custom_bias         False
-superbias_file      /path/to/custom/superbias/fits/file
+superbias_file	    /path/to/custom/superbias/fits/file
 
 # Saturation
 update_sat_flags    False   # Wheter to update the saturation flags more aggressively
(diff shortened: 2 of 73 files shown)


demos/JWST/S1_miri_template.ecf
```diff
@@ -26,6 +26,21 @@ skip_gain_scale     False
 #Pipeline stages parameters
 jump_rejection_threshold  4.0 #float, default is 4.0, CR sigma rejection threshold
 
+# Custom linearity reference file
+custom_linearity    False
+superbias_file      /path/to/custom/linearity/fits/file
```
kevin218, line 31: Should this be linearity_file?

demos/JWST/S1_nirx_template.ecf
```diff
@@ -25,39 +25,43 @@ skip_gain_scale     False
 #Pipeline stages parameters
 jump_rejection_threshold  4.0 #float, default is 4.0, CR sigma rejection threshold
 
-#Custom bias frames
+# Custom linearity reference file
+custom_linearity    False
+superbias_file      /path/to/custom/linearity/fits/file
```
kevin218, line 30: Should this be linearity_file?

docs/media/S1_template.ecf
```diff
@@ -25,6 +25,44 @@ skip_gain_scale     False
 #Pipeline stages parameters
 jump_rejection_threshold  4.0 #float, default is 4.0, CR sigma rejection threshold
 
+# Custom linearity reference file
+custom_linearity    False
+superbias_file      /path/to/custom/linearity/fits/file
```
kevin218, line 30: Should this be linearity_file?

src/eureka/S3_data_reduction/bright2flux.py
```diff
@@ -83,31 +83,36 @@ def dn2electrons(data, meta):
     nx = data.attrs['mhdr']['SUBSIZE1']
     ny = data.attrs['mhdr']['SUBSIZE2']
 
-    # Load gain array in units of e-/ADU
-    gain_header = fits.getheader(meta.gainfile)
-    xstart_gain = gain_header['SUBSTRT1']
-    ystart_gain = gain_header['SUBSTRT2']
+    if hasattr(meta, 'gain') and meta.gain is not None:
```
kevin218, line 86: If I want to specify meta.gain as a 2D array, would I have to write it out in the ECF or is there an easier way?

src/eureka/S5_lightcurve_fitting/differentiable_models/SinusoidPhaseCurve.py
```diff
@@ -0,0 +1,162 @@
...
+            if self.starry_model is not None:
+                systems = self.starry_model.systems
+
+        lcfinal = lib.zeros(0)
+        for c in range(nchan):
+            chan = channels[c]
+            phaseVars = lib.ones(len(self.time))
+            
+            # Compute orbital phase
+            # FINDME: for now this assumes a circular orbit
```
kevin218, line 135: Is this here for future work?

kevin218 (changes requested):

src/eureka/S3_data_reduction/bright2flux.py
taylorbell57: At present yes, you'd have to manually enter the whole array in the ECF. The one thing I've thought of to allow easier manual input of array gains is allowing the user to set `meta.gainfile` instead of forcing users to use the current best file from CRDS. I could also add a check if the extension is .fits (as would be expected if using a CRDS gain file) or if the file extension is .csv (and then do some special processing). I don't think there is currently any demand for user-supplied 2D gain values though and allowing CSVs wouldn't be completely trivial (with complexities coming from subarrays), so I'd be more inclined to just allow the user to manually set `meta.gainfile` to a .fits file (and they can edit the FITS values if they want a fully custom array).

src/eureka/S5_lightcurve_fitting/differentiable_models/SinusoidPhaseCurve.py
taylorbell57: Ah yes, although the fact of the matter is that sinusoidal phase variations only really make sense where the eccentricity is quite small (where the current approximation isn't terrible). The issue here is that you cannot analytically solve for the true anomaly of an eccentric orbit, so I'm not sure that it's possible to do the same thing I did for the numpy-based model in this theano-based model. I'll see what I can do though

kevin218 (approved):

PR #507, opened 2023-02-22
Bias Correction

kevin218:
This PR modifies/overrides the JWST pipeline's superbias step (within Stage 1) to apply a scale factor correction.  There are numerous options available in the ECF that can be applied, but the general procedure is the same.  The code always computes a scale factor (SF) correction as follows: SF = (median of group)/(median of superbias), using a background region that is `meta.expand_mask`  pixels from the measured trace.  Depending on `meta.bias_correction` (options include: mean, group_level, smooth, None) `and meta.bias_group` (options include: 1, 2, ..., each), the code applies various corrections to the superbias before subtracting from the data.

### DIFF
--- a/docs/source/ecf.rst
+++ b/docs/source/ecf.rst
@@ -37,15 +37,15 @@ If True, skip the named step.
 
 bias_correction
 '''''''''''''''''
-Method applied to correct the superbias using a scale factor (SF).  Here, SF = (median of group)/(median of superbias), using a background region that is ``expand_mask`` pixels from the measured trace.  The option ``None`` applies no correction; ``group_level`` computes SF for every integration in ``bias_group``; ``smooth`` applies a smoothing filter of length ``bias_smooth_length`` to the ``group_level`` SF values; and ``mean`` uses the mean SF over all integrations.  We currently recommend using ``smooth`` with a ``bias_smooth_length`` that is ~15 minutes.
+Method applied to correct the superbias using a scale factor (SF) when no bias pixels are available (i.e., with NIRSpec).  Here, SF = (median of group)/(median of superbias), using a background region that is ``expand_mask`` pixels from the measured trace.  The default option ``None`` applies no correction; ``group_level`` computes SF for every integration in ``bias_group``; ``smooth`` applies a smoothing filter of length ``bias_smooth_length`` to the ``group_level`` SF values; and ``mean`` uses the mean SF over all integrations.  For NIRSpec, we currently recommend using ``smooth`` with a ``bias_smooth_length`` that is ~15 minutes.
 
 bias_group
 '''''''''''''''''
-Integer or string.  Specifies which group number should be used when applying the bias correction.  We currently recommend using the first group (``bias_group`` = 1).  There is no group 0.  Users can also specify ``each``, which computes a unique bias correction for each group.
+Integer or string.  Specifies which group number should be used when applying the bias correction.  For NIRSpec, we currently recommend using the first group (``bias_group`` = 1).  There is no group 0.  Users can also specify ``each``, which computes a unique bias correction for each group.
 
 bias_smooth_length
 '''''''''''''''''
-Integer. When ``bias_correction = smooth``, this value is used as the window length during smoothing.
+Integer. When ``bias_correction = smooth``, this value is used as the window length during smoothing across integrations.
 
 custom_bias
 '''''''''''''''''
--- a/docs/source/ecf.rst
+++ b/docs/source/ecf.rst
@@ -35,6 +35,18 @@ If True, skip the named step.
 .. note::
    Note that some instruments and observing modes might skip a step either way! See the `calwebb_detector1 docs <https://jwst-pipeline.readthedocs.io/en/latest/jwst/pipeline/calwebb_detector1.html>`__ for the list of steps run for each instrument/mode by the STScI's JWST pipeline.
 
+bias_correction
+'''''''''''''''''
+Method applied to correct the superbias using a scale factor (SF).  Here, SF = (median of group)/(median of superbias), using a background region that is ``expand_mask`` pixels from the measured trace.  The option ``None`` applies no correction; ``group_level`` computes SF for every integration in ``bias_group``; ``smooth`` applies a smoothing filter of length ``bias_smooth_length`` to the ``group_level`` SF values; and ``mean`` uses the mean SF over all integrations.  We currently recommend using ``smooth`` with a ``bias_smooth_length`` that is ~15 minutes.
+
+bias_group
+'''''''''''''''''
+Integer or string.  Specifies which group number should be used when applying the bias correction.  We currently recommend using the first group (``bias_group`` = 1).  There is no group 0.  Users can also specify ``each``, which computes a unique bias correction for each group.
+
+bias_smooth_length
+'''''''''''''''''
+Integer. When ``bias_correction = smooth``, this value is used as the window length during smoothing.
+
 custom_bias
 '''''''''''''''''
 Boolean, allows user to supply a custom bias file and overwrite the default file
@@ -91,7 +103,11 @@ verbose
 '''''''''''''''''
 See Stage 3 inputs
 
-isplots
+isplots_S1
+'''''''''''''''''
+Sets how many plots should be saved when running Stage 1. A full description of these outputs is available here: :ref:`Stage 3 Output <s3-out>`
+
+nplots
 '''''''''''''''''
 See Stage 3 inputs
 
@@ -109,7 +125,7 @@ Smoothing length for the trace location
 
 expand_mask
 '''''''''''''''''
-Aperture around the trace to mask
+Aperture (in pixels) around the trace to mask
 
 ignore_low
 '''''''''''''''''
@@ -551,6 +567,11 @@ Start and End of the wavelength range being considered. Set to None to use the s
 allapers
 ''''''''
 If True, run S4 on all of the apertures considered in S3. Otherwise the code will use the only or newest S3 outputs found in the inputdir. To specify a particular S3 save file, ensure that "inputdir" points to the procedurally generated folder containing that save file (e.g. set inputdir to /Data/JWST-Sim/NIRCam/Stage3/S3_2021-11-08_nircam_wfss_ap10_bg10_run1/).
+
+
+mask_columns
+''''''''
+List of pixel columns that should not be used when constructing a light curve.  Absolute (not relative) pixel columns should be used. Figure 3102 is very helpful for identifying bad pixel columns.
 
 
 recordDrift
--- a/src/eureka/S1_detector_processing/bias_sub.py
+++ b/src/eureka/S1_detector_processing/bias_sub.py
@@ -31,7 +31,6 @@ def do_correction(input_model, bias_model, meta, log):
     -------
     output_model: data model object
         bias-subtracted science data
-
     """
     # Check for subarray mode and extract subarray from the
     # bias reference data if necessary
@@ -55,7 +54,7 @@ def do_correction(input_model, bias_model, meta, log):
         for jj in range(ngroup):
             data_median = np.nanmedian(data[jj][np.where(trace_mask)])
             scale_factor[ii, jj] = data_median/bias_median
-    mean_scale_factor = np.mean(scale_factor, axis=0)
+    mean_scale_factor = np.nanmean(scale_factor, axis=0)
     log.writelog(f'  Mean bias scale factors (by group): {mean_scale_factor}')
     
     # Get segment number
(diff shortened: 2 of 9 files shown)


src/eureka/S1_detector_processing/bias_sub.py
```diff
@@ -0,0 +1,187 @@
...
+from ..lib.smooth import smooth
+
+log = logging.getLogger(__name__)
+log.setLevel(logging.DEBUG)
+
+
+def do_correction(input_model, bias_model, meta, log):
+    """
+    Short Summary
+    -------------
```
taylorbell57, line 19: No need to have these lines

src/eureka/S1_detector_processing/bias_sub.py
```diff
@@ -0,0 +1,187 @@
...
+        science data to be corrected
+
+    bias_model: super-bias model object
+        bias data
+
+    Returns
+    -------
+    output_model: data model object
+        bias-subtracted science data
+
```
taylorbell57, line 34: No blank line here

src/eureka/S1_detector_processing/bias_sub.py
```diff
@@ -0,0 +1,187 @@
...
+    # Compute median of each integration/group outside of trace
+    # Using double-for loop to avoid heavy memory usage
+    nint, ngroup, nrow, ncol = input_model.data.shape
+    scale_factor = np.zeros((nint, ngroup))
+    for ii in range(nint):
+        data = input_model.data[ii]
+        for jj in range(ngroup):
+            data_median = np.nanmedian(data[jj][np.where(trace_mask)])
+            scale_factor[ii, jj] = data_median/bias_median
+    mean_scale_factor = np.mean(scale_factor, axis=0)
```
taylorbell57, line 58: Should this be `np.nanmean` to protect against the case where one of the scale_factors ends up being nan?

src/eureka/S1_detector_processing/bias_sub.py
```diff
@@ -0,0 +1,187 @@
...
+        elif (meta.bias_correction == "group_level") or \
+             (meta.bias_correction == "smooth"):
+            # Apply correction for zeroframe
+            bias_model.data *= mean_scale_factor[0]
+            if isinstance(meta.bias_group, int):
+                # Scale superbias frame using values from one group
+                jj = meta.bias_group
+                assert jj > 0, f"Group number should be >0, got: {jj}"
+                assert jj <= ngroup, f"Group number should be <={ngroup}" + \
+                    ", got: {jj}"
```
taylorbell57, line 114: I'm pretty sure this needs to be an fstring as well as I think this would just print "{jj}" instead of the value of jj

src/eureka/S1_detector_processing/bias_sub.py
```diff
@@ -0,0 +1,187 @@
...
+        bias_model.data[np.newaxis, np.newaxis, :, :]
+    
+    try:
+        if meta.bias_correction == "mean":
+            if isinstance(meta.bias_group, int):
+                # Scale superbias frame using single value for all groups
+                jj = meta.bias_group
+                assert jj > 0, f"Group number should be >0, got: {jj}"
+                assert jj <= ngroup, f"Group number should be <={ngroup}" + \
+                    ", got: {jj}"
```
taylorbell57, line 91: I'm pretty sure this needs to be an fstring as well as I think this would just print "{jj}" instead of the value of jj

src/eureka/S1_detector_processing/superbias.py
```diff
@@ -0,0 +1,64 @@
...
+            # Open the superbias ref file data model
+            bias_model = datamodels.SuperBiasModel(self.bias_name)
+
+            # Do the bias subtraction
+            result = bias_sub.do_correction(input_model, bias_model, 
+                                            self.s1_meta, self.s1_log)
+
+            # Close the superbias reference file model and
+            # set the step status to complete
+            bias_model.close()
```
taylorbell57, line 61: The safer way would be to do `with datamodels.SuperBiasModel(self.bias_name) as bias_model` in the same way as you did `input_model`.

src/eureka/S1_detector_processing/bias_sub.py
```diff
@@ -0,0 +1,187 @@
...
+                log.writelog('  Applying mean bias correction using ' +
+                             f'group {jj}.')
+                bias_model.data *= mean_scale_factor[jj-1]
```
taylorbell57, line 94: Here you only modufy `bias_model.data`, while below you sometimes modify only `bias_data` or modify both of them. I just want to clarify that's your intended behaviour since I don't really understand the differences between the two

src/eureka/S1_detector_processing/bias_sub.py
```diff
@@ -0,0 +1,187 @@
...
+                bias_data *= scale_factor[:, :, np.newaxis, np.newaxis]
+            else:
+                raise ValueError('Incorrect meta.bias_group value: ' + 
+                                 f'{meta.bias_group}. Should be ' +
+                                 '[1, 2, ..., each].')
+        else:
+            log.writelog('  No bias correction applied.')
+            bias_data = None
+    except Exception as error:
+        print(repr(error))
```
taylorbell57, line 132: Should this be logged instead of printed? If it is logged, it should be forced to be verbose

src/eureka/S1_detector_processing/bias_sub.py
```diff
@@ -0,0 +1,187 @@
...
+        bias_model = reffile_utils.get_subarray_model(input_model, bias_model)
+
+    # Compute trace mask, where ones are good background regions
+    input_model = mask_trace(input_model, log, meta)
+    trace_mask = input_model.trace_mask
+
+    # Compute median of bias frame outside of trace
+    bias_median = np.nanmedian(bias_model.data[np.where(trace_mask)])
+
+    log.writelog('  Computing scale factor for superbias correction.')
```
taylorbell57, line 48: Do we want to force all of these logs to be verbose, or should we allow the prints to the console to be suppressed if `not meta.verbose`

src/eureka/S1_detector_processing/bias_sub.py
```diff
@@ -0,0 +1,187 @@
...
+                             'each group.')
+                bias_data *= scale_factor[:, :, np.newaxis, np.newaxis]
+            else:
+                raise ValueError('Incorrect meta.bias_group value: ' + 
+                                 f'{meta.bias_group}. Should be ' +
+                                 '[1, 2, ..., each].')
+        else:
+            log.writelog('  No bias correction applied.')
+            bias_data = None
+    except Exception as error:
```
taylorbell57, line 131: Do we really want to catch these exceptions? Does the code just fail more gracefully if the exceptions are caught and None is returned, or are you wanting the code to continue on without the user's requested effects? Because there is so much text logged to the console during Stage 1, I can easily imagine someone missing the error messages printed to the console and assuming everything ran successfully if this causes the code not to crash when an error is hit

src/eureka/S1_detector_processing/bias_sub.py
```diff
@@ -0,0 +1,187 @@
...
+
+    # Subtract the bias data from the science data
+    output_model = subtract_bias(input_model, bias_model, bias_data)
+
+    output_model.meta.cal_step.superbias = 'COMPLETE'
+
+    return output_model
+
+
+def subtract_bias(input, bias, bias_data=None):
```
taylorbell57, line 143: Need to document the bias_data input

src/eureka/S1_detector_processing/bias_sub.py
```diff
@@ -0,0 +1,187 @@
...
+    bias: superbias model object
+        the superbias image data
+
+    Returns
+    -------
+    output: data model object
+        bias-subtracted science data
+
+    """
+
```
taylorbell57, line 164: No blank lines above or below this """

src/eureka/S1_detector_processing/bias_sub.py
taylorbell57: Should also give credit to jwst pipeline here since this is mostly copy-pasted from their function.

taylorbell57 (changes requested):

docs/source/ecf.rst
```diff
@@ -35,6 +35,18 @@ If True, skip the named step.
...
    Note that some instruments and observing modes might skip a step either way! See the `calwebb_detector1 docs <https://jwst-pipeline.readthedocs.io/en/latest/jwst/pipeline/calwebb_detector1.html>`__ for the list of steps run for each instrument/mode by the STScI's JWST pipeline.
 
+bias_correction
+'''''''''''''''''
+Method applied to correct the superbias using a scale factor (SF).  Here, SF = (median of group)/(median of superbias), using a background region that is ``expand_mask`` pixels from the measured trace.  The option ``None`` applies no correction; ``group_level`` computes SF for every integration in ``bias_group``; ``smooth`` applies a smoothing filter of length ``bias_smooth_length`` to the ``group_level`` SF values; and ``mean`` uses the mean SF over all integrations.  We currently recommend using ``smooth`` with a ``bias_smooth_length`` that is ~15 minutes.
+
+bias_group
+'''''''''''''''''
+Integer or string.  Specifies which group number should be used when applying the bias correction.  We currently recommend using the first group (``bias_group`` = 1).  There is no group 0.  Users can also specify ``each``, which computes a unique bias correction for each group.
```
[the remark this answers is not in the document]
kevin218: I only see this being useful for NIRSpec BOTS since it's the only instrument/mode without bias pixels.  I'll not that in the docs.

docs/source/ecf.rst
```diff
@@ -35,6 +35,18 @@ If True, skip the named step.
...
+
+bias_smooth_length
+'''''''''''''''''
+Integer. When ``bias_correction = smooth``, this value is used as the window length during smoothing.
```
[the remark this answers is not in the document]
kevin218: Number of integrations. I added this to the description.

src/eureka/S1_detector_processing/bias_sub.py
kevin218: I agree, but it was there when I copied this file from the `jwst` pipeline.  When I made edits to enable our custom bias correction, I left as much original code as possible.

src/eureka/S1_detector_processing/superbias.py
kevin218: While true, this code is a straight copy/paste from the `jwst` pipeline.

src/eureka/S1_detector_processing/bias_sub.py
kevin218: Nice catch!  I forgot to include `bias_data = None` in this section.  

`bias_model.data` is the calibrated superbias file from commissioning.  It's dimensions (ny, nx) cannot be changed to allow for group- or integration-level variations.

`biad_data` is the 4D equivalent (nint, ngroup, ny, nx) that enables time-dependent corrections.

src/eureka/S1_detector_processing/bias_sub.py
kevin218: Yes!  Unlike a lot of Stage 1 output, this is actually helpful.

src/eureka/S1_detector_processing/bias_sub.py
kevin218: Credit is given at the top of the file

src/eureka/S1_detector_processing/bias_sub.py
kevin218: I'm trying to catch an assortment of possible errors within the `try` statement and provide meaningful error messages.  When returning `None`, the code crashes on setting the step to complete, which works out well.

taylorbell57: 1. So this screenshot is what I get if I run your new code with `bias_group=-1` to try to throw an error:
Screen Shot 2023-03-07 at 14 50 32

2. If we keep the try+except for the purposes of putting the error into the log, but then do `raise error` instead of `return None`:
Screen Shot 2023-03-07 at 15 13 44

3. Meanwhile, if we remove the try+except, then we get:
Screen Shot 2023-03-07 at 15 18 20


People so rarely read further up into their tracebacks, so whatever error ends up on that bottom line is the one they're most likely to see. I think we should go with option 2 or 3, dependeing on whether you want the error in the log file.

src/eureka/S1_detector_processing/bias_sub.py
taylorbell57: I still think we should remove this to keep uniformity between all of our other files.

src/eureka/S1_detector_processing/bias_sub.py
kevin218: Ahh, yes.  I'll use `raise error`

kevin218:
@taylorbell57 The latest commit resolves the last two issues.

taylorbell57 (approved):

PR #509, opened 2023-03-16
Added new centroiding method mgmc and fixed bugs.

### LINKED ISSUE #470: [Bug]: NIRCam photometry centroiding sometimes breaks
### FAQ check

### Instrument

NIRCam (Stages 1-3)

### What happened?

As is shown in the figure below, the NIRCam SW photometry code can at least sometimes completely break. After my intern Isaac (@isaedelman) manually sigma clipped some especially bad values, it appears that most of the time the centroiding does work adequately, but sometimes it seems to just completely blow up. It also appears that negative PSF widths are allowed (the average sx value was -2.47 and the average sy value was -3.03 for reference) which is technically okay since the parameters get squared, but it will also lead to bimodal posteriors for the Gaussian centroiding fit and that could at least partially lead to some of the mess on the figure below. We'll need to tweak the centroiding code to have some priors to make sure that things don't blow up like this.

NIRCamSwCentroids

### Code of Conduct

isaedelman:
Stage3 is having an issue with the method used for centroiding. The method currently used is fgc it is also the cuase of the issue. When running SW photometry fgc struggles with centroiding. So I added a new method mgmc.
This new method is called mgmc and uses gaussian centroiding. This allows for the detection of a centroid location x, y. Along with 1-sig gaussian widths sx, sy.
This enables the code to correclty find the source in the calints files. Allowing for a more precise and fucntioning method for SW photometry. Bugs were also addressed such as masking NANs and plotting figures.

This pull request resolves #470 by implementing a new centroiding method that does not suffer from the same bug and uses other open source software.

flake8 and pytest both pass on my machine, and I have documented all new .ecf parameters in the docs. I have also written doc strings for all new functions.

On my laptop the new mgmc method takes (109.72 min) to run while the old fgc method takes (108.68 min) to run (used 1681 integrations for time trials). But the mgmc method works while the fgc method fails.


isaedelman:
Here is an example of the (fgc method) output with figures 3108 & 3109:
Figure(3108): fig3108-1D_LC

Figure(3109): fig3109-Centroid

isaedelman:
Here is an example of the (mgmc method) output with figures 3108 & 3109:

Figure(3108): fig3108-1D_LC

Figure(3109): fig3109-Centroid

taylorbell57:
So GitHub's automated tests don't seem to like forks... I can confirm flake8 passes right now, and am running pytest locally now

taylorbell57:
Can also confirm that all tests currently pass

src/eureka/lib/gaussian_min.py
```diff
@@ -0,0 +1,210 @@
...
+
+    # Create meshgrid
+    x_shape = np.arange(img.shape[1])
+    y_shape = np.arange(img.shape[0])
+    x_mesh, y_mesh = np.meshgrid(x_shape, y_shape)
+
+    # Fit the gaussian width by minimizing minfunc with Nelder-Mead method. 
+    # The inital guess is [400, 41, 41] 
+    # or [brightness of pixles in area of interest, xsigma, ysigma].
+    results = minimize(minfunc, [400, 41, 41],
```
taylorbell57, line 199: Can you add an if/else statement above that uses these values as the initial guess if `meta.inst == 'nircam'` and instead uses `[400, 5, 5]` if `meta.inst == 'miri'`. To avoid duplicated code, make a variable called `guess` that equals these initial guesses and then pass that variable into this function. Also update the comment above to be more general

src/eureka/lib/gaussian_min.py
```diff
@@ -0,0 +1,210 @@
...
+    
+    # Cropping frame to speed up guassian fit
+    minx = -int(meta.gauss_frame)+int(x)
+    maxx = int(meta.gauss_frame)+int(x)
+
+    # Set Frame size based off of frame crop
+    frame = img[:, minx:maxx]
+
+    # Create meshgrid
+    x_shape = np.arange(img.shape[1])
```
taylorbell57, line 192: Can you apply minx and maxx to this array on this line, and then remove the minx:maxx from the lines below. That'll save the computer from unnecessarily making the meshgrid for pixels we know we don't need (using less CPU and RAM resources)

taylorbell57:
Also, if the `ctr_cutout_size` values don't have the exact same meaning as in the 'fgc' method, please edit the description as needed in the ECF templates and in the ecf.rst file.

taylorbell57:
Finally, you'll also need to update the MIRI photometry ECF template and test ECF

taylorbell57:
All tests and flake8 pass successfully on my laptop - I'll read through the code when I get the time

src/eureka/lib/centerdriver.py
```diff
@@ -52,50 +58,48 @@ def centerdriver(method, data, guess, trim, radius, size,
...
-    # elif method == 'lac':
-    #     y, x = ctr.actr(img, loc, asym_rad=radius,
-    #                     asym_size=size, method='col')
-    # elif method == 'bpf' or method == 'ipf':
-    #     y, x, flux, sky = pf.spitzer_fit(img, msk, weights, psf, psfctr,
-    #                                      expand, method)
-    #     extra = flux, sky
+    # only plot when we do the second fit
+    if (meta.isplots_S3 >= 5 and method[-4:] == '_sec' and i < meta.nplots and
+            len(img) == (2 * meta.ctr_cutout_size + 1)):
```
taylorbell57, line 101: What is the purpose of the clause `len(img) == (2 * meta.ctr_cutout_size + 1))`? What are you trying to do here, and is that not already achieved by the `method[-4:] == '_sec'` clause?

src/eureka/lib/gaussian_min.py
```diff
@@ -0,0 +1,207 @@
...
+        First guess of y centroid position.
+
+    Notes
+    -----
+    History:
+
+    - Feb 22, 2023 Isaac Edelman 
+        Initial implementation.
+    """
+    # Create median frame from Integration
```
taylorbell57, line 112: `# Create median frame from batch`

src/eureka/lib/gaussian_min.py
```diff
@@ -0,0 +1,207 @@
...
+
+    # Create initial centroid guess using specified method
+    if meta.centroid_tech.lower() in ['com', '1dg', '2dg']:
+        cent_func = getattr(sys.modules[__name__], 
+                            ("centroid_" + meta.centroid_tech.lower()))
+        x, y = cent_func(median_Frame)
+    else:
+        print("Invalid centroid_tech option")
+
+    # Returns x,y coordinates of centroid position
```
taylorbell57, line 123: Don't need to add so obvious a comment

src/eureka/lib/gaussian_min.py
```diff
@@ -0,0 +1,207 @@
...
+    y_shape = np.arange(img.shape[0])
+    x_mesh, y_mesh = np.meshgrid(x_shape, y_shape)
+
+    # The inital guess for miri is [400, 41, 41] 
+    # or [brightness of pixles in area of interest, xsigma, ysigma].
+    # This guess is used as a starting point for the algorithm.
+    if (meta.inst == 'nircam'):
+        initial_guess = [400, 5, 5]
+    elif (meta.inst == 'miri'):
+        initial_guess = [400, 41, 41]
```
taylorbell57, line 191: You have these backwards - MIRI's PSF width is ~5 pixels and NIRCam's is ~41.  Also only add parentheses to if statements if you need multiple lines of clauses. And adjust the comment to be more generic along the lines of `# The initial guess for [Gaussian amplitude, xsigma, ysigma]`

src/eureka/lib/gaussian_min.py
```diff
@@ -0,0 +1,207 @@
...
+    else:
+        print("Possible mgmc gaussian centroiding intial guess issue. " +
+              "Default values used for guess in gaussian_min minimize " + 
+              "function")
+        initial_guess = [400, 20, 20]
```
taylorbell57, line 196: I don't think you really need this, but if you want to have this then you should be a bit more detailed than "possible initial guess issue". If you keep this warning, instead put something that'll help people troubleshoot the issue without having to dig into the code like `f"Warning: Photometry has only been tested on MIRI and NIRCam while meta.inst is set to {meta.inst}"`

taylorbell57 (changes requested):

docs/source/ecf.rst
```diff
@@ -480,13 +480,17 @@ interp_method
...
+Only used for photometry analyses. For the 'fgc' method this parameter is the amount of pixels all around the guessed centroid location which should be used for the more precise second centroid determination after the coarse centroid calculation. E.g., if ctr_cutout_size = 10 and the centroid (as determined after coarse step) is at (200, 200) then the cutout will have its corners at (190,190), (210,210), (190,210) and (210,190). The cutout therefore has the dimensions 21 x 21 with the centroid pixel (determined in the coarse centroiding step) in the middle of the cutout image. For the 'mgmc' method it is the same however the value must be odd E.g. ctr_cutout_size = 11.
```
taylorbell57, line 493: Are you sure that it is the same - if so, there'd be absolutely no reason for it to have to be odd... If it really must be odd, then the MIRI ECF template will need to be updated (currently set to 10)

src/eureka/S3_data_reduction/plots_s3.py
```diff
@@ -1160,3 +1175,155 @@ def stddev_profile(meta, n, m, stdevs, p7thresh):
...
+    data : Xarray Dataset
+        The Dataset object.
+    log : logedit.Logedit
+        The current log.
+    m : int
+        The file number.
+    position : ndarray
+        The y, x position of the star.
+    median_tilt_frame : ndarray
+        The median of the first 10 integrations.
```
taylorbell57, line 1200: Since you've renamed this input, you need to rename it in the docs here

src/eureka/S3_data_reduction/plots_s3.py
```diff
@@ -1289,33 +1291,39 @@ def tilt_events(meta, data, m, position, median_tilt_frame):
...
+
+        # Create list of all .png tilt images in tilt_event folder
+        all_images = []
+        for file in os.listdir(meta.outputdir + 
+                               f'figs{os.sep}tilt_events{os.sep}'):
+            if file.endswith(".png"):
+                in_filenames = os.path.join(meta.outputdir + 
+                                            f'figs{os.sep}' + 
+                                            f'tilt_events{os.sep}', file)
+        in_filenames.sort()
```
taylorbell57, line 1315: `in_filenames` gets over-written every time it finds a new png which is definitely not your intended behaviour. You're going to need to append the result of `os.path.join` to some list

src/eureka/S3_data_reduction/plots_s3.py
```diff
@@ -1289,33 +1291,39 @@ def tilt_events(meta, data, m, position, median_tilt_frame):
...
+
+        # Create list of all figure names to pull from later to create .gif
+        for i in range(len(in_filenames)):
+            all_images.append(imageio.v2.imread(meta.outputdir + 
+                                                f'figs{os.sep}' + 
+                                                f'tilt_events{os.sep}' + 
+                                                in_filenames[i]))
```
taylorbell57, line 1322: A much cleaner way of doing this is to do:

```
for fname in in_filenames:
    all_images.append(imageio.v2.imread(
        meta.outputdir + f'figs{os.sep}tilt_events{os.sep}{fname}'))
```

src/eureka/lib/centerdriver.py
```diff
@@ -97,8 +97,8 @@ def centerdriver(method, data, guess, trim, radius, size, i, m, meta,
         extra = sy, sx  # Gaussian 1-sigma half-widths
 
     # only plot when we do the second fit
-    if (meta.isplots_S3 >= 5 and method[-4:] == '_sec' and i < meta.nplots and
-            len(img) == (2 * meta.ctr_cutout_size + 1)):
+    if (meta.isplots_S3 >= 5 and method[-4:] == '_sec' and
+            i < meta.nplots):
```
taylorbell57, line 101: No need to put that `i < meta.nplots):` on a new line - you can just move that back up to the line above (I think I may have accidentally told you to do that during our meeting today)

taylorbell57 (changes requested):
Just a few minor comments and then this is good to merge!

docs/source/ecf.rst
isaedelman: Yes, I think it does need to be odd. ctr_cutout_size is used as the variable for box_size in th eentoid_sources function and in the documentation it states that "box_size must have odd values for both axes." I also tested it out in a jupyter notebook and it errors out if an even number is used stating: "ValueError: box_size must have an odd value for both axes". I updated the MIRI ECF to address it.

taylorbell57: If that's the case, then I really don't think this variable has the same meaning as in the fgc method

taylorbell57: The documentation below makes it seem like `box_size` is the *full-width* of the cutout box, while fgc treats `2*ctr_cutout_size+1` as the full-width.

Screen Shot 2023-04-13 at 09 55 08

isaedelman: How does this sound? "For the 'mgmc' method ctr_cutout_size is the full length of the cutout image along each axis and has to be an odd value. E.g. ctr_cutout_size = 11 then the cutout image will have the dimensions of 11 x 11 with the centroid pixle (determined in the first centroid guess) in the middle of the cutout image."

taylorbell57: Maybe to keep things simple we can keep the original meaning of `ctr_cutout_size` and do the conversion inside the code? Aka, pass `box_size=(2*ctr_cutout_size+1)` into centroid_sources? And then adjust the NIRCam template to use a value of 5 instead and leave MIRI at a value of 10

isaedelman: Sure, I'll update the code.

isaedelman: How is this? "Only used for photometry analyses. For the 'fgc' method this parameter is the amount of pixels all around the guessed centroid location which should be used for the more precise second centroid determination after the coarse centroid calculation. E.g., if ctr_cutout_size = 10 and the centroid (as determined after coarse step) is at (200, 200) then the cutout will have its corners at (190,190), (210,210), (190,210) and (210,190). The cutout therefore has the dimensions 21 x 21 with the centroid pixel (determined in the coarse centroiding step) in the middle of the cutout image. For the 'mgmc' method ctr_cutout_size is the full length of the cutout image along each axis. E.g. , if ctr_cutout_size = 10 then the cutout image will be (2*ctr_cutout_size+1) along the x-axis and (2*ctr_cutout_size+1) along the y-axis, giving the cutout image the dimensions of 21 x 21 with the first centroid guess pixle in the center."  

I also noticed that in the 'fgc' defintion it gives the dimensions of "(190,190), (210,210), (190,210) and (210,190). The cutout therefore has the dimensions 21 x 21" but 210-190 = 20 would that not give a dimension of 20 x 20?

taylorbell57 (changes requested):

taylorbell57 (approved):

taylorbell57:
Gah, it doesn't want to merge in automatically - I'll try fixing any conflicts in 1-2 hours

taylorbell57:
Huh, so it did successfully merge the code into main but somehow failed to acknowledge the fact that it did so... I'll just manually close this PR, but fear not, the code merged successfully and gave you credit where it was due

PR #511, opened 2023-03-28
Bug patches after latest PR

taylorbell57:
The latest PR unfortunately introduced some glaring bugs which I didn't catch myself nor did the automated tests.

1. Because of the way I implemented the new `meta.nints` code, unshared fits would break after fitting for the first lightcurve.
2. SinusoidalPhaseCurve fits would break with the PyMC3 version of the code as `chan` was always used but not always defined.
3. Also, at present, setting waverange_start and waverange_end to None causes a crash for NIRSpec spectra which shouldn't happen. Users should be able to use the default extraction if they want (or need).

Aside from resolving these bugs, I added a bit nicer formatting for the log statements inside straighten.py.

### DIFF
--- a/src/eureka/S2_calibrations/s2_calibrate.py
+++ b/src/eureka/S2_calibrations/s2_calibrate.py
@@ -140,13 +140,15 @@ def calibrateJWST(eventlabel, ecf_path=None, s1_meta=None, input_meta=None):
             # (or someone is putting weird files into Eureka!)
             pipeline = EurekaSpec2Pipeline()
 
-            # By default pipeline can trim the dispersion axis,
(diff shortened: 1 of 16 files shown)


taylorbell57:
I've just re-run my Sagan Summer Workshop notebook for the MIRI data, and it seems there's something wrong with the ypos S5 model as well as the ypos parameter is completely unconstrained with the new code. I'll look into this as soon as I can

taylorbell57:
Alright, well that's all of the known bugs I introduced with the latest PR taken care of!...

kevin218 (approved):

PR #512, opened 2023-03-31
Enable Binned LC Plots

kevin218:
Working with datasets that have many integrations makes it difficult to see what's going on in the light curve plots.  So, when `meta.nbin_plot` is given, the light curves in all figures are now binned to the desired time resolution.  Before, only some plots were binned.

I also updated the `jwst` version to 1.8.2.  Version 1.9 appears to still be buggy.

### DIFF
--- a/src/eureka/S5_lightcurve_fitting/lightcurve.py
+++ b/src/eureka/S5_lightcurve_fitting/lightcurve.py
@@ -6,7 +6,7 @@ from . import models as m
 from . import fitters
 from . import gradient_fitters
 from .utils import COLORS, color_gen
-from ..lib import plots
+from ..lib import plots, util
 
 
 class LightCurve(m.Model):
@@ -180,12 +180,22 @@ class LightCurve(m.Model):
                 flux = flux[channel*len(self.time):(channel+1)*len(self.time)]
                 unc = unc[channel*len(self.time):(channel+1)*len(self.time)]
 
+            # Get binned data and times
+            if not hasattr(meta, 'nbin_plot') or meta.nbin_plot is None or \
+               meta.nbin_plot > len(self.time):
+                nbin_plot = len(self.time)
+            else:
+                nbin_plot = meta.nbin_plot
+            binned_time = util.binData(self.time, nbin_plot)
+            binned_flux = util.binData(flux, nbin_plot)
+            binned_unc = util.binData(unc, nbin_plot, err=True)
+
             fig = plt.figure(5103, figsize=(8, 6))
             fig.clf()
             # Draw the data
             ax = fig.gca()
-            ax.errorbar(self.time, flux, unc, fmt='.', color=self.colors[i],
-                        zorder=0)
+            ax.errorbar(binned_time, binned_flux, binned_unc, fmt='.',
+                        color=self.colors[i], zorder=0)
 
             # Make a new color generator for the models
             plot_COLORS = color_gen("Greys", 6)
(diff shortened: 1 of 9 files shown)


src/eureka/S5_lightcurve_fitting/plots_s5.py
```diff
@@ -87,27 +87,39 @@ def plot_fit(lc, model, meta, fitter, isTitle=True):
 
         residuals = flux - model
 
+        # Get binned data and times
+        if not hasattr(meta, 'nbin_plot') or meta.nbin_plot is None or \
+           meta.nbin_plot > len(lc.time):
+            nbin_plot = len(lc.time)
```
taylorbell57, line 93: These should use `time` and not `lc.time` (because `lc.time` can be longer than `time` for multi-white fits).

src/eureka/S5_lightcurve_fitting/plots_s5.py
```diff
@@ -87,27 +87,39 @@ def plot_fit(lc, model, meta, fitter, isTitle=True):
...
+        else:
+            nbin_plot = meta.nbin_plot
+        binned_time = util.binData(lc.time, nbin_plot)
```
taylorbell57, line 96: Again, `time` and not `lc.time` here.

src/eureka/S5_lightcurve_fitting/lightcurve.py
```diff
@@ -200,12 +200,22 @@ def plot(self, meta, fits=True):
                 time, flux, unc = split([time, flux, unc],
                                         meta.nints, channel)
 
+            # Get binned data and times
+            if not hasattr(meta, 'nbin_plot') or meta.nbin_plot is None or \
+               meta.nbin_plot > len(self.time):
+                nbin_plot = len(self.time)
```
taylorbell57, line 206: Again, `time` and not `self.time`

src/eureka/S5_lightcurve_fitting/lightcurve.py
```diff
@@ -200,12 +200,22 @@ def plot(self, meta, fits=True):
...
+            else:
+                nbin_plot = meta.nbin_plot
+            binned_time = util.binData(self.time, nbin_plot)
```
taylorbell57, line 209: Again, `time` and not `self.time`

taylorbell57 (changes requested):
In addition to the specific feedback below, do we exclusively want to plot the binned data (like I did for Fig5104) or do we want to put the unbinned data in the background using faint, translucent points (like I did for Fig5304) - or do we want to show both using different plots, as I did. I have examples of those two figures below

Fig 5104 example:
fig5104_ch0_phaseVariations_exoplanet

Fig 5304 example:
fig5304_ch0_phaseVariations_exoplanet

kevin218:
I'm going on vacation until April 12th, so you're welcome to make changes to this PR while I'm gone.

I support having Figures 5104 and 5304 as separate plots.

taylorbell57:
Can you clarify what you mean by "I support having Figures 5104 and 5304 as separate plots."? To clarify my question, I had meant do we want to:

1. have fig5104 (just binned data) and fig5304 (binned data with unbinned data in the background) equivalents for all the figures you've changed, or
2. just have the fig5014 (just binned data) equivalents that you've made

kevin218:
Ok, now I see what you're asking.

For the 3 panel plot (raw, normalized, residuals), I would prefer using only the binned data so that the default ylim range gives meaningful information.

I'm ok with plotting both for the single-panel light curve.

taylorbell57:
Okay, I'll try to make both for the single-panel lightcurve if I find the time before you get back.

taylorbell57 (approved):

PR #518, opened 2023-04-13
Updating docs to v0.9

taylorbell57:
Simply incrementing docs to point to the new v0.9 (I will make v0.9 once this PR is merged).

### DIFF
--- a/.github/workflows/draft-pdf.yml
+++ b/.github/workflows/draft-pdf.yml
@@ -7,7 +7,7 @@ jobs:
     name: Paper Draft
     steps:
       - name: Checkout
-        uses: actions/checkout@v2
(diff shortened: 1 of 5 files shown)


taylorbell57:
So Codecov just hit v5.0 which means that we need to slightly change the way in which we use it for the automated testing on GitHib ([read more](https://docs.codecov.com/docs/deprecated-uploader-migration-guide#python-uploader)). It was just a small, simple tweak though and we should be good to go now

taylorbell57:
Apparently my intern Isaac just found a bug with the PR of his that I just merged earlier today. We'll try to rapidly patch that tomorrow before this PR gets approved

kevin218 (approved):

PR #520, opened 2023-04-14
Fixing S3 plotting bugs

taylorbell57:
Isaac and I noticed a couple of plotting bugs (one from Isaac's PR, and two from some long ago PR) that were super easy to fix quickly. Let's fix them with this PR before v0.9


taylorbell57:
@kevin218 the tests fail just because the PR is from Isaac's fork, so the tests can't be run automatically. The PR is super simple and fixes bugs where plots are cleared after plt.subplots is called (which then ruins the plots) as well as a bug related to filenames which causes NIRCam photometry runs to crash if isplots_S5>=5. I'm happy with the PR and would approve it, but can't since I was the one to make it

PR #524, opened 2023-04-20
Skip masking trace, etc when bias_correction is None; enable row-by-row BG subtraction; edit NIRCam extraction height

### LINKED ISSUE #522: [Enhancement]: mask_trace used by bias_sub.do_correction without clear documentation
### Instrument

NIRSpec (Stages 1-3)

### What is your suggestion?

I'm working on reproducing the ERS' PRISM data reduction, and I noticed that the S1 log wrote "Masking Curved Trace." despite the fact that I set `masktrace=False` in my S1 ECF. Looking more closely, I see that the trace gets masked during the bias correction step and uses the settings related to `masktrace` regardless of the value of `masktrace`. Generally if we set a boolean variable like that to False, one can just safely ignore all of the related parameters, and I'm sure I'm not the only one who would either not notice this behaviour (and potentially get poor results) or be confused by it (since I specifically told the code not to mask curved traces).

At a minimum this behaviour needs to be well documented, but ideally we can come up with something less confusing.

### Code of Conduct

kevin218:
This resolves #522 by skipping a bunch of unnecessary steps in bias_sub.py when `meta.bias_correction is None`.

### DIFF
--- a/src/eureka/S1_detector_processing/group_level.py
+++ b/src/eureka/S1_detector_processing/group_level.py
@@ -142,12 +142,12 @@ def mask_trace(input_model, log, meta):
         for nc in range(meta.bg_x1, meta.bg_x2):
             mask_low = np.nanmax([0, smooth_coms[nc]-meta.expand_mask])
             mask_hih = np.nanmin([smooth_coms[nc]+meta.expand_mask, nrow-1])
+            trace_mask[mask_low:mask_hih, nc] = np.nan
     else:
         for nc in range(ncol):
             mask_low = np.nanmax([0, smooth_coms[nc]-meta.expand_mask])
             mask_hih = np.nanmin([smooth_coms[nc]+meta.expand_mask, nrow-1])
-
-        trace_mask[mask_low:mask_hih, nc] = np.nan
+            trace_mask[mask_low:mask_hih, nc] = np.nan
 
     input_model.trace_mask = trace_mask
 
--- a/src/eureka/S1_detector_processing/group_level.py
+++ b/src/eureka/S1_detector_processing/group_level.py
@@ -71,24 +71,22 @@ def GLBS(input_model, log, meta):
         data = xrio.makeDataset(dictionary=xrdict)
         data['flux'].attrs['flux_units'] = 'n/a'
         data.attrs['intstart'] = meta.intstart
+        meta.bg_dir = 'CxC'
         if ngrp == all_data.shape[1]-1:
-            data = bkg.BGsubtraction(data, meta,
-                                     log, 0,
+            data = bkg.BGsubtraction(data, meta, log, meta.m,
                                      meta.isplots_S1)
         else:
             # do not show plots except for the last group
-            data = bkg.BGsubtraction(data, meta,
-                                     log, 0, 0)
+            data = bkg.BGsubtraction(data, meta, log, meta.m, 0)
         # Perform BG subtraction along dispersion direction
         # (only useful for NIRCam data)
         if hasattr(meta, 'bg_disp') and meta.bg_disp:
-            log.writelog('  Performing BG subtraction along ' +
-                         'dispersion direction...',
-                         mute=(not meta.verbose))
+            meta.bg_dir = 'RxR'
             if ngrp == all_data.shape[1]-1:
-                data = bkg.BGsubtraction(data, meta, log, 0, meta.isplots_S1)
+                data = bkg.BGsubtraction(data, meta, log, meta.m,
+                                         meta.isplots_S1)
             else:
-                data = bkg.BGsubtraction(data, meta, log, 0, 0)
+                data = bkg.BGsubtraction(data, meta, log, meta.m, 0)
         # Overwrite values in all_data
         all_data[:, ngrp, :, :] = data['flux'].values
     input_model.data = all_data
--- a/src/eureka/S1_detector_processing/group_level.py
+++ b/src/eureka/S1_detector_processing/group_level.py
@@ -1,7 +1,7 @@
 #! /usr/bin/env python
 
 # File with functions to perform GLBS
-# trace masking, 
+# trace masking,
 # custom ref pixel correction for PRISM
 # EMMay, Nov 2022
 import numpy as np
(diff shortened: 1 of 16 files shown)


src/eureka/S3_data_reduction/s3_reduce.py
```diff
@@ -238,7 +238,7 @@ def reduce(eventlabel, ecf_path=None, s2_meta=None, input_meta=None):
 
             datasets = []
             saved_refrence_tilt_frame = None
-            
+
```
taylorbell57, line 241: Can you tell me which tool you're using to be able to see these formatting issues? I'm totally fine with these formatting changes, it's just that flake8 doesn't show them to me so I have no easy way of checking for them when I make or review PRs

taylorbell57 (changes requested):
I don't think this really resolves #522, although the PR is still a useful fix. My point with #522 is that there is no documentation anywhere that would make me expect that the mask trace settings will be used by the bias correction code if I've set `masktrace=False`. For my analyses, I'd figured there's no need to mask a curved trace for the fairly flat PRISM spectrum while doing background subtraction, so I set `masktrace=False` and just ignored the rest of the related settings. I was then surprised that the masktrace related code was still being used (because I'd asked S1 to do the bias correction step), and I'd guess I'm not the only one that'd be surprised by that. The danger is that I hadn't put any thought into the masktrace settings (assuming I could safely ignore them), so we should document somewhere in the ECF and docs pages where and how the masktrace settings get used

kevin218:
Got it!  I had fixed a related problem, but not the exact one you were describing.

src/eureka/S3_data_reduction/s3_reduce.py
kevin218: I think these changes are being made by Atom automatically.  I don't see them and have no control over them.

src/eureka/S1_detector_processing/group_level.py
```diff
@@ -125,19 +134,24 @@ def mask_trace(input_model, log, meta):
...
+    if meta.bg_x1 is not None and meta.bg_x2 is not None:
+        for nc in range(meta.bg_x1, meta.bg_x2):
+            mask_low = np.nanmax([0, smooth_coms[nc]-meta.expand_mask])
+            mask_hih = np.nanmin([smooth_coms[nc]+meta.expand_mask, nrow-1])
+            trace_mask[mask_low:mask_hih, nc] = np.nan
+    else:
+        for nc in range(ncol):
+            mask_low = np.nanmax([0, smooth_coms[nc]-meta.expand_mask])
+            mask_hih = np.nanmin([smooth_coms[nc]+meta.expand_mask, nrow-1])
+            trace_mask[mask_low:mask_hih, nc] = np.nan
```
taylorbell57, line 148: Rather than duplicating code, let's make a new variable `cols` or something, set it equal to `range(meta.bg_x1, meta.bg_x2)` in the `if` statement, and set it to `range(ncol)` in the `else` statement. Then do `for nc in cols:` outside of the if/else statements

src/eureka/S2_calibrations/s2_calibrate.py
```diff
@@ -232,15 +232,21 @@ def run_eurekaS2(self, filename, meta, log):
...
-            # Controls the cross-dispersion extraction
+        if hasattr(meta, 'slit_y_high') and meta.slit_y_high is not None:
+            #  NIRSpec subarray upper bound in cross-dispersion direction
             self.assign_wcs.slit_y_high = meta.slit_y_high
 
+        if hasattr(meta, 'tsgrism_extract_height') and \
+           meta.tsgrism_extract_height is not None:
+            # NIRCam grism subarray height in cross-dispersion direction
+            self.extract_2d.tsgrism_extract_height = \
+                meta.tsgrism_extract_height
```
taylorbell57, line 248: Thanks for adding this feature - recent GTO data analyses have seemingly shown the extra pixels to be important

src/eureka/S3_data_reduction/s3_reduce.py
```diff
@@ -293,6 +291,19 @@ def reduce(eventlabel, ecf_path=None, s2_meta=None, input_meta=None):
...
+                if hasattr(meta, 'bg_disp') and meta.bg_disp:
+                    meta.bg_dir = 'RxR'
+                    # Create bad pixel mask (1 = good, 0 = bad)
+                    data['mask'] = (['time', 'y', 'x'],
+                                    np.ones(data.flux.shape, dtype=bool))
+                    data = bg.BGsubtraction(data, meta, log,
+                                            m, meta.isplots_S3)
+                    meta.bg_disp = False
+                    meta.bg_deg = None
+                meta.bg_dir = 'CxC'
```
taylorbell57, line 305: Why set this to CxC afterwards? Don't you want this to be in an `else` statement?

tests/NIRCam_ecfs/S3_NIRCam.ecf
```diff
@@ -20,9 +20,12 @@ ff_outlier      True        # Set False to use only background region (recommend
...
 # Background parameters
-bg_hw       12          # Half-width of exclusion region for BG subtraction (relative to source position)
-bg_deg      1           # Polynomial order for column-by-column background subtraction, -1 for median of entire frame
-p3thresh    5           # X-sigma threshold for outlier rejection during background subtraction
+bg_hw           12          # Half-width of exclusion region for BG subtraction (relative to source position)
+bg_deg          0           # Polynomial order for column-by-column background subtraction, -1 for median of entire frame
+p3thresh        5           # X-sigma threshold for outlier rejection during background subtraction
+bg_disp         False       # Row-by-row BG subtraction (only useful for NIRCam)
+bg_x1           None        # Left edge of exclusion region for row-by-row BG subtraction
+bg_x2           None        # Right edge of exclusion region for row-by-row BG subtraction
```
taylorbell57, line 28: Why not set bg_disp to True in the testing ECF to get some more testing coverage?

taylorbell57 (changes requested):

src/eureka/S3_data_reduction/s3_reduce.py
```diff
@@ -204,14 +204,12 @@ def reduce(eventlabel, ecf_path=None, s2_meta=None, input_meta=None):
...
                 from . import nirspec as inst
-                log.writelog('WARNING: Are you using real JWST data? If so, '
-                             'you should edit the flag_bg() function in '
-                             'nirspec.py and look at Issue #193 on Github!')
             elif meta.inst == 'niriss':
                 raise ValueError('NIRISS observations are currently '
                                  'unsupported!')
             elif meta.inst == 'wfc3':
                 from . import wfc3 as inst
+                meta.bg_dir = 'CxC'
```
taylorbell57, line 212: Why only set this default here?

src/eureka/S3_data_reduction/s3_reduce.py
taylorbell57: I think this at least needs some comment to help us remember what's going on here 6 months from now

src/eureka/S3_data_reduction/s3_reduce.py
kevin218: The WFC3 test fails if I don't set it here.  The remaining instruments can all get set with a single command on Line 309.

kevin218:
@taylorbell57 Do you have any other concerns about this PR before we merge?

taylorbell57 (approved):
Ah sorry, for some reason I don't think I was notified of your changes. Good to go!

PR #532, opened 2023-05-04
Corrected issues with mgmc centroiding, cleaned up formatting, and corrected plotting features.

isaedelman:
Corrected plotting features, addressed edge cases with the mgmc centroiding method, and cleaned up formatting.

----
-Added dashed zero lines to fig.3109 to help identify if any slopes occur.

Before:
---
fig3109-Centroid

after:
---
fig3109-Centroid

-Reformatted fig.3503 to display and corrected 1D gaussians to source in x and y directions.

Before:
---
fig3503_file0_int001_Centroid_Fit

After:
---
fig3503_file0_int000_Centroid_Fit

-Corrected labeling in fig.3507 to display batch values instead of segment values.

-Tweaked s3_reduce.py to remember the first median frame for centroiding method to address issues where a new batch could have a bad start frame for the primary centroid guess causing the position to drift over time.

-Renamed the initial centroiding guess position variable for the mgmc method to address issues with position variable being over written by fgc position.

-Changed font size for fig.5101 to make the side labels not smooshed.

Before:
---
fig5101_ch0_lc_nuts

After:
---
fig5101_ch0_lc_nuts

-Updated the mgmc centroiding method to account for the saved median frame used for centroiding position.

----

This passes test_Photometry_NIRCam.py on my laptop and is flake8 compliant.

### DIFF
--- a/src/eureka/S3_data_reduction/plots_s3.py
+++ b/src/eureka/S3_data_reduction/plots_s3.py
@@ -749,6 +749,8 @@ def phot_centroid(data, meta):
         Enchanced graph layout, 
         added sig display values for sy,sx,
         and fixed issue with ax[2] displaying sy instead of sx.
+    - 2023-04-21 Isaac Edelman
+        Added flat "0" lines to plots.
     """
     plt.figure(3109)
     plt.clf()
@@ -763,24 +765,30 @@ def phot_centroid(data, meta):
     csx_rms = np.sqrt(np.nanmean((csx - np.nanmedian(csx)) ** 2))
     csy = data.centroid_sy.values
     csy_rms = np.sqrt(np.nanmean((csy - np.nanmedian(csy)) ** 2))
+    flat_line = range(data.time)
+    flat_line_zeros = np.zeros(data.time)
 
     ax[0].plot(data.time, data.centroid_x-np.nanmean(data.centroid_x),
                label=r'$\sigma$x = {0:.4f} pxls'.format(cx_rms))
+    ax[0].plot(flat_line, flat_line_zeros, linestyle=':', color='r')
     ax[0].set_ylabel('Delta x')
     ax[0].legend(bbox_to_anchor=(1.03, 0.5), loc=6)
 
     ax[1].plot(data.time, data.centroid_y-np.nanmean(data.centroid_y),
                label=r'$\sigma$y = {0:.4f} pxls'.format(cy_rms))
+    ax[1].plot(flat_line, flat_line_zeros, linestyle=':', color='r')
     ax[1].set_ylabel('Delta y')
     ax[1].legend(bbox_to_anchor=(1.03, 0.5), loc=6)
 
     ax[2].plot(data.time, data.centroid_sx-np.nanmean(data.centroid_sx),
                label=r'$\sigma$sx = {0:.4f} pxls'.format(csx_rms))
+    ax[2].plot(flat_line, flat_line_zeros, linestyle=':', color='r')
     ax[2].set_ylabel('Delta sx')
     ax[2].legend(bbox_to_anchor=(1.03, 0.5), loc=6)
 
     ax[3].plot(data.time, data.centroid_sy-np.nanmean(data.centroid_sy),
                label=r'$\sigma$sy = {0:.4f} pxls'.format(csy_rms))
+    ax[3].plot(flat_line, flat_line_zeros, linestyle=':', color='r')
     ax[3].set_ylabel('Delta sy')
     ax[3].set_xlabel('Time')
     ax[3].legend(bbox_to_anchor=(1.03, 0.5), loc=6)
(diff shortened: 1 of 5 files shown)


src/eureka/S3_data_reduction/plots_s3.py
```diff
@@ -763,24 +765,29 @@ def phot_centroid(data, meta):
     csx_rms = np.sqrt(np.nanmean((csx - np.nanmedian(csx)) ** 2))
     csy = data.centroid_sy.values
     csy_rms = np.sqrt(np.nanmean((csy - np.nanmedian(csy)) ** 2))
+    flat_line = np.zeros(len(data.centroid_x))
 
     ax[0].plot(data.time, data.centroid_x-np.nanmean(data.centroid_x),
                label=r'$\sigma$x = {0:.4f} pxls'.format(cx_rms))
+    ax[0].plot(data.time, flat_line, linestyle=':', color='r')
```
taylorbell57, line 772: For these flat lines, can you instead use `ax[0].axhline` instead so that it spans the full x-axis

src/eureka/S3_data_reduction/plots_s3.py
```diff
@@ -861,34 +868,55 @@ def phot_centroid_fgc(img, x, y, sx, sy, i, m, meta):
...
+    ax[1, 1].plot(norm_distr_y_scaled/norm_y_factor, y_plot,
+                  linestyle='dashed')
+    ax[1, 1].set_ylabel('y position')
+    ax[1, 1].set_xlabel('Normalized Flux')
 
+    # Last plot in (0,1) not used
+    ax[0, 1].set_axis_off()
+
+    fig.tight_layout()
+    plt.tight_layout()
```
taylorbell57, line 915: You shouldn't use both of these - just `plt.tight_layout()` should do

src/eureka/S5_lightcurve_fitting/plots_s5.py
```diff
@@ -109,19 +109,19 @@ def plot_fit(lc, model, meta, fitter, isTitle=True):
         if isTitle:
             ax[0].set_title(f'{meta.eventlabel} - Channel {channel} - '
                             f'{fitter}')
-        ax[0].set_ylabel('Normalized Flux', size=14)
+        ax[0].set_ylabel('Normalized Flux', size=10)
```
taylorbell57, line 112: These are too small and will be much too small when using LaTeX fonts. I'll have to think a bit as to how to better fix these figures, but for now please reset these back to 14

src/eureka/lib/centerdriver.py
```diff
@@ -5,7 +5,8 @@
 from ..S3_data_reduction import plots_s3
 
 
-def centerdriver(method, data, guess, trim, radius, size, i, m, meta,
+def centerdriver(method, data, guess, trim, radius, size, i, m, meta, 
+                 saved_ref_median_frame,
```
taylorbell57, line 9: You need to document this new input

src/eureka/lib/centerdriver.py
```diff
@@ -90,15 +91,17 @@ def centerdriver(method, data, guess, trim, radius, size, i, m, meta,
...
         extra = sy, sx  # Gaussian 1-sigma half-widths
+        refrence_median_frame = None
 
     # only plot when we do the second fit
     if (meta.isplots_S3 >= 5 and method[-4:] == '_sec' and i < meta.nplots):
         plots_s3.phot_centroid_fgc(img, x, y, sx, sy, i, m, meta)
 
     # Make trimming correction and return
-    return ((y, x) + cen - trim), extra
+    return ((y, x) + cen - trim), extra, refrence_median_frame
```
taylorbell57, line 107: You also need to document this new output in the docstring above

src/eureka/lib/gaussian_min.py
```diff
@@ -110,17 +110,22 @@ def pri_cent(img, meta):
         Initial implementation.
     """
     # Create median frame from batch
-    median_Frame = np.ma.median(img, axis=0)
+    # median_Frame = np.ma.median(img, axis=0)
```
taylorbell57, line 113: Delete these old lines

taylorbell57 (changes requested):

src/eureka/S5_lightcurve_fitting/plots_s5.py
isaedelman: This stack overflow might help (if we choose to address this)? https://stackoverflow.com/questions/33942210/consistent-fonts-between-matplotlib-and-latex

taylorbell57: So I use LaTeX fonts most of the time, but my understanding is that they only work if you've installed LaTeX on your computer which is a lengthy and annoying process. It's possible matplotlib is able to do LaTeX fonts without LaTeX installed, but not from my experience. Additionally, using LaTeX fonts substantially slows figure creation, so when I fit NIR data with hundreds of channels I turn off LaTeX fonts to save like 50% of the runtime. Either way, I think this is something that'll take a moderate amount of fiddling on my part to get just right (it's definitely on my to-do list though)

src/eureka/lib/centerdriver.py
```diff
@@ -10,33 +10,40 @@ def centerdriver(method, data, guess, trim, radius, size, i, m, meta,
...
+    saved_ref_median_frame : ndarray
+                             Stored median frame of the first batch.
 
     Returns
     -------
     A y,x tuple (scalars) with the coordinates of center of the target
-    in data.
+    in data. refrence_median_frame an ndarray which is the median
+    of the first batch to be used as the refrence frame for the first
+    centroid location guess in the mgmc method.
```
taylorbell57, line 46: When you have multiple returns like this, it's clearer to break things down in a similar way as the Parameters section is broken down. For this Returns, it would look like:
```
    Returns
    -------
    tuple
        A (y,x) tuple (scalars) with the coordinates of center of the target
        in data.
    ndarray
        The median of the first batch to be used as the refrence
        frame for the first centroid location guess in the mgmc method.
```

src/eureka/lib/gaussian_min.py
```diff
@@ -13,16 +13,16 @@ def evalgauss(params, x, y, x0, y0):
     Parameters
     ----------
     params : list
-        List components : amplitude, x_stddev, y_stddev. 
-        Acts as the inital guess from meta (params_guess).
+             List components : amplitude, x_stddev, y_stddev.
```
taylorbell57, line 16: Similar to what I said about centerdriver.py, all these description lines are also over-indented now

src/eureka/S3_data_reduction/plots_s3.py
```diff
@@ -1232,7 +1256,7 @@ def tilt_events(meta, data, log, m, position, saved_refrence_tilt_frame):
     asb_xpos_max = data.x.values[maxx]
     asb_ypos_min = data.y.values[miny]
     asb_ypos_max = data.y.values[maxy]
-
+
```
taylorbell57, line 1259: Don't have any spaces on blank lines

taylorbell57 (changes requested):

taylorbell57 (approved):

PR #533, opened 2023-05-08
(1) Allows negative Rp/Rs for shallow transits, (2) nansums in optspec extraction to address masking issues, (3) allows for skipping/fitting only certain bins in S5, and (4) updates aperture to be centered properly

erinmmay:
Lots of tiny changes that I'm finally looking to merge into the main branch.

**1. Allows negative Rp/Rs for shallow transits**
For shallow transits, we shouldn't reject steps that try negative Rp/Rs. This results in chopped-off posteriors. I simply removed that condition from the "physicality" check.

**2. nansums in optspec extraction to address masking issues**
In the spectral extraction, the masking is setting masked values to nan. Optspec.py uses regular sums, so you end up with fully rejected columns if there was a single bad pixel/outlier, which can get out of hand if you have may bad/trimmed pixels. Switching to nansum fixes this.

**3. allows for skipping/fitting only certain bins in S5**
Simple change to allow input in S5 that lets you only fit certain bins, or skip certain bins. Also wraps the fitting in a try/except statement so a failed fit doesn't crash the full run.

**4.  updates aperture to be centered properly**
Based on discussions with Kevin, aperture extraction was from [pos-aperture: pos+aperture], which means you get one fewer pixel above the center than below. Updated to be centered, results in an odd number of pixels within the aperture centered at the center of the trace. So the total number of pixels is the 2*ap_hw + 1 now. Similar update for the background region. This also addresses a half pixel offset that is used for computing the center of mass for computing how to roll columns when straightening the trace which isn't removed prior to calculating the number of pixels to roll by.


src/eureka/S5_lightcurve_fitting/models/BatmanModels.py
```diff
@@ -121,7 +121,7 @@ def eval(self, channel=None, **kwargs):
 
             # Enforce physicality to avoid crashes from batman by returning
             # something that should be a horrible fit
-            if not ((0 < bm_params.rp) and (0 < bm_params.per) and
+            if not ((0 < bm_params.per) and
```
taylorbell57, line 124: Does batman actually allow negative radii without crashing? And if we're making this change, we should also make the same change in the src/eureka/S5_lightcurve_fitting/differentiable_models/StarryModel.py file (and make any tweaks that may be needed to allow negative radii in starry, e.g. just putting a negative sign on the lightcurve or something). The rp>0 prior is enforced here:
https://github.com/kevin218/Eureka/blob/3c1092661415f429c4ea685f898afa09679990d9/src/eureka/S5_lightcurve_fitting/differentiable_models/PyMC3Models.py#L353

src/eureka/S5_lightcurve_fitting/s5_fit.py
```diff
@@ -410,18 +409,68 @@ def fitlc(eventlabel, ecf_path=None, s4_meta=None, input_meta=None):
...
+                                meta, params = fit_channel(meta, time, flux, 
+                                                           channel, flux_err, 
+                                                           eventlabel, params,
+                                                           log, longparamlist, 
+                                                           time_units, 
+                                                           paramtitles, 
+                                                           chanrng, ld_coeffs,
+                                                           xpos, ypos, 
+                                                           xwidth, ywidth)
+                            except:
```
taylorbell57, line 426: Except statements should only be used to catch particular expected exceptions (that's a tough phrase to read/write XD). At a bare minimum, we should still print the error message to the terminal (otherwise people will have no idea how/why their fit failed), and ideally we'd only catch exceptions that we expect to encounter. For example, the danger with how this is currently written is that hitting Ctrl+c will cause an "exception" which will then trigger this except clause and the fit will just move on to the next channel. So if I realize I forgot to change a parameter in my ECF/EPF, I have to do Ctrl+c once per channel before the fitting run finally gets killed.

src/eureka/S5_lightcurve_fitting/s5_fit.py
```diff
@@ -410,18 +409,68 @@ def fitlc(eventlabel, ecf_path=None, s4_meta=None, input_meta=None):
...
+                            log.writelog('Saving results', 
+                                         mute=(not meta.verbose))
+                            me.saveevent(meta, (meta.outputdir +
+                                                'S5_'+meta.eventlabel +
+                                                "_Meta_Save"), save=[])
+                        else:
+                            print(' >>> skipping bin')
+                            continue  
+                                                
+                    elif hasattr(meta, 'skip_bins'):
```
taylorbell57, line 440: I think there's likely a cleaner way of doing this that avoids a bunch of duplicated code and allows both `fit_bins` and `skip_bins` to be used together. Let's make a list of channels that'll be iterated through (e.g. `chan_list`) using `np.arange` and taking into account any `fit_bins` and/or `skip_bins` settings, and then use that above instead of the current `for channel in range(chanrng):`

taylorbell57 (changes requested):

src/eureka/S5_lightcurve_fitting/models/BatmanModels.py
erinmmay: Batman does allow negative radii without crashing. I'm unsure if starry does, I've never used it.

src/eureka/S5_lightcurve_fitting/s5_fit.py
erinmmay: Do you have an alternative suggestion here? Because the problem arises when you set up a large run you intend to go overnight and then it crashes halfway because of a bad channel for any number of reasons.

taylorbell57: I'm not sure I've ever encountered such a crash before - can you give me an example of the kinds of issues you've run into before? Were they all related to some parameter not having any dynamic range when making the corner plot? Or what other kinds of sporadic crashes have you encountered?

erinmmay: Hmm, Not sure I have notes on these anywhere. I'm happy to remove this for now until it's encountered again and then we can look at the underlying cause of the crash.

kevin218: My guess is the crashes are related to memory issues.  We have a server that can be running several JWST analyses at different steps and sometimes the load is too much.  Since we can't be sure what the exception will be, my suggestion is to print the exception and then pause for ~5 seconds.  This should give the user enough time to Ctrl+c again and avoid the repetitive loop mentioned by Taylor.

erinmmay: My issues are rarely memory related (that typically crashes the whole thing, not just a single channel), and more like every point (or most points) in a channel are masked for some reason, or weird things with errors, or something else. Single channels crashing for random reasons happens quite often to me at native pixel res fitting. I prefer skipping and letting the rest continue, perhaps with a print out at the end to remind you which channels failed. With the option to fit only specific channels that is also in this PR that would allow the user to go back and try to refit only the failed channels, as well.

taylorbell57:
Another thing to consider is what impact all of this will have on the Stage 6 code - my guess is that many of the changes in this PR are likely to break much of the Stage 6 functionality.

erinmmay:
Yeah, sorry about the delay! I think the easiest/quickest thing will be to pull out the stage 5 changes for now since that requires some extra work and is the least important of these changes. I think then it's just looking at negative transits in starry (which I've never used and would definitely appreciate you looking into if you have used it) and stripping back the nansums so your changes to the masking can be done instead. If that plan sounds good I can make this a priority.

taylorbell57:
Yeah, punting your point 3 to another PR to be able to quickly merge points 1 and 4 sounds like a good idea to me. That's not to say that point 3 isn't worth pursuing, it's just going to be a bit finicky to get coded just right.

I'll quickly test a negative radius in starry to see what happens. Even if starry doesn't allow it (which it may since I don't know if all that fancy math would break with a negative radius), we can likely still make it work by manually inverting the transit signal

taylorbell57:
Alright, so while starry doesn't technically crash with negative radii, it doesn't end up giving the expected behaviour. I suggest the following which should have fairly minimial impacts on speed (you're welcome to implement it, or I can too). I'll note that I haven't tested this fully implemented or anything, but I have tested it in a Jupyter notebook and think it should work.

<br/>

In `eureka.S5_lightcurve_fitting.differentiable_models.StarryModel.eval()` in place of the current line which reads:
```
lcpiece = systems[chan].flux(time)
```
we need to instead put:
```
fstar, fp = systems[chan].flux(time, total=False)
if lessthan(planet.r, 0):
    fstar = 2-fstar
    fp *= -1
lcpiece = fstar+fp
```
where we'll need to define `greater` above in the `if eval:` section as:
```
        if eval:
            lib = np
            lessthan = np.less
            systems = self.fit.systems
        else:
            lib = tt
            lessthan = tt.lt
            systems = self.systems
```

<br/><br/>

And in `eureka.S5_lightcurve_fitting.differentiable_models.StarryModel.setup()` where we have:
```
r=temp.rp*temp.Rs
```
we instead need
```
r=tt.abs_(temp.rp)*temp.Rs
```
and in `eureka.S5_lightcurve_fitting.differentiable_models.StarryModel.update()` where we have:
```
r=temp.rp*temp.Rs
```
we instead need
```
r=np.abs(temp.rp)*temp.Rs
```

<br/><br/>

Finally, in `eureka.S5_lightcurve_fitting.differentiable_models.PyMC3Models.CompositePyMC3Model.__init__()`, where we currently have:
```
                                elif parname in ['rp', 'per', 'scatter_mult',
                                                 'scatter_ppm', 'c0', 'r1',
                                                 'r4', 'r7', 'r10']:
```
we'll need
```
                                elif parname in ['per', 'scatter_mult',
                                                 'scatter_ppm', 'c0', 'r1',
                                                 'r4', 'r7', 'r10']:
```

erinmmay:
@taylorbell57 I think I've successfully removed the necessary commits. double checking things right now

I've also added a new commit that addresses a plotting issue where the y-axis wasn't correct in the residual background plot. The aperture markings on that make much more sense now.

erinmmay:
ahh, I see it deleted s5_fit instead of ignoring the file... can you reject that change on your end or should I add it back in?

erinmmay:
👍 wasn't sure if you could do it in the "resolve conflicts" stage. I'll get that file added back in.

taylorbell57:
I'll add the PyMC3 edits to this PR now if that's alright with you?

erinmmay:
Yep, sounds good!

taylorbell57:
@erinmmay I don't think you've set it so that reviewers can make edits to your PR - I've made edits to PRs from forks before and can't seem to do so for yours, so my assumption is that you didn't check that box. Can you see if you're able to check it now (or confirm that it's already checked and I'm just missing something on my end)?

taylorbell57:
Aha, never mind - I am able to make commits. It's just been a long time since I did that and couldn't remember how

src/eureka/S3_data_reduction/straighten.py
```diff
@@ -41,7 +41,7 @@ def find_column_median_shifts(data, meta, m):
         smooth_coms[~np.isnan(smooth_coms)][-1]
 
     # Convert to interget pixels
-    int_coms = np.around(smooth_coms).astype(int)
+    int_coms = np.around(smooth_coms - 0.5).astype(int)
```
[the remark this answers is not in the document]
erinmmay: I spent a long time wrangling with this in my head too. Half a pixel is added when the center of light is calculated, then when converting to an integer for shifting the half pixel isn't removed so some columns were being shifted a pixel too much due to rounding.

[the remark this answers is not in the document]
taylorbell57: Ah good point

taylorbell57:
Alright, I think this PR looks ready to me - do you want to weigh in @kevin218, or should I approve it?

taylorbell57:
I can also confirm that all of the tests pass locally as well (GitHub automated testing is broken for all our PRs right now)

taylorbell57 (approved):

PR #534, opened 2023-05-09
HST ramp model and enabling calibrated spectra

kevin218:
I wrote up the HST ramp model and am committing it here for others to test down the road.

I added some code that masks the uncalibrated part of the NIRSpec 2d array.  It's quick and dirty, but it works.  Once I get more experience with other instruments, I'll certainly want to make changes/enhancements.  It's just not worth devoting more time until I know how the other instrument will work.


src/eureka/S5_lightcurve_fitting/models/HSTRampModel.py
```diff
@@ -0,0 +1,105 @@
+import numpy as np
+
+from .Model import Model
+from ...lib.readEPF import Parameters
+
+
+class HSTRampModel(Model):
```
taylorbell57, line 7: This will need to be added to the s5_fit.py file. E.g., this is how the `polynomial` gets added to the mix:
https://github.com/kevin218/Eureka/blob/3c1092661415f429c4ea685f898afa09679990d9/src/eureka/S5_lightcurve_fitting/s5_fit.py#L621-L637

taylorbell57: Also, we should add this model to the differentiable_models folder as well (annoying I know, but using PyMC3 has been super helpful for me at least). I definitely want to try merging the differentiable and non-differentiable models sometime, but it's a bit tricky to make the code fast for emcee,dynesty,lsq while still working for PyMC3. It should be doable, it'll just take a bit of finessing.

taylorbell57: All new parameters should also be added to the docs page (under "S5 Fitted Parameters" I believe) and documented in the S5 ECF and maybe the EPF template. Useful for me testing the code too as it's not immediately clear what range of values might be reasonable

src/eureka/S5_lightcurve_fitting/models/HSTRampModel.py
```diff
@@ -0,0 +1,105 @@
...
+        if self.parameters is None:
+            coeff_dict = kwargs.get('coeff_dict')
+            params = {hN: coeff for hN, coeff in coeff_dict.items()
+                      if hN.startswith('h') and hN[1:].isdigit()}
+            self.parameters = Parameters(**params)
+
+        # Set parameters for multi-channel fits
+        self.longparamlist = kwargs.get('longparamlist')
+        self.nchan = kwargs.get('nchan')
+        self.paramtitles = kwargs.get('paramtitles')
```
taylorbell57, line 39: These shouldn't be needed anymore, similar to the PolynomialModel and ExpRampModel files

src/eureka/S5_lightcurve_fitting/models/HSTRampModel.py
```diff
@@ -0,0 +1,105 @@
...
+        """A getter for the time."""
+        return self._time
+
+    @time.setter
+    def time(self, time_array):
+        """A setter for the time."""
+        self._time = time_array
+        if self.time is not None:
+            # Convert to local time
+            self.time_local = self.time - self.time[0]
```
taylorbell57, line 56: Take a look at the current PolynomialModel and ExpRampModel time setter functions. There's some added complexity with the recent addition of multwhite fits

src/eureka/S5_lightcurve_fitting/models/HSTRampModel.py
```diff
@@ -0,0 +1,105 @@
...
+        lcfinal : ndarray
+            The value of the model at the times self.time.
+        """
+        # Get the time
+        if self.time is None:
+            self.time = kwargs.get('time')
+
+        # Create the ramp from the coeffs
+        lcfinal = np.array([])
+        for c in np.arange(self.nchan):
```
taylorbell57, line 97: Again, you'll need to look at the current PolynomialModel and ExpRampModel files for nuances from the new multwhite fits related to channel numbers and the time arrays

taylorbell57 (changes requested):

src/eureka/S5_lightcurve_fitting/models/HSTRampModel.py
kevin218: @taylorbell57 Looks like these are still here.

kevin218:
Closing PR after merging #544

PR #536, opened 2023-05-15
Add super-sampling capabilities along cross-dispersion direction

kevin218:
The PR adds new functionality to super-sample the pixel resolution along the cross-dispersion direction for NIRCam, NIRSpec, and MIRI spectroscopy.  The super-sampling factor is determined by meta.expand in the S3 ECF.


demos/JWST/S3_nircam_wfss_template.ecf
```diff
@@ -15,6 +15,7 @@ xwindow         [100,1700]  # Horizontal axis as seen in DS9
 src_pos_type    gaussian    # Determine source position when not given in header (Options: header, gaussian, weighted, max, or hst)
 record_ypos     True        # Option to record the y position and width for each integration (only records if src_pos_type is gaussian)
 dqmask          False       # Mask pixels with an odd entry in the DQ array
+expand          1           # Super-sampling factor along cross-dispersion direction
```
taylorbell57, line 18: This should be added to the other S3 templates as well

src/eureka/S3_data_reduction/s3_reduce.py
```diff
@@ -111,6 +111,19 @@ def reduce(eventlabel, ecf_path=None, s2_meta=None, input_meta=None):
     else:
         meta = me.mergeevents(meta, s2_meta)
 
+    # Increase relevant meta parameter values
+    if not hasattr(meta, 'expand'):
+        meta.expand = 1
+    if meta.expand > 1:
+        meta.spec_hw *= meta.expand
+        meta.bg_hw *= meta.expand
```
taylorbell57, line 119: Can you confirm that this does not have any unintended consequences on filenames or foldernames?

tests/NIRCam_ecfs/S3_NIRCam.ecf
```diff
@@ -1,5 +1,6 @@
 # Eureka! Control File for Stage 3: Data Reduction
 
+inst        nircam      # Only needed during testing since we haven't run S2
```
taylorbell57, line 3: This shouldn't need to be added - the pipeline should be able to run from jwst-produced S2 files. We haven't needed it so far - where is the new requirement coming from?

src/eureka/S3_data_reduction/s3_reduce.py
taylorbell57: Ah, I can see from your testing edits that it does change the filenames. I really don't think that's a good idea - ap8 with spec_hw=8 & expand=1 is very different from ap8 with spec_hw=4 & expand=2. I think it's much better to leave the filenames unchanged

tests/test_MIRI.py
```diff
@@ -126,16 +126,16 @@ def test_MIRI(capsys):
     # run assertions for S3
     meta.outputdir_raw = (f'data{os.sep}JWST-Sim{os.sep}MIRI{os.sep}'
                           f'Stage3{os.sep}')
-    name = pathdirectory(meta, 'S3', 1, ap=4, bg=10)
+    name = pathdirectory(meta, 'S3', 1, ap=8, bg=20)
```
taylorbell57, line 129: I think all these edits should be reverted after adjusting the filename scheme as I commented above.

src/eureka/lib/naninterp1d.py
```diff
@@ -0,0 +1,51 @@
...
+    """
+    nans, x = nan_helper(y)
+    if len(y[~nans]) > 0:
+        y[nans] = np.interp(x(nans), x(~nans), y[~nans])
+    else:
+        y[nans] = replace_val
+    return y
+
+
+def nan_helper(y):
```
taylorbell57, line 30: This doesn't really feel like it needs to be a separate function to me, but we can leave it as one if you want. I'd likely just do the following up above:
```
nans = np.isnan(y)
x = lambda z: z.nonzero()[0]
```

taylorbell57 (changes requested):
Is the expand feature not something that could be relevant for WFC3? I guess it's probably not that useful for scanning mode observations since the trace is so large, but it could be useful for staring mode

tests/NIRCam_ecfs/S3_NIRCam.ecf
kevin218: This comes from Line 120 in s3_reduce.py.
```
if meta.inst == 'miri':
    meta.xwindow[0] *= meta.expand
    meta.xwindow[1] *= meta.expand
else:
    meta.ywindow[0] *= meta.expand
    meta.ywindow[1] *= meta.expand
```
`meta.inst` gets set when running Stage 2 of Eureka!, but never gets set in Stage 3.

kevin218: I don't see an obvious workaround because `meta = util.readfiles(meta, log)` can't get called until the log file is defined, but that doesn't happen until we're inside the double for loop and we don't want to repeatedly expand xwindow and ywindow each time we iterate.

src/eureka/S3_data_reduction/s3_reduce.py
taylorbell57: I still think changing the folder names based on the `expand` setting is really confusing. I think it's worth just dividing by `meta.expand` in the few rare cases where we need the the input `spec_hw` and `bg_hw`. I think the one area where we define the output folder and Fig 3304 are the only places where we'd need to divide by `meta.expand`

tests/NIRCam_ecfs/S3_NIRCam.ecf
taylorbell57: > This comes from Line 120 in s3_reduce.py.
> 
> ```
> if meta.inst == 'miri':
>     meta.xwindow[0] *= meta.expand
>     meta.xwindow[1] *= meta.expand
> else:
>     meta.ywindow[0] *= meta.expand
>     meta.ywindow[1] *= meta.expand
> ```
> 
> `meta.inst` gets set when running Stage 2 of Eureka!, but never gets set in Stage 3. This means we can't currently load in JWST-produced Stage 2 files without setting `inst` in the ECF.

Why can't these lines just be moved to the `inst.read` functions? I see no reason to have them in s3_reduce.py. There is already code in the miri.py file that only gets run for the first batch, so we already have the capability to do special things the first time through the read function

src/eureka/lib/naninterp1d.py
taylorbell57: Looking more closely now, there's also no need to define the x lambda function. I think a simpler `naninterp1d` function would be:
```
nans = np.isnan(y)
if np.all(nans):
    y[nans] = replace_val
else:
    y[nans] = np.interp(np.where(nans)[0], np.where(~nans)[0], y[~nans])
return y
```

I'm not 100% sure you need the `[0]`'s above, but I think so

src/eureka/S3_data_reduction/s3_reduce.py
kevin218: Ok, I've made a few more edits in this section to:
- Prevent using expand with photometry
- Work correctly when spec_hw or bg_hw is a list
- Use folder names that divide by meta.expand

taylorbell57 (changes requested):
The pytest tests also failed - looks like it was related to gcc somehow. Can you check if the tests pass locally for you? Hopefully this was just some one-off GitHub glitch

src/eureka/S3_data_reduction/miri.py
```diff
@@ -436,3 +439,34 @@ def flag_bg_phot(data, meta, log):
...
+    -----
+    History:
+
+    - 2023-07-17, KBS
+        Initial version.
+    """
+    # Convert from MJy/sr to mJy
+    log.writelog("  Calibrated MIRI spectra not yet supported...",
+                 mute=(not meta.verbose))
+
```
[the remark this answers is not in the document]
kevin218: I would think so, but I also haven't tested a MIRI dataset and have learned not to assume that MIRI will work the same way as the NIR instruments.

src/eureka/S3_data_reduction/s3_reduce.py
```diff
@@ -355,18 +366,10 @@ def reduce(eventlabel, ecf_path=None, s2_meta=None, input_meta=None):
...
-                                 mute=(not meta.verbose))
-                    boolmask = np.abs(data.flux.data) > cutoff
-                    data['flux'].data = np.where(np.abs(data.flux.data) > 
-                                                 cutoff, 0,
-                                                 data.flux.data)
-                    log.writelog(f"    Zeroed {np.sum(boolmask.data)} " + 
-                                 "pixels in total.", mute=(not meta.verbose))
+                # Instrument-specific steps for generating
+                # calibrated stellar spectra
+                if meta.convert_to_e is False:
```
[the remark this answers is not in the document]
kevin218: Yeah, I think that would be more informative.

src/eureka/S4_generate_lightcurves/s4_genLC.py
```diff
@@ -109,8 +109,12 @@ def genlc(eventlabel, ecf_path=None, s3_meta=None, input_meta=None):
     meta.run_s4 = None
     for spec_hw_val in meta.spec_hw_range:
         for bg_hw_val in meta.bg_hw_range:
+            if not isinstance(bg_hw_val, str):
+                # Only divide if value is not a string (spectroscopic modes)
+                bg_hw_val //= meta.expand
             meta.run_s4 = util.makedirectory(meta, 'S4', meta.run_s4,
-                                             ap=spec_hw_val, bg=bg_hw_val)
+                                             ap=spec_hw_val//meta.expand, 
+                                             bg=bg_hw_val)
```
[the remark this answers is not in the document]
kevin218: Are you suggesting a folder naming structure of ``apX_bgY_expZ``?

kevin218:
> The pytest tests also failed - looks like it was related to gcc somehow. Can you check if the tests pass locally for you? Hopefully this was just some one-off GitHub glitch

I'm not sure why it keeps failing on GitHub.  The tests all pass on my machine.  Maybe it'll work itself out in a day or two.

src/eureka/S4_generate_lightcurves/s4_genLC.py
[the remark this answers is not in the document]
taylorbell57: No, just to move the repeated  `//expand` code inside the function to avoid repetition

src/eureka/S3_data_reduction/miri.py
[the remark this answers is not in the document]
taylorbell57: Sounds good, no need to wait for MIRI, and I can help get MIRI integrated too since I want some calibrated spectra as well. I'd just wanted to know if you'd tried already and encountered an issue so that I knew what I'd be up against

src/eureka/S4_generate_lightcurves/s4_genLC.py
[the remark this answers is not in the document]
kevin218: This suggestion risks creating other problems down the road.  `spec_hw_val` and `bg_hw_val` are already referenced later on in s4_genLC.py, so we would eventually need to apply `//= meta.expand`.  With the current implementation, we can always rely on `spec_hw_val` and `bg_hw_val` as the non-super-sampled representations, whereas `meta.spec_hw` and `meta.bg_hw` are the super-sampled representations.

taylorbell57 (approved):

PR #539, opened 2023-06-15
Several bug fixes

### LINKED ISSUE #476: Help users resolve wave_max outside array warning in S4
To help avoid others encountering the first of your issues (wave_max larger than the longest wavelengt), perhaps someone can add onto the warning in Stage 4 to say to consider extending the xwindow in Stage 3 if the wave_min or wave_max are outside of the bounds of the wavelength array.

_Originally posted by @taylorbell57 in https://github.com/kevin218/Eureka/issues/465#issuecomment-1281541094_

### LINKED ISSUE #483: [Bug]: ECF files parsing removes spaces in strings
### FAQ check

### Instrument

Other (any stage)

### What happened?

When specifying a filepath in an ECS file that contains spaces, for example `/home/user/my data` the program fails giving back the error:
```
WARNING: Unable to find an output metadata file from Eureka!'s S1 in the folder:
"/home/user/mydata/"
Assuming this S1 data was produced by the JWST pipeline instead.
Traceback (most recent call last):
  File "/home/user/eureka/run_eureka.py", line 22, in <module>
    s2_meta = s2.calibrateJWST(eventlabel, ecf_path=ecf_path)
  File "/home/user/miniconda3/envs/eureka/lib/python3.9/site-packages/eureka/S2_calibrations/s2_calibrate.py", line 82, in calibrateJWST
    meta = util.find_fits(meta)
  File "home/user/miniconda3/envs/eureka/lib/python3.9/site-packages/eureka/lib/util.py", line 339, in find_fits
    raise AssertionError(message)
AssertionError: Unable to find any "rateints.fits" files in the inputdir: 
"/home/user/mydata/"!
You likely need to change the inputdir in S2_nirspec_fs_template.ecf to point to the folder containing the "rateints.fits" files.
```

The parsing function strips all whitespace characters from the string. Using quotes or backspaces does not work. Looking at the parsing function `MetaClass.Read` in `eureka/lib/util.py` it seems like [these lines](https://github.com/kevin218/Eureka/blob/3389d9b56dd632f2568c8372b1dc340318e7393d/src/eureka/lib/readECF.py#L181-L183) are the culprit:

```
        for line in cleanlines:
            name = line.split()[0]
            val = ''.join(line.split()[1:])
```

I agree with the suggestion in #239 as using an existing parser for an established config file format would probably be the most robust approach. One could of course also change the parser such that spaces after the initial split between parameter name and value are conserved. Happy to help with a small bug fix.

Linux Ubuntu 22.10

### Code of Conduct

### LINKED ISSUE #501: [Bug]: custom ld_file_white fail to read
### Instrument

### What is your suggestion?

Hi! Thanks a lot for this super useful package!
When I try to use custom `ld_file_white`, it fails to read the `.txt` file because `numpy` array shape is not correct. I constructed my `ld_file_white.txt` as you posted in the doc: columns for u1 and u2, rows for different wavelength. When using np.loadtxt() in `S5_fit.py` to load .txt, it cannot read `array([[0.02468399, 0.15435096]])` but `array([0.02468399, 0.15435096])`, i think this is kind of the reason why it crashes when call `BatmanTransitModel`. I would suggest add one line  `ld_coeffs = ld_coeffs[np.newaxis,:]` in line 199 under `if meta.whitep:` condition. 
But if I am inputing a wrong format of .txt file, just ignore this request. Thanks!

### Error traceback output

Traceback (most recent call last):
  File "/home/qiao/Documents/JWST/hd209458/2/dataanalysis/run_eureka.py", line 29, in <module>
    s5_meta = s5.fitlc(eventlabel, ecf_path=ecf_path, s4_meta=None)
  File "/home/qiao/Documents/JWST/Eureka/src/eureka/S5_lightcurve_fitting/s5_fit.py", line 233, in fitlc
    meta, params = fit_channel(meta, lc, time, flux, 0, flux_err,
  File "/home/qiao/Documents/JWST/Eureka/src/eureka/S5_lightcurve_fitting/s5_fit.py", line 421, in fit_channel
    t_transit = m.BatmanTransitModel(parameters=params, name='transit',
  File "/home/qiao/Documents/JWST/Eureka/src/eureka/S5_lightcurve_fitting/models/BatmanModels.py", line 79, in __init__
    self.ld_file_array[c][param-1]
IndexError: invalid index to scalar variable.

_No response_

_No response_

_No response_

### Code of Conduct

### LINKED ISSUE #521: [Bug]: S1 custom bias file and skip_superbias overwritten by Eureka_SuperBiasStep
### FAQ check

### Instrument

NIRSpec (Stages 1-3)

### What happened?

I'm working on reproducing the ERS' PRISM data reduction, and I noticed that my S1 runs weren't using the custom bias file I had provided. Looking at the S1 code, I see it is because all the values passed into the superbias step get over-written by the line `self.superbias = Eureka_SuperBiasStep()` which also over-writes the specified skip_superbias setting. This line should be done earlier on so that it does not over-write other ECF settings.

### Code of Conduct

### LINKED ISSUE #528: [Bug]: S3 drift plots are overwritten
### FAQ check

### Instrument

Other (any stage)

### What happened?

The Stage 3 DriftYPos and DriftYWidth figures made for spectroscopy do not use the _file## naming scheme, so they are overwritten during each batch, leaving only the last batch's figure remaining.

### Code of Conduct

### LINKED ISSUE #531: [Enhancement]: Further clarify manual_clip conventions on ECF documentation page
### FAQ check

### Instrument

### What happened?

If trying to clip the indices for several lists, during Stage 5 manual clipping, the indices will shift. 

For example, if you're trying to clip every tenth index: 

You would expect the clipping to behave as it did in Stages 3 & 4, where the indices to be clipped would be designated as [[10,10], [20,20], [30,30]]. However, to achieve this example clipping, at the moment, the indices need to be specified as [[10,11], [19,20], [28,29]].  

### Error traceback output

_No response_

_No response_

_No response_

_No response_

### Code of Conduct

### LINKED ISSUE #537: [Bug]: S4 wave_1d_trimmed variable
### FAQ check

### Instrument

Other (any stage)

### What happened?

In Stage 3 and 4, with some MIRI observations I get ~4 crazy noisy columns in my 2D lightcurves. Initially I thought there was something wrong with the data, but if in Stage 4 I change the `wave_min` to 7 instead of 5, then the left four columns of the 2D lightcurve are still horribly bad even though they were fine in the previous plot. Clearly there must be a bug in the normalization function or something.

With `wave_min` set to 5:
fig4101_2D_LC

With `wave_min` set to 7:
fig4101_2D_LC

### Error traceback output

_No response_

_No response_

_No response_

Using the version of Eureka! from the dev/tjb branch, but with all the installed packages I've been using for months (so other package updates can't be the cause).

### Code of Conduct

taylorbell57:
This PR:

- Makes minor cosmetic changes to several log statements
- Removes the h5py upper limit which is no longer needed and breaks python>3.9 installs
- Avoids over-writing ypos and ywidth plots (figs 3104 & 3105) when there are multiple segments (resolves #528)
- Resolves #537 by removing the bug-inducing and unnecessary `wave_1d_trimmed` variable
- Resolves #521 by moving the `self.superbias = Eureka_SuperBiasStep()` line earlier
- Resolves #501 by reshaping `ld_coeffs` for for input LD files for white lightcurves.
- Makes progress on #540, but we should still add documentation about exotic-ld on the installation instructions page
- Resolves #531 by further clarifying the `manual_clip` index convention
- Resolves #476 by helping users understand how to get access to longer or shorter wavelengths in S4 if they are outside of the current wavelength range.
- Resolves #483 by using the built-in shlex package to preserve spaces in quoted substrings since some people put spaces in their folder names.

### DIFF
--- a/src/eureka/S1_detector_processing/s1_process.py
+++ b/src/eureka/S1_detector_processing/s1_process.py
@@ -163,6 +163,11 @@ class EurekaS1Pipeline(Detector1Pipeline):
         - February 2022 Aarynn Carter /  Eva-Maria Ahrer
             Updated for JWST version 1.3.3, code restructure
         '''
+        # Define superbias offset procedure
+        self.superbias = Eureka_SuperBiasStep()
+        self.superbias.s1_meta = meta
+        self.superbias.s1_log = log
+
         # Figure out which instrument we're working on
         with fits.open(filename) as f:
             instrument = f[0].header['INSTRUME']
@@ -201,11 +206,6 @@ class EurekaS1Pipeline(Detector1Pipeline):
             self.firstframe.skip = meta.skip_firstframe
             self.lastframe.skip = meta.skip_lastframe
             self.rscd.skip = meta.skip_rscd
-
-        # Define superbias offset procedure
-        self.superbias = Eureka_SuperBiasStep()
-        self.superbias.s1_meta = meta
-        self.superbias.s1_log = log
 
         # Define ramp fitting procedure
         self.ramp_fit = Eureka_RampFitStep()
--- a/src/eureka/S1_detector_processing/s1_process.py
+++ b/src/eureka/S1_detector_processing/s1_process.py
@@ -162,7 +162,12 @@ class EurekaS1Pipeline(Detector1Pipeline):
         - February 2022 Aarynn Carter /  Eva-Maria Ahrer
             Updated for JWST version 1.3.3, code restructure
         '''
-        # Run the pipeline
+        # Over-write the superbias offset function
+        self.superbias = Eureka_SuperBiasStep()
+        self.superbias.s1_meta = meta
+        self.superbias.s1_log = log
+
+        # Figure out which instrument we're working on
         with fits.open(filename) as f:
             instrument = f[0].header['INSTRUME']
 
@@ -201,11 +206,6 @@ class EurekaS1Pipeline(Detector1Pipeline):
             self.lastframe.skip = meta.skip_lastframe
             self.rscd.skip = meta.skip_rscd
 
-        # Define superbias offset procedure
-        self.superbias = Eureka_SuperBiasStep()
-        self.superbias.s1_meta = meta
-        self.superbias.s1_log = log
-        
         # Define ramp fitting procedure
         self.ramp_fit = Eureka_RampFitStep()
         self.ramp_fit.algorithm = meta.ramp_fit_algorithm
--- a/docs/source/ecf.rst
+++ b/docs/source/ecf.rst
@@ -842,7 +842,7 @@ where starry replaces both the batman_tr and batman_ecl models and offers a more
 
 manual_clip
 '''''''''''
-Optional. A list of lists specifying the start and end integration numbers for manual removal. E.g., to remove the first 20 data points specify [[0,20]], and to also remove the last 20 data points specify [[0,20],[-20,None]].
+Optional. A list of lists specifying the start and end integration numbers for manual removal. E.g., to remove the first 20 data points specify [[0,20]], and to also remove the last 20 data points specify [[0,20],[-20,None]]. If you want to clip the 10th integration, this would be index 9 since python uses zero-indexing. And the manual_clip start and end values are used to slice a numpy array, so they follow the same convention of *inclusive* start index and *exclusive* end index. In other words, to trim the 10th integrations, you would set manual_clip to [[9,10]].
 
 
 Limb Darkening Parameters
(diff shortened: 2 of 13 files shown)


src/eureka/S1_detector_processing/s1_process.py
```diff
@@ -163,7 +163,12 @@ def run_eurekaS1(self, filename, meta, log):
         - February 2022 Aarynn Carter /  Eva-Maria Ahrer
             Updated for JWST version 1.3.3, code restructure
         '''
-        # Run the pipeline
+        # Define superbias offset procedure
```
kevin218, line 166: What's the reason for moving this up from where it was before?

taylorbell57: That was to resolve #521 by moving the `self.superbias = Eureka_SuperBiasStep()` line earlier

kevin218 (approved):

PR #543, opened 2023-06-22
Edit DL script to put HST files in same directory.

jbrande:
Edited the demo HST MAST download script to remove the call to `sortHST()` that splits up the direct images and spectra into different directories. Also edited the S3 ECF docs page to remove text that implies these should be split up into sci and cal directories, and clarified they should be in the same directory. This should remove any confusion by new HST users (and me, after several weeks of not thinking about HST).

Should we also remove `sortHST()` as a function? Probably not, right?

### DIFF
--- a/demos/HST/download_data_HST_template.py
+++ b/demos/HST/download_data_HST_template.py
@@ -34,9 +34,7 @@ for vis in visits:
     if result is not None:
         # Consolodate and move data into new directory
         md.consolidate(result, final_dir)
-        # Sort data into science and calibration folders (scan vs direct image)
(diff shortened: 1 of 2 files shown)


taylorbell57 (dismissed):
This looks good to me, but you'll need to resolve the merge conflict first

jbrande:
The merge conflict was easy to fix but the supposed pytest failure is actually something weird in codecov that's failing after the tests run. No clue what it is.

taylorbell57 (approved):
I'm not too worried, the actual pytest tests passed, they just failed to upload their results which happens sometimes

PR #544, opened 2023-06-30
HST ramp model and enabling calibrated spectra - v2

kevin218:
This PR is in response to requested changes to #534.  I'm not sure why my push created a new PR, but here we are.  Once this gets approved, I'll delete the other PR.


src/eureka/S3_data_reduction/s3_reduce.py
```diff
@@ -340,6 +340,19 @@ def reduce(eventlabel, ecf_path=None, s2_meta=None, input_meta=None):
...
+                if meta.convert_to_e is False and meta.inst == 'nirspec':
+                    cutoff = 1e-4
+                    log.writelog("  Masking manually identified bad pixels...",
+                                 mute=(not meta.verbose))
+                    boolmask = np.abs(data.flux.data) > cutoff
+                    log.writelog(f"    Masking {np.sum(boolmask.data)} " + 
+                                 "pixels.", mute=(not meta.verbose))
+                    data['flux'].data = np.where(np.abs(data.flux.data) > 
+                                                 cutoff, 0,
+                                                 data.flux.data)
```
taylorbell57, line 355: Why set them to zero instead of updating the mask array?

src/eureka/S5_lightcurve_fitting/s5_fit.py
```diff
@@ -669,17 +669,25 @@ def fit_channel(meta, time, flux, chan, flux_err, eventlabel, params,
...
-        t_ramp = ExpRampModel(parameters=params, name='ramp', fmt='r--',
-                              log=log, time=time, time_units=time_units,
-                              freenames=freenames,
-                              longparamlist=lc_model.longparamlist,
-                              nchannel=chanrng,
-                              nchannel_fitted=nchannel_fitted,
-                              fitted_channels=fitted_channels,
-                              paramtitles=paramtitles,
-                              multwhite=lc_model.multwhite,
-                              nints=lc_model.nints)
```
taylorbell57, line 681: These removed lines will need to be added back (as they're important for `multwhite` fits) and presumably need to be added to the `HSTRampModel` as well

taylorbell57 (changes requested):

src/eureka/S3_data_reduction/s3_reduce.py
kevin218: The uncalibrated BG region is everything >8 pixels from the trace.  I originaly tried masking those pixels, but Eureka! started throwing way too many errors for that to be a reliable solution.  I ended up picking this solution, which effectively converts the uncalibrated BG region to MJ/sr (by setting the values to 0 since GLBS already happened in Stage 1).

taylorbell57:
Thanks @kevin218, for some reason this time I did get emails about your commits, but better safe than sorry. I'm completely occupied with OWL this week, but this is on my to-do list for next week

src/eureka/S3_data_reduction/s3_reduce.py
```diff
@@ -340,6 +340,19 @@ def reduce(eventlabel, ecf_path=None, s2_meta=None, input_meta=None):
                 # Manually mask regions [colstart, colend, rowstart, rowend]
                 if hasattr(meta, 'manmask'):
                     data = util.manmask(data, meta, log)
+                
+                # Mask uncalibrated BG region for NIRSpec observations
+                # Code used for generating a calibrated stellar spectrum
+                if meta.convert_to_e is False and meta.inst == 'nirspec':
+                    cutoff = 1e-4
+                    log.writelog("  Masking manually identified bad pixels...",
+                                 mute=(not meta.verbose))
```
taylorbell57, line 349: This comment doesn't seem to belong here as these are not manually identified pixels being masked, right?

src/eureka/S3_data_reduction/s3_reduce.py
taylorbell57: But there is no check that GLBS has happened in Stage 1, and it would seem entirely possible to me that someone would not want to run GLBS or want to investigate not doing it

taylorbell57 (changes requested):

src/eureka/S3_data_reduction/s3_reduce.py
kevin218: Going back to my initial PR...
``I added some code that masks the uncalibrated part of the NIRSpec 2d array. It's quick and dirty, but it works. Once I get more experience with other instruments, I'll certainly want to make changes/enhancements. It's just not worth devoting more time until I know how the other instrument will work.``

If someone decides not to use GLBS in Stage 1 and wants calibrated NIRSpec pectra, then their results will not be accurate.  For NIRSpec, there's just no way of accurately subtracting the BG after performing the Photom step in Stage 2.

taylorbell57 (approved):

PR #549, opened 2023-07-07
Fixing buggy masking of entire wavelength LCs

### LINKED ISSUE #547: [Bug]: Resurfacing of S3 NaN masking bug
### FAQ check

### Instrument

Other (any stage)

### What happened?

In S3, there is once again a bug that frequently causes entire wavelengths to become masked, most likely because of an improperly handled/masked NaN somewhere. I first see it pop up in the fig3302_file00_int0000_Spectrum.png files, so it doesn't appear to be a bug with the lightcurve normalization code. My fig3303_file00_int0000_Profile.png files do have some dark pixels which are likely NaNs or masked pixels, but I can't immediately tell if they're coincident with the masked spectroscopic lightcurves.

### Error traceback output

_No response_

_No response_

_No response_

_No response_

### Code of Conduct

taylorbell57:
Resolves #547

Because the apmask wasn't applied when computing the stdspec, any wavelength that ever had a masked pixel within the source aperture would end up having it's entire lightcurve be masked. This PR applies the mask to the apdata-like arrays and throws in some extra np.ma calls in optspex for good measure.

### DIFF
--- a/src/eureka/S3_data_reduction/optspex.py
+++ b/src/eureka/S3_data_reduction/optspex.py
@@ -549,8 +549,8 @@ def clean_median_flux(data, meta, log, m):
             log.writelog('  Using a default value of median_thresh=5',
                          mute=(not meta.verbose))
             meta.median_thresh = 5
-        outliers = sigma_clip(residuals, sigma=meta.median_thresh,
-                              maxiters=None, axis=1, cenfunc='median')
+        outliers = sigma_clip(residuals, sigma=meta.median_thresh, maxiters=5,
+                              axis=1, cenfunc=np.ma.median, stdfunc=np.ma.std)
 
         # Interpolate over bad pixels
         clean_med = np.zeros((ny, nx))
--- a/src/eureka/S3_data_reduction/optspex.py
+++ b/src/eureka/S3_data_reduction/optspex.py
@@ -30,14 +30,42 @@ def standard_spectrum(data, apdata, apmask, aperr):
     data : Xarray Dataset
         The updated Dataset object in which the spectrum data will stored.
     """
-    # Get numbers to account for missing pixels
-    npix = apmask.shape[1]
-    nmasked = np.sum(apmask == 0, axis=1)
-    correction = npix/(npix-nmasked)
+    # Replace masked pixels with spectral neighbors
+    apdata_cleaned = np.copy(apdata)
+    aperr_cleaned = np.copy(aperr)
+    for t, y, x in np.array(np.where(apmask == 0)).T:
+        # Do not extend to negative indices (short and long wavelengths
+        # do not have similar profiles)
+        lower = x-2
+        if lower < 0:
+            lower = 0
+        # Get mask for current neighbors
+        mask_temp = np.append(apmask[t, y, lower:x],
+                              apmask[t, y, x+1:x+3])
 
-    # Apply mask and missing pixel correction
-    stdspec = np.nansum((apdata*apmask), axis=1)*correction
-    stdvar = np.nansum((aperr*apmask)**2, axis=1)*correction
+        # Gather current data neighbors and apply mask
+        replacement_val = mask_temp*np.append(apdata_cleaned[t, y, lower:x],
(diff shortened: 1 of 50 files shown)


kevin218 (commented):
If we're going to convert optspex to masked arrays, it may be confusing to also carry around submask.  I would think the masked arrays should replace any need for submask, but some testing is needed to confirm that both produce the same results.

taylorbell57:
Alright, running the tests locally now, but I think this one should be ready for you now @kevin218

taylorbell57:
Confirmed that all tests pass locallly

src/eureka/S3_data_reduction/optspex.py
```diff
@@ -28,8 +30,18 @@ def standard_spectrum(data, apdata, aperr):
     data : Xarray Dataset
         The updated Dataset object in which the spectrum data will stored.
     """
-    data['stdspec'] = (['time', 'x'], np.sum(apdata, axis=1))
-    data['stdvar'] = (['time', 'x'], np.sum(aperr ** 2, axis=1))
+    # Get numbers to account for missing pixels
+    npix = apmask.shape[1]
+    nmasked = np.sum(apmask == 0, axis=1)
+    correction = npix/(npix-nmasked)
```
kevin218, line 36: I'm not keen on this correction strategy. The flux isn't uniform down a pixel column. It's really concentrated into 2-3 pixels. Thus, there's a 50/50 chance of under/over correcting and no chance getting it right. Ultimately, the optimal spectral extraction will get it right.

I don't have a helpful suggestion at the moment. Happy to chat.

kevin218 (changes requested):

src/eureka/S3_data_reduction/optspex.py
kevin218: Maybe we can replace the bad pixel with an average of its neighbors on the left and right.

taylorbell57: While I agree that this is certainly not ideal, it was a small step in the right direction compared to not doing any correction based on the testing I did. I also agree that the optimal extraction ultimately figures things out all the same - this was just a beginning attempt at patching the standard spectra a bit since several people have asked me if they're able to use the standard spectra instead of the optimal spectra (for various reasons, e.g. easier aperture correction for flux calibrated spectra).

taylorbell57: That's not a bad suggestion either, although it comes with complications if there are adjacent masked pixels

taylorbell57: My most recent commit implements your suggestion (using two neighbors on either side since adjacent bad pixels are fairly likely) and looks like it works really quite well.

Here is the stdspec and optspec for a GTO MIRI observation that I got with my previous hacky method
fig3302_file00_int0000_Spectrum

while here is the stdspec and optspec of the same data with the new algorithm
fig3302_file00_int0000_Spectrum

The stdspec is now looking *much* better and is actually quite usable (for those that want it)

taylorbell57:
The tests all pass locally now FYI

src/eureka/S3_data_reduction/optspex.py
```diff
@@ -28,8 +30,46 @@ def standard_spectrum(data, apdata, aperr):
...
+        # Compute the mean of the unmasked data neighbors
+        replacement_val = np.nansum(replacement_val)/denom
+        # Replace masked value with the newly computed data value
+        apdata_cleaned[t, y, x] = replacement_val
+
+        # Gather current err neighbors and apply mask
+        replacement_val = mask_temp*np.append(aperr_cleaned[t, y, lower:x],
+                                              aperr_cleaned[t, y, x+1:x+3])
+        # Figure out how many err neighbors are being used
+        denom = np.sum(mask_temp)
```
kevin218, line 60: This line is unnecessary.  Denom hasn't changed.

kevin218 (commented):
Nice implementation! Only one small issue

kevin218 (approved):

PR #551, opened 2023-07-18
Updated install instructions to fix deprecation

taylorbell57:
The current pip installation method using egg= raises a warning about a deprecation that is coming with pip version 25.0. To get ahead of that, let's change things now.

The warning that is currently raised is:

> DEPRECATION: git+https://github.com/kevin218/Eureka.git@dev/tjb#egg=eureka[jwst] contains an egg fragment with a non-PEP 508 name pip 25.0 will enforce this behaviour change. A possible replacement is to use the req @ url syntax, and remove the egg fragment. Discussion can be found at https://github.com/pypa/pip/issues/11617

### DIFF
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -49,13 +49,13 @@ Once in your new conda environment, you can install the ``Eureka!`` package with
 
 .. code-block:: bash
 
-	pip install -e git+https://github.com/kevin218/Eureka.git@v0.9#egg=eureka[jwst]
+	pip install eureka[jwst]@git+https://github.com/kevin218/Eureka.git@v0.9
 
 Other specific branches can be installed using:
 
 .. code-block:: bash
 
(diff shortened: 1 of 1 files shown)


taylorbell57:
@kevin218, can we aim to merge this tiny PR before the Sagan Summer Workshop so that we don't end up with any headaches if the new pip version gets released between now and then

kevin218 (approved):

kevin218:
I don't know what's going on with pytest

taylorbell57:
Me neither, and I can't do a thorough test because apparently my miniconda installation is broken somehow (can't make new environments). I'm running a local test right now, but I know that this installation method works from the Colab notebooks

taylorbell57:
Tests run fine locally on my pre-existing conda environment, and I know this installation method works fine from the Google Colab installations I've been doing, so I'm just going to merge this now

PR #556, opened 2023-08-11
Removing 390Hz noise from LRS SLITLESSPRISM data

taylorbell57:
Some MIRI subarrays exhibit 390Hz noise in the pixels within a group.
We roughly knew about this from the LRS TSO commissioning
paper, but Michael Ressler did some great investigative work and showed
that there is a clear 390Hz noise in the group-level data which has a
fixed waveform but has a varying phase. I did some testing myself and
determined the waveform for the LRS SLITLESSPRISM subarray and wrote
code to find the phase. Interestingly, I found that the phase remains
constant within an integration, and while the phase varies a bit between
integrations, I noticed that it has cyclical variations with a period of
8 integrations (in at least one dataset). I couldn't figure out a clean
way to deterministically compute the phase however, so I just fit it in
my added code.

I had hoped that this would improve the high-frequency noise I've
observed in LRS spectra (the noise is in the final transit/eclipse
spectra). However, I have found that the 390Hz step seems to have nearly
zero impact on the transit/eclipse spectrum, while the related
MIRI-specific GLBS code I added at the same time also has little impact.
That said, I don't think I'm the only one who would want to investigate
this effect, so I am still contributing these edits to allow others to
investigate the impact on their data and allow for potential further
improvements.

I will note that this code was only built with the LRS SLITLESSPRISM
subarray in mind and has been tested on several such observations, but a
few other subarrays also exhibit a similar noise pattern. I don't have
any such observations to test the code on, so for now this step should
only be applied to LRS SLITLESSPRISM data.


taylorbell57:
I'll also point out that it's theoretically possible that this 390Hz noise removal could be combined with separate background observations to try to avoid self-subtraction. I believe early attempts at using separate background observations after the science observations were not very successful, but perhaps there'd be more success with the inclusion of this 390Hz noise removal code applied to both the science and calibration observations.

src/eureka/S1_detector_processing/remove390.py
```diff
@@ -0,0 +1,448 @@
...
+    
+    # Reshape and add together to get 2D time array
+    dt_rows = dt_rows.reshape((-1, 1))
+    dt_cols = dt_cols.reshape((1, -1))
+    pixel_times = dt_cols+dt_rows+(dt_grp*grp)+(dt_int*integ)
+    
+    return pixel_times
+
+
+def smoothData(tempdata, boundary='extend', box_width=100):
```
kevin218, line 71: Would it make sense to move this function to smooth.py? It might be useful for other routines.

Alternatively, should we use an existing smoothing function?

src/eureka/S1_detector_processing/remove390.py
```diff
@@ -0,0 +1,448 @@
...
+        A masked array containing all data for a single integration.
+    smooth : bool; optional
+        Whether or not boxcar smoothing should be applied when computing the
+        background level. This is useful when trying to measure the 390Hz noise
+        component which is largely but not entirely removed by row-by-row
+        background subtraction. True by default.
+
+    Returns
+    -------
+    _type_
```
kevin218, line 100: Fill in missing details

src/eureka/S1_detector_processing/remove390.py
```diff
@@ -0,0 +1,448 @@
...
+                                             NGROUPS, TSAMPLE, TGROUP))
+    pixel_times = np.array(pixel_times)
+    
+    # Subtract a smoothed background
+    pgrams = []
+    for grp in range(images2.shape[0]-1):
+        bg = computeBG(cleaned[grp], smooth=True)
+        cleaned[grp] -= bg
+    
+        if plot:
```
kevin218, line 279: Should we make this an isplots = 3?

src/eureka/S1_detector_processing/remove390.py
```diff
@@ -0,0 +1,448 @@
...
+
+    p0 = [0, ]
+    # Fit the 390 Hz noise
+    p1 = remove_390(p0, x, y)
+    if prnt:
+        print(p1)
+
+    if plot and meta.m == 0 and integ < meta.nplots:
+        # Show the fit to a small chunk of data
+        plt.figure(1501)
```
kevin218, line 295: 1501 is used for when isplots = 5. Maybe use 1301 if you go with isplots = 3.

kevin218:
What's the timeline on this PR? No hurry, just curious.

taylorbell57:
It'll probably only take me an hour or two to make all the required edits, just need to find the time to get around to that

sebastian-zieba:
Do you have an idea how your code compares to the 390 Hz correction, which is implemented in jwst==1.13 (https://github.com/spacetelescope/jwst/pull/7857)?
Couldn't we just use what the jwst pipeline does?

taylorbell57:
@sebastian-zieba, I haven't done a whole bunch of testing yet, but I did do a test recently with `jwst==1.13.4` and found that there was still significant 390Hz noise after starting with my previously downloaded _uncal files. I'll run another test or two to double-check that the `emicorr` step is being run (and re-download the _uncal files if I need to). I've also already contributed to a few papers that reference this new Eureka! feature, so I still want to add this code (at least for now) so that the results of those papers are reproducible.

src/eureka/S1_detector_processing/remove390.py
```diff
@@ -0,0 +1,436 @@
...
+                plt.pause(0.2)
+    
+    if returnp1:
+        return cleaned_no390, p1, integ
+    else:
+        return cleaned_no390, integ
+
+
+def run(input_model, log, meta):
+    """_summary_
```
kevin218, line 376: Missing description

kevin218 (approved):

PR #558, opened 2023-08-12
Minor bug fixes

taylorbell57:
This PR has some minor bug fixes. Specifically, it fixes some small bugs with S3 photometry code, and it fixed a bug where setting fill_value to 'boxcar' in Stage 4 still resulted in points being masked while they should've been left unmasked and simply replaced with the value from the boxcar smoothed data.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -4,40 +4,46 @@ channels:
     - defaults
 dependencies:
     - python==3.9.7
-    - asdf
     - astropy
     - batman-package
     - bokeh[version='<3.0']
     - ccdproc
-    - celerite
+    - celerite # Needed for GP
     - corner
     - cython
-    - dynesty[version='>1.0']
-    - emcee[version='>3.0.0']
-    - george
-    - gwcs
-    - h5py[version='<3.2']
+    - dynesty[version='>1.0'] # Lower limit needed for specific arguments
+    - emcee[version='>3.0.0'] # Lower limit needed for specific arguments
+    - flake8 # Needed for testing
+    - george # Needed for GP
+    - h5py
+    - ipykernel
     - ipython
+    - jupyter
     - lmfit
     - matplotlib
     - nbsphinx # Needed for documentation
-    - numpy[version='>=1.20.0']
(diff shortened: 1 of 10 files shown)


taylorbell57:
Sigh, well I removed some of GitHub's issues, but now we're back to some gcc error that I don't get locally (and I do remember seeing with some recent PRs until the new numpy blas errors starting popping up even more recently). I'll try to read through the error messages closely and see if I can figure out what the issue is

environment.yml
```diff
@@ -4,40 +4,46 @@ channels:
...
+    - george # Needed for GP
+    - h5py
+    - ipykernel
     - ipython
+    - jupyter
     - lmfit
     - matplotlib
     - nbsphinx # Needed for documentation
-    - numpy[version='>=1.20.0']
+    - numpy[version='>=1.20.0,<=1.23'] # Upper limit needed for Apple silicon
```
kevin218, line 25: I see that the upper limit is 1.23 here and 1.22 in [environment_pymc3.yml](https://github.com/kevin218/Eureka/pull/558/files#diff-15acc9e6ddc11f0b6781d4f3a13a02fe1aa14a2aeb776a107b611f63ab4a3c95).  What's the reason for this difference?  Would it just be better to remain consistent and adopt 1.22 here as well or should we aim for maximum flexibility on the version numbers?

kevin218:
I see that Pytest Testing fails with a new error, possibly relating to gcc.

taylorbell57:
Yeah, I'm actively working on that gcc issue - I'll re-request your review once I've figured out how to get gcc working again (my best guess is that a gcc version number changed and broke GitHub's testing)

taylorbell57:
FYI, I had no luck with changing the gcc version, so I reverted those commits and my next best guess is that a cython version is to blame. I'll have to come back to that later (perhaps in a few days). In the meantime, you could approve this PR if you want to and leave the resolution to the GitHub-exclusive gcc error for another PR

kevin218 (approved):

PR #559, opened 2023-08-14
Adding the option to impose positive values on fitted eclipse maps

mark-hammond:
The default Starry-based eclipse mapping method fits the spherical harmonic coefficients directly, which allows maps containing regions of negative emission. In some circumstances it is desirable to require positive maps only.

This PR therefore adds the "pixel sampling" method described in https://starry.readthedocs.io/en/latest/notebooks/PixelSampling/, which is activated by adding a "pixel_ydeg" independent EPF parameter. The fitted map can then be required to be positive with a True "force_positive_map" ECF flag, which assigns a LogNormal prior to each pixel. The new "record_map" ECF flag allows the posterior distribution of the fitted maps to be recorded and plotted.

If "pixel_ydeg" is present and "force_positive_map" is False, each pixel is assigned a Normal prior allowing negative values, which gives similar results to the original method.

If this looks OK, I'm happy to add the new parameters to the documentation. I have tested fits with a variety of pixel distributions and flag combinations on my own machine.


mark-hammond:
I've fixed the conflict in S5_lightcurve_fitting/differentiable_models/StarryModel.py where the main branch was updated to allow negative rp; the positive maps branch now allows that to happen with or without pixel sampling

taylorbell57:
Thanks for submitting this Mark - I'm excited to get this incorporated! Is it alright with you if I make a few edits directly to your PR?

mark-hammond:
Yes, go for it!

src/eureka/S5_lightcurve_fitting/differentiable_models/StarryModel.py
```diff
@@ -185,8 +195,51 @@ def setup(self):
             planet.theta0 = 180.0
             planet.t0 = temp.t0
 
+            # Pixel sampling setup
+            if 'pixel_ydeg' in self.paramtitles:
+                # Get pixel transform matrix and number of pixels
+                if self.ydeg == 2:
+                    self.oversample = 3
+                else:
+                    self.oversample = 2
```
taylorbell57, line 204: In the else part of this if/else statement, were you assuming ydeg=1 or ydeg>2? I'm trying to figure out how to generalize this to any random ydeg

src/eureka/S5_lightcurve_fitting/differentiable_models/StarryModel.py
```diff
@@ -185,8 +195,51 @@ def setup(self):
...
+                    oversample=self.oversample)[3]
+                self.npix = A.shape[1]
+
+                # Set prior to either be log normal, or normal around zero
+                if self.force_positivity:
+                    p = pm.LogNormal("p", mu=np.log(0.3*amp), tau=1.0,
+                                     shape=(self.npix,))
+                else:
+                    p = pm.Normal("p", mu=0.2*amp, sd=0.2*amp,
+                                  shape=(self.npix, ))
```
taylorbell57, line 216: Is a prior of 0.2*fp ± 0.2*fp really broad enough to be minimally informative and be hardcoded?

src/eureka/S5_lightcurve_fitting/differentiable_models/StarryModel.py
```diff
@@ -185,8 +195,51 @@ def setup(self):
...
+                    self.oversample = 2
+
+                A = planet.map.get_pixel_transforms(
+                    oversample=self.oversample)[3]
+                self.npix = A.shape[1]
+
+                # Set prior to either be log normal, or normal around zero
+                if self.force_positivity:
+                    p = pm.LogNormal("p", mu=np.log(0.3*amp), tau=1.0,
+                                     shape=(self.npix,))
```
taylorbell57, line 213: Similarly, is this prior really broad enough to be minimally informative and be hard coded?

src/eureka/S5_lightcurve_fitting/differentiable_models/StarryModel.py
```diff
@@ -155,12 +159,18 @@ def setup(self):
...
                 planet_map2 = starry.Map(ydeg=self.ydeg)
-                for ell in range(1, self.ydeg+1):
-                    for m in range(-ell, ell+1):
-                        if hasattr(temp, f'Y{ell}{m}'):
-                            planet_map[ell, m] = getattr(temp, f'Y{ell}{m}')
-                            planet_map2[ell, m] = getattr(temp, f'Y{ell}{m}')
-                amp = temp.fp/planet_map2.flux(theta=0)[0]
+                # Set up amplitude differently if using pixel sampling
+                if 'pixel_ydeg' in self.paramtitles:
+                    amp = temp.fp
```
taylorbell57, line 164: Have you confirmed that the retrieved fp values are the same between sinusoids and spherical harmonics? It's important that the final retrieved value be the actual eclipse depth and not some messy normalization factor that's hard to interpret

src/eureka/S5_lightcurve_fitting/differentiable_models/StarryModel.py
mark-hammond: I've updated this to be oversample = 3 always to use the appropriate 4L^2 scaling for the number of pixels for an equiangular grid

src/eureka/S5_lightcurve_fitting/differentiable_models/StarryModel.py
mark-hammond: I've changed this to be set by independent "pixel_prior_mean" and "pixel_prior_width" variables in the run parameters file; I didn't use the default free parameter setup as it doesn't include LogNormal priors so this seemed simpler. Using the peak value of the light curve (same as the fp parameter) works fine for both the mean and width parameters. I wouldn't expect a need to use anything different for either parameter, but this gives the option if needed

src/eureka/S5_lightcurve_fitting/differentiable_models/StarryModel.py
mark-hammond: See above comment, this is now settable in the run parameters file

mark-hammond:
I found that the pixel trace saving in gradient_fitters.py wasn't working as it needed converting to arviz format before saving (did it work for you previously? It could be different dependency versions) so I changed it to convert to arviz and save. I also made it only save if it's calculating a pixel map (as the rest of the trace is already saved elsewhere in the *_nuts_samples_*.h5 file without the pixel values)

mark-hammond:
I've also added the option to output the various parts of the calculated lightcurve from plot_s5.py with a "record_plot_data" switch, which I've needed for plotting things like residual eclipse mapping signals, as I don't think they're easily accessible from any of the default output options

src/eureka/S5_lightcurve_fitting/differentiable_models/StarryModel.py
mark-hammond: In fact, the way it's set up at the moment, fp is irrelevant when fitting a map with pixel sampling. We could define the pixel values as normalised by an fp value which we fit for, but they're still entirely degenerate with it which slows down the sampling for no benefit. I think the best way to do it is to use the "pixel_prior_mean" as the equivalent of fp, where the user puts in the expected planetary flux magnitude, then to post-process the actual eclipse depth if it's needed

src/eureka/S5_lightcurve_fitting/differentiable_models/StarryModel.py
mark-hammond: I've changed this to be "oversample=3" by default but allow it to be changed in the run parameters file, as oversample=3 may be more than needed for l>2

taylorbell57:
Alright, all of the code parts of this seem well-integrated now and I've done some pretty thorough manual testing. The main remaining task for me to do is update all the relevant parts of the documentation.

taylorbell57:
I'll wrap up with the documentation on Monday, and then I'll ping you for a review @kevin218 (though if you're bored tomorrow, you could get a head start)

tests/test_lightcurve_fitting.py
```diff
@@ -1,483 +1,467 @@
...
+    # Test the auto attribute assignment
+    assert params.param1.values == ['param1', 123.456, 'free']
+    assert params.param2.values == ['param2', 234.567, 'free', 200, 300]
+    # Test FileNotFoundError
+    with pytest.raises(FileNotFoundError):
+        Parameters(None, 'non-existent.epf')
+
+# Tests for the models.py module
+
+
```
kevin218, line 82: Remove extra lines

tests/test_lightcurve_fitting.py
```diff
@@ -1,483 +1,467 @@
...
+    assert model.time_units == 'MJD'
+
+
+def test_compositemodel(capsys):
+    """Tests for the CompositeModel class"""
+    model1 = models.Model()
+    model2 = models.Model()
+    comp_model = model1*model2
+    comp_model.name = 'composite'
+
```
kevin218, line 102: Do we need an assertion for this test?

tests/test_lightcurve_fitting.py
```diff
@@ -1,483 +1,467 @@
...
+    # Evaluate and test output
+    time = np.linspace(0, 1, 100)
+    exp_model.time = time
+    vals = exp_model.eval()
+    assert vals.size == time.size
+
+# Test for the simulations.py module
+
+
+def test_simulation(capsys):
```
kevin218, line 457: Remove extra lines

kevin218 (changes requested):

tests/conftest.py
```diff
@@ -0,0 +1,27 @@
+# conftest.py
+def pytest_collection_modifyitems(session, config, items):
+    """Modifies test items to ensure test functions run in a given order"""
```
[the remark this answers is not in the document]
taylorbell57: The motivation is to test the most basic things first before moving on to more complex tests. That way if there's an issue with something basic, it'll take fewer completed tests before uncovering the issue. This isn't a big deal for GitHub's automated testing, but I semi-regularly run the tests offline while developing, and this will let me interrupt the test run quicker. It also just seems like best practice to test smaller components before testing larger integrated sections of code.

tests/test_lightcurve_fitting.py
taylorbell57: The comments delineate a collection of multiple tests for one module, and are not just for one specific test; as a result, I don't want to have zero spaces between the comment and the subsequent test. And flake8 requires two blank lines immediately above each function, so I can't have just one blank line.

tests/conftest.py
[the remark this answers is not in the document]
taylorbell57: This also isn't a user-facing function btw; it modifies the order of tests when you do `pytest tests`

src/eureka/S5_lightcurve_fitting/s5_meta.py
```diff
@@ -146,6 +154,20 @@ def set_defaults(self):
...
+        if self.pixelsampling:
+            # Must be provided in the ECF if relevant
+            self.ydeg = getattr(self, 'ydeg')
+            # Compute the number of pixels used in sampling
+            map = starry.Map(ydeg=self.ydeg)
+            A = map.get_pixel_transforms(oversample=self.oversample)[3]
+            self.npix = A.shape[1]
+        else:
+            self.ydeg = getattr(self, 'ydeg', None)
+            self.npix = 0
```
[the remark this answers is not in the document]
taylorbell57: `ydeg` can be used when `pixelsampling == False` if the user is using starry's spherical harmonic phase function (without the pixelsampling technique). Setting it to `None` here allows the user to set priors on Ylm terms in their EPF, and then the code will infer the `ydeg` value.

`npix` is indeed not used when `pixelsampling == False`, but I think I might've needed to define a default value to make code simpler elsewhere and it also just seemed like good practice to me.

tests/test_lightcurve_fitting.py
taylorbell57: I don't think so, no. This just tests that two models can be combined using multiplication (using the `Model.__mul__` function) which is a utility we've never used.

kevin218 (approved):

PR #573, opened 2023-10-31
Add HSTRampModel to __init__ for proper importing

jbrande:
The various `__init__.py` files in `S5/models` and `S5/differentiable_models` needed to have the actual HST ramp models specified so they could be used when fitting WFC3 data. I also added the KeplerOrbit as that wasn't available either, but I have not actually tried to use that orbit model and just assumed it was an oversight.


taylorbell57 (approved):

taylorbell57:
Thanks for the patch!

PR #578, opened 2023-11-09
Removing Cython as a dependency

taylorbell57:
Cython was only needed for niriss_cython.pyx which (since we currently don't support NIRISS) could just be a python file instead of a cython file. This will hopefully resolve our GitHub automated testing (we'll see with this PR's tests).


taylorbell57:
Oops, forgot to remove cython from setup.cfg. I'll make another commit soon

kevin218 (approved):

PR #581, opened 2023-11-16
Logging Eureka! version and CRDS pmap

taylorbell57:
To help strengthen reproducibility and to allow users to manually control the used pmap, we now (silently) log the Eureka! version for Stages 1-6 and the CRDS pmap from Stages 1-3 (4-6 aren't relevant). For all future analyses, warnings will be printed to the terminal if the Eureka! version or CRDS pmap changes (only in the relevant stages), but this will not apply to previous partial reductions of the data (e.g., running the new code on v0.10 Stage 1 files won't raise any pmap or Eureka! version warnings).

### DIFF
--- a/docs/source/ecf.rst
+++ b/docs/source/ecf.rst
@@ -18,6 +18,12 @@ suffix
 ''''''
 Data file suffix (e.g. uncal).
 
+
+pmap
+''''
+Optional. If you want to use a specific CRDS context pmap (e.g. to reproduce someone else's work), you can specify the pmap number here. For example, to use ``jwst_1089.pmap``, set this ``pmap`` parameter to ``1089``.
+
+
 ramp_fit_algorithm
 ''''''''''''''''''
 Algorithm to use to fit a ramp to the frame-level images of uncalibrated files. Only default (i.e. the JWST pipeline) and mean can be used currently.
@@ -219,6 +225,9 @@ Data file suffix (e.g. rateints).
 .. note::
 	Note that other Instruments might used different suffixes!
 
+pmap
+''''
+Optional. If you want to use a specific CRDS context pmap (e.g. to reproduce someone else's work), you can specify the pmap number here. For example, to use ``jwst_1089.pmap``, set this ``pmap`` parameter to ``1089``.
 
 slit_y_low & slit_y_high
 ''''''''''''''''''''''''
@@ -317,6 +326,10 @@ An optional input parameter. If False (default), convert the units of the images
 poly_wavelength
 '''''''''''''''
 If True, use an updated polynomial wavelength solution for NIRCam longwave spectroscopy instead of the linear wavelength solution currently assumed by STScI.
+
+pmap
+''''
+Optional. If you want to use a specific CRDS context pmap (e.g. to reproduce someone else's work), you can specify the pmap number here. For example, to use ``jwst_1089.pmap``, set this ``pmap`` parameter to ``1089``.
 
 gain
 ''''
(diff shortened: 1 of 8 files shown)


taylorbell57:
I just remembered that the pmap differs between JWST and HST, and the current code also doesn't check if the user has set their pmap in their `.bashrc` file, so I'll switch to using `crds.get_context_name('jwst')` (and `crds.get_context_name('hst')` if there are any reference files that are automatically downloaded for HST, which I don't think there are) instead of the `crds.get_default_context()` I'm using right now

taylorbell57:
Hah, the automated testing also caught a bug in the cases where pmap is not explicitly set (`File 'jwst_jwst_1148.pmap.pmap' is not known to CRDS.`)

docs/source/ecf.rst
```diff
@@ -18,6 +18,12 @@ suffix
 ''''''
 Data file suffix (e.g. uncal).
 
+
+pmap
```
kevin218, line 22: In the Conversation tab, you mentioned explicitly calling `jwst` vs `hst`.  Is there a chance users will want to use the `pmap` variable to reference an HST context?  Or maybe a future telescope, like Pandora?  If so, the parameter could be `jwst_1089`.  If not, then the current implementation is fine.

src/eureka/S2_calibrations/s2_calibrate.py
```diff
@@ -74,6 +76,15 @@ def calibrateJWST(eventlabel, ecf_path=None, s1_meta=None, input_meta=None):
     else:
         meta = input_meta
 
+    # If a specific CRDS context is entered in the ECF, apply it.
+    # Otherwise, log and fix the default CRDS context to make sure it doesn't
+    # change between different segments.
+    if not hasattr(meta, 'pmap') or meta.pmap is None:
+        # Get just the numerical value
+        meta.pmap = crds.get_context_name('jwst')[5:-5]
+    os.environ['CRDS_CONTEXT'] = f'jwst_{meta.pmap}.pmap'
```
kevin218, line 85: If the ECF definition of `pmap` changes, here is where we also need to make changes.

src/eureka/S3_data_reduction/s3_reduce.py
```diff
@@ -232,12 +225,50 @@ def reduce(eventlabel, ecf_path=None, s2_meta=None, input_meta=None):
...
+                    os.environ['CRDS_SERVER_URL'] = \
+                        'https://hst-crds.stsci.edu/'
+
+                # If a specific CRDS context is entered in the ECF, apply it.
+                # Otherwise, log and fix the default CRDS context to make sure
+                # it doesn't change between different segments.
+                if not hasattr(meta, 'pmap') or meta.pmap is None:
+                    # Get just the numerical value
+                    meta.pmap = crds.get_context_name('hst')[4:-5]
+                os.environ['CRDS_CONTEXT'] = f'hst_{meta.pmap}.pmap'
```
kevin218, line 242: By having `meta.pmap = hst_1234`, I think we can avoid repeating some code here and on Lines 254-257.  Thoughts?

kevin218 (commented):
What's the motivation behind muting the Eureka! version in the output?  Are you worried about information overload on the screen or is there something else?

taylorbell57:
> What's the motivation behind muting the Eureka! version in the output? Are you worried about information overload on the screen or is there something else?

Yeah, I feel like that's not really something you want to know while running the code but rather is something you'd want to know a month or few later when you can't quite remember what you did

docs/source/ecf.rst
taylorbell57: So I'm currently working on tweaks to automatically add the instrument to that pmap name. Generally speaking this works well, but in the case where you work on two observatories in the same python instance then there appears to be some kind of bug with the crds package which doesn't allow you to switch (it warns me that I'm using hst for jwst data); this would be fairly rare in the real world, but happens every time we run the pytest suite... I've made some progress on this, but just need to figure out one more tweak

taylorbell57: Since it isn't just a matter of passing in a different pmap and we need to do some more annoying updates to crds variables, I think it's worth just leaving it in the way it's currently setup

kevin218 (approved):

Issue #538, opened 2023-06-11
[Enhancement]: Use single median frame for all segments


kevin218:
### Instrument

NIRSpec (Stages 1-3), NIRCam (Stages 1-3), MIRI (Stages 1-3)

### What is your suggestion?

When processing Stage 3 data in batches, a median frame is computed for each batch.  This sometimes leads to flux offsets in the light curves between batches.  By default, the code should only use the median frame from the first batch.  Users should provide an optional argument if they want a median frame computed for each batch.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

taylorbell57:
@kevin218 I'm adding this right now. Any preferences on what we call the optional ECF parameter that overrides the new default behaviour and computes a seperate median frame for each batch?

kevin218:
> @kevin218 I'm adding this right now. Any preferences on what we call the optional ECF parameter that overrides the new default behaviour and computes a seperate median frame for each batch?

Nothing comes to mind at the moment. Feel free to suggest something here.

taylorbell57:
What about `meta.independent_batches` or something like that, set to `False` by default?

PR #582, opened 2023-11-16
Default to only compute median frame for 1st batch

taylorbell57:
Before this PR, a different median frame was computed for each batch. For those like me who always only a single huge batch, this bug was absent. However, if you used more than one batch (e.g. had too little RAM to load all segments at once), then sometimes things like the computed centroid or curvature correction would change between different segments which would cause jump discontinuities between segments

This partially resolves Issue #538, although I still need to add an override option to the ECF

### DIFF
--- a/docs/source/ecf.rst
+++ b/docs/source/ecf.rst
@@ -287,6 +287,10 @@ max_memory
 ''''''''''
 Sets the maximum memory fraction (0--1) that should be used by the loaded in data files. This will reduce nfiles if needed. Note that more RAM than this may be used during operations like sigma clipping, so you're best off setting max_memory <= 0.5.
 
+indep_batches
+'''''''''''''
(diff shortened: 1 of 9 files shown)


docs/source/ecf.rst
```diff
@@ -287,6 +287,10 @@ max_memory
 ''''''''''
 Sets the maximum memory fraction (0--1) that should be used by the loaded in data files. This will reduce nfiles if needed. Note that more RAM than this may be used during operations like sigma clipping, so you're best off setting max_memory <= 0.5.
 
+indep_batches
+'''''''''''''
+Do you want to independently treat each batch of files? Strongly recommended to leave this as False unless you have a clear reason to set it to True. If set to True, you may end up with jump discontinuities between batches.
```
kevin218, line 292: We could add more context of what happens when `False`.  Something like: "When False, the median spectrum from the first batch is applied too all batches."

kevin218 (approved):

PR #585, opened 2023-11-20
Calibrated MIRI spectra

kevin218:
Generating calibrated MIRI spectra was super easy... just a few lines.


taylorbell57 (approved):
Thanks for doing this! I did the same thinga while ago for a paper and guess I forgot about it and never made a PR

taylorbell57:
Gah, but for some reason the pytest failed which really can't be related to your changes... Let me see what I can do - probably just need to set an upper limit or something

taylorbell57:
Looks to me like it is just that your flexo branch is based on an old version of the main branch, and I think merging in the updates from main should let the tests pass without any new edits to the requirements

Issue #588, opened 2023-12-02
[Bug]: Errors in Stage 5 using LD coefficients from a file


Witchblade101:
### FAQ check

- [X] Yes, I checked the FAQ and my question has not been addressed.

### Instrument

Light curve fitting (Stages 4-6)

### What happened?

I tried to run Stage 5 using a file of pre-calculated limb-darkening coefficients. I wasn't sure in that case what to put for u1 and u2 in the .epf file, so I tried 2 different approaches: commenting out the u1, u2 entries or then setting them to dummy values. 

I got errors in both cases. The first looks like it's just the wrong approach, but then the second looks like a bug.

### Error traceback output

WARNING: There are multiple metadata save files in the folder:
"/Users/dlong/DataAnalysis/JWST/eureka/t1e/Obs1/Stage4/"
Using the metadata file:
/Users/dlong/DataAnalysis/JWST/eureka/t1e/Obs1/Stage4/S4_2023-12-01_t1e_1_run1/ap6_bg7/S4_t1e_1_Meta_Save.dat
and will consider aperture ranges listed there. If this metadata file is not a part
of the run you intended, please provide a more precise folder for the metadata file.
Finished loading parameters from /Users/dlong/DataAnalysis/JWST/eureka/t1e/Obs1/Stage4/S4_2023-12-01_t1e_1_run1/ap6_bg7//S4_t1e_1_ap6_bg7_LCData.h5

Starting Stage 5: Light Curve Fitting

Input directory: /Users/dlong/DataAnalysis/JWST/eureka/t1e/Obs1/Stage4/S4_2023-12-01_t1e_1_run1/ap6_bg7/
Output directory: /Users/dlong/DataAnalysis/JWST/eureka/t1e/Obs1/Stage5/S5_2023-12-01_t1e_1_run3/ap6_bg7/
Copying S5 control file
Copying S5 parameter control file

Starting Fit of White-light Light Curve

Using the following limb-darkening values:
=========================
Starting lsq fit.
2023-12-01 21:04:23,054 - stpipe - WARNING - /Users/dlong/Eureka/src/eureka/S5_lightcurve_fitting/likelihood.py:134: RuntimeWarning: invalid value encountered in double_scalars
  lnprior_prob -= (0.5*(np.sum(((theta[i]-prior1[i])/prior2[i])**2

[...]

PR #589, opened 2023-12-05
Addresses Errors in Stage 5 using LD coefficients from a file

kevin218:
This PR primarily address Issue #588, where using a fixed LD file causes an error in BatmanModels.py.  The fix was simple enough and appears to work under the small number of tests that I ran.

I'm also including an assortment of minor improvements in this PR.  The most consequential of these changes is enabling the user to manually specify the source position in the ECF.


src/eureka/S3_data_reduction/s3_reduce.py
```diff
@@ -696,7 +696,11 @@ def reduce(eventlabel, ecf_path=None, s2_meta=None, input_meta=None):
             else:
                 normspec = util.normalize_spectrum(meta, data.aplev.values)
                 meta.mad_s3 = util.get_mad_1d(normspec)
-            log.writelog(f"Stage 3 MAD = {int(np.round(meta.mad_s3))} ppm")
+            try:
+                log.writelog(f"Stage 3 MAD = {int(np.round(meta.mad_s3))} ppm")
+            except:
+                log.writelog("Could not compute Stage 3 MAD")
```
taylorbell57, line 702: Might be worth adding WARNING at the start of this line since a NaN MAD value isn't a good sign

src/eureka/S3_data_reduction/source_pos.py
```diff
@@ -162,9 +162,12 @@ def source_pos(flux, meta, shdr, m, n, plot=True, guess=None):
         src_ypos, src_ywidth = source_pos_gauss(flux, meta, m, n, plot)
     elif meta.src_pos_type == 'hst':
         src_ypos = guess
-    else:
+    elif meta.src_pos_type == 'max':
         # brightest row for source location
         src_ypos = source_pos_median(flux, meta, m, n, plot)
+    else:
```
taylorbell57, line 168: We should make this an `elif` to do a type check here (e.g. `isnumeric(meta.src_pos_type)`) and have an `else` statement that throws an exception or something

taylorbell57 (changes requested):

src/eureka/S3_data_reduction/source_pos.py
```diff
@@ -162,9 +164,20 @@ def source_pos(flux, meta, shdr, m, n, plot=True, guess=None):
         src_ypos, src_ywidth = source_pos_gauss(flux, meta, m, n, plot)
     elif meta.src_pos_type == 'hst':
         src_ypos = guess
-    else:
+    elif meta.src_pos_type == 'max':
         # brightest row for source location
         src_ypos = source_pos_median(flux, meta, m, n, plot)
+    elif not isinstance(meta.src_pos_type, str):
```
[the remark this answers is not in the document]
kevin218: No, the method doesn't work like that.  You're probably thinking of `meta.src_pos_type.isnumeric()`, but that only works when `meta.src_pos_type` is a string.  In our case, it can be a string or an integer (or a float, I guess), so that solution wouldn't work.

kevin218:
@taylorbell57 Several people need this fix asap because they are working towards an impending deadline.  It would be good to get this merged today.  Thanks!

taylorbell57 (approved):

Issue #593, opened 2023-12-18
[Enhancement]: Allow user to supply custom stellar model for limb darkening in Stage 4


erinmmay:
### Instrument

Light curve fitting (Stages 4-6)

### What is your suggestion?

Add input to ecf and necessary adjustments so user can supply custom stellar model rather than the 1D or 3D grids already provided by exotic-ld. Exotic-ld already allows this, just needs to be added as an option in Eureka. I will open a PR with my edits to do this.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #598, opened 2023-12-20
Allowing for custom stellar intensity files to be fed into S4 for limb darkening calculations

erinmmay:
Addresses Issue/Enchanment #593. Allows for custom stellar intensity profiles which is already supported by exotic-ld, simply adds the ability to read them in and feed to exotic-ld. Adds one input to the S4 ecf (and an extra setting option to another input). I've included an example of how the file should be formatted similar to the custom throughput example file.


demos/JWST/S4_template.ecf
```diff
@@ -40,7 +40,8 @@ metallicity     0.1     # Metallicity of the star
 teff            6000    # Effective temperature of the star in K
 logg            4.0     # Surface gravity in log g
 exotic_ld_direc /home/User/exotic-ld_data/ # Directory for ancillary files for exotic-ld, download from: https://zenodo.org/record/6344946
-exotic_ld_grid  3D      # 1D or 3D model grid
+exotic_ld_grid  3D      # 1D or 3D model grid, or custom file
+custom_si_grid  /home/User/path/to/custom/stellar/intensity/profile  #Custom Stellar Intensity profile. for examples see Eureka/demos/JWST/Custom_Stellar_Intensity_Files
```
taylorbell57, line 44: Please note these changes on the documentation pages

src/eureka/S4_generate_lightcurves/generate_LD.py
```diff
@@ -81,8 +81,22 @@ def exotic_ld(meta, spec, log, white=False):
...
     # compute stellar limb darkening model
-    sld = StellarLimbDarkening(meta.metallicity, meta.teff, meta.logg,
-                               meta.exotic_ld_grid, meta.exotic_ld_direc)
+    if hasattr(meta, "custom_si_grid") and meta.exotic_ld_grid == 'custom':
+        s_wvs = (np.genfromtxt(meta.custom_si_grid, 
+                               skip_header=2, usecols=[0]).T)*1e4
+        s_mus = np.flip(np.genfromtxt(meta.custom_si_grid, 
+                                      skip_header=1, max_rows=1))
+        custom_si = np.flip(np.genfromtxt(meta.custom_si_grid, 
+                                          skip_header=2)[:, 1:], axis=1)
```
taylorbell57, line 90: Why load the file into memory three separate times? Would seem much more efficient to load it once. There is also some nuanced processing here that is not easily understandable to me at first glance, which is going to make the code hard to maintain. Can you add some comments here to explain what is happening? What is the expected formatting of the file (should also be noted in the docs), why do you sometimes use `np.flip` but not always, why do you sometimes set `skip_header` to 1 and other times 2, etc.

taylorbell57 (changes requested):

src/eureka/S4_generate_lightcurves/generate_LD.py
erinmmay: It's read in three times because you're grabbing different parts of the file each time - but I can read it in once and just trim out the necessary parts for each input if you'd prefer. The file is a table with a 1 line header, then the first column is the wavelengths and the first row is the mus. Exotic requires arrays of wavelength and mu separate from the 2D array of intensities. So when grabbing the mus, you have to skip the one row header and only read the first row, but for the wavelengths and intensities you skip the header and the row of the mu values.  And mu has to be decreasing for exotic, whereas intensity files are typically in increasing mu (hence the flip for those). I'll add some notes into here on what is being done.

docs/source/ecf.rst
```diff
@@ -767,12 +767,16 @@ Used by exotic-ld if compute_ld=True. The fully qualified path to the directory
 
 exotic_ld_grid
 ''''''''''''''
-Used by exotic-ld if compute_ld=True. 1D or 3D model grid.
+Used by exotic-ld if compute_ld=True. 1D or 3D model grid, or "custom" for a custom stellar intensity grid.
```
taylorbell57, line 770: @kevin218, just an FYI that we'll need to be careful when merging the code from this PR together your #609 later since I think both PRs change the use of this parameter

docs/source/ecf.rst
```diff
@@ -767,12 +767,16 @@ Used by exotic-ld if compute_ld=True. The fully qualified path to the directory
...
 
 exotic_ld_file
 ''''''''''''''
 Used by exotic-ld as throughput input file. If none, exotic-ld uses throughput from ancillary files. Make sure that wavelength is given in Angstrom!
 
+custom_si_grid
+''''''''''''''
+if exotic_ld_grid = custom, supply the path to your stellar intensity grid here
```
taylorbell57, line 778: Capitalize the first letter of `if`, and specify that it needs to be the fully qualified path (to clarify that it isn't with respect to `topdir`)

taylorbell57 (changes requested):
Just a tiny tweak needed to the docs - I'll just make that myself and then will approve the PR

src/eureka/S4_generate_lightcurves/generate_LD.py
taylorbell57: Reading it in once would speed things up (since I/O tends to be one of the slowest parts of programs), but its not worth slowing down this PR for since your code works and its not like this is done often and the files aren't huge. Thanks for the description and the added comments in the code

taylorbell57 (approved):

Issue #570, opened 2023-10-25
[Bug]: x-axes inconsistent on Stage 3 output plots


arikaegan:
### FAQ check

- [X] Yes, I checked the FAQ and my question has not been addressed.

### Instrument

MIRI (Stages 1-3)

### What happened?

The x-axes are inconsistent between the 2D Background Subtracted Frame( e.g. fig3301_file0_int00000_ImageAndBackground.png)  plots and the 1D Spectrum plots (e.g. fig3302_file0_int00000_Spectrum.png). It seems like the 1D spectrum plots have the Detector Pixel Position flipped.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

taylorbell57:
Ah indeed, the x-axis for Fig 3301 is labelled backwards for MIRI observations only, and for Fig 3302 it would probably be best to invert the x-axis so that short wavelengths are on the left. I'll aim to fix this minor bug in a week or so

taylorbell57:
Alright, looking more carefully it looks like even more plots are messed up for MIRI. Specifically, figs 3301, 3303?, 3308, and 3506 have the detector pixels labelled backwards, and fig 3302 should be flipped so that wavelength increases toward the right and not pixel position. It's also not immediately clear to me if the column numbering in figs 3501 are in the correct for MIRI. While I'm at it, fig 3304 appears not to be using `bbox_inches='tight'` or something since parts of the plot are getting cut off in the saved image.

I'll work on this today and hopefully submit a PR.

taylorbell57:
PR #600 resolved the issues with figures 3301, 3302, 3308, and the bug with 3304 doesn't seem to occur anymore. That only leaves figs 3303 and 3506 which aren't wrong but just use different units (relative pixel position rather than detector pixel position). This seems fine to me, so I'm going to close this issue now

PR #600, opened 2023-12-23
Assorted HST and JWST enhancements

kevin218:
This PR implements a few, unrelated enhancements based on my needs while working on several projects over the past week.  Fitting LCs to HST WFC3 data is now less sensitive to HST's period because, by adding h6, the reset now happens during Earth occultation instead of at the time of the first science frame.  This PR also solves an issue reported in #570 in a slightly different manner than @taylorbell57's solution because I wanted to get the half-pixel edges correct (bounds for detector pixels 5-28 need to be 4.5-28.5).  Here's a brief list of fixes:
- Enable HST FLT reduction
- Enhance HSTRampModel
- Update WFC3 template
- Implement `plots_s3.get_bounds()` to correctly define xlim and ylim while accounting for instrument and half-pixels offsets
- Enable calibrated spectra with NIRSpec
- Small bug fixes

### DIFF
--- a/demos/HST/s5_fit_par_wfc3_template.epf
+++ b/demos/HST/s5_fit_par_wfc3_template.epf
@@ -50,7 +50,9 @@ u2           0.1           'free'         0            1            U
 c0           1             'free'         0.95         1.05         U
 h0           0             'free'         -0.1         0.1          U
 h1           10            'free'         0            100          U
-h3           0             'free'         -1           1            U
+h3           0.0           'fixed'        -1           1            U
+h5           6.6422e-02    'fixed'        6.6422e-02   0.001        N
+h6           0.03          'fixed'        0.03         0.0          N
 # -----------
 # ** White noise **
 # Use scatter_mult to fit a multiplier to the expected noise level from Stage 3 (recommended)
--- a/setup.cfg
+++ b/setup.cfg
@@ -57,7 +57,7 @@ where = src
 
 [options.extras_require]
 jwst =
-    jwst==1.11.4
+    jwst==1.12.2
     stcal>=1.0.0 # Lower limit needed for our create_integration_model function
 hst =
     image_registration@git+https://github.com/keflavich/image_registration@master # Need GitHub version to avoid np.float issue
--- a/setup.cfg
+++ b/setup.cfg
@@ -74,6 +74,8 @@ test =
     pytest-cov
     pytest-doctestplus
 pymc3 =
+    jwst==1.11.4
+    stcal>=1.0.0 # Lower limit needed for our create_integration_model function
     exoplanet
     # mkl-service # Needed for theano - only available by conda install mkl-service
     pymc3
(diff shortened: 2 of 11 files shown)


src/eureka/S3_data_reduction/s3_reduce.py
```diff
@@ -292,11 +292,6 @@ def reduce(eventlabel, ecf_path=None, s2_meta=None, input_meta=None):
                 meta.max_memory = 0.5
             if not hasattr(meta, 'nfiles'):
                 meta.nfiles = 1
-            if meta.nfiles == 1 and meta.nfiles > 1:
-                log.writelog('WARNING: Strange behaviors can occur if you set '
-                             'nfiles to 1. If your computer has enough RAM to '
-                             'load many/all of your Stage 2 files, it is '
-                             'strongly recommended to increase nfiles.')
```
taylorbell57, line 299: What's the mentality for removing this? You figure all the issues with only using one file were solved by the recent option to avoid independent batches? I'd assume performance would still be significantly better with more than one file since that allows more averaging. Or is the issue that HST round-trip scanning wasn't working with larger batches or something?

demos/HST/s5_fit_par_wfc3_template.epf
```diff
@@ -50,7 +50,9 @@ u2           0.1           'free'         0            1            U
 c0           1             'free'         0.95         1.05         U
 h0           0             'free'         -0.1         0.1          U
 h1           10            'free'         0            100          U
-h3           0             'free'         -1           1            U
+h3           0.0           'fixed'        -1           1            U
+h5           6.6422e-02    'fixed'        6.6422e-02   0.001        N
+h6           0.03          'fixed'        0.03         0.0          N
```
taylorbell57, line 55: Need to document the new parameter on the docs

setup.cfg
```diff
@@ -74,6 +74,8 @@ test =
     pytest-cov
     pytest-doctestplus
 pymc3 =
+    jwst==1.11.4
+    stcal>=1.0.0 # Lower limit needed for our create_integration_model function
```
taylorbell57, line 78: This implementation is probably going to cause us at least some headaches, especially since the user isn't going to know that the jwst version will change with different installation options. And also, what happens if you put both `jwst` and `pymc3` in the optional installs, does the `jwst` version that gets installed depend on the order of `jwst` and `pymc3` in the install command (headaches will ensue if so)

src/eureka/S3_data_reduction/s3_reduce.py
kevin218: Yes, it seemed like the warning is no longer necessary now that you've implemented the recent fix.  Maybe the better solution would be to change the warning to encourage users to set `indep_batches = False` in the S3 meta (after checking to see if it is set to True).

setup.cfg
kevin218: I didn't try installing with both, good idea.  I assume it'll break since both version requirements cannot be met.

I'd rather deal with a minor headache than with broken code.  After a discussion with STScI's help desk, I learned that some calibration files changed their formatting in 1.11, but the pipeline wasn't made aware of this change for TSO data until 1.12.

I actually think this is the temporary solution that we've bene looking for.  We can freeze the JWST version for `pymc3` users but continue taking advantage of enhancements for other users.  I'm open to relabelling `pymc3` to something else (like 'jwst+pymc3') and updating the documentation to make it more obvious to the users.

kevin218:
@taylorbell57 Good to go whenever you get a chance.

taylorbell57:
Sorry for the slow response, just reviewed the PR again and it was all good

PR #616, opened 2024-02-08
Updating instructions for newer exotic-ld versions

### LINKED ISSUE #540: [Bug]: Current code/documentation incompatible with exotic-ld v3
### FAQ check

### Instrument

### What happened?

I just did a fresh install and I now if I ask for 3D LD models in S4, it says I'm missing stagger grid models. We should set a cap on the exotic-ld version and/or update our documentation to make it very clear that you need to have the reference files related to your installed exotic-ld version (from zenodo). We may need to update the docs to clarify which exotic-ld grids are available based on the newer version.

### Code of Conduct

taylorbell57:
This PR resolves #540

I've set a lower limit on the exotic-ld version to 3.1.2 which is their latest version (contains the updated post commissioning sensitivity files for JWST), and I've updated the documentation to make sure folks know how to download those reference files and know that the reference files change with different ExoTiC-LD reference files.

### DIFF
--- a/demos/HST/S4_wfc3_template.ecf
+++ b/demos/HST/S4_wfc3_template.ecf
@@ -40,7 +40,7 @@ metallicity     0.1     # Metallicity of the star
 teff            6000    # Effective temperature of the star in K
 logg            4.0     # Surface gravity in log g
 exotic_ld_direc /home/User/exotic-ld_data/ # Directory for ancillary files for exotic-ld, download from: https://zenodo.org/record/6344946
-exotic_ld_grid  3D      # 1D or 3D model grid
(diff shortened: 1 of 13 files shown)


kevin218 (dismissed):

taylorbell57:
Just realized this is failing because the ExoTiC-LD code (on pip and GitHub) only has version 3.0.0 while the reference files are on 3.1.2. I've sent David and Hannah an email to make sure we're supposed to use those two together; Hannah is out of the office for a while, so hopefully David responds fairly soon. Either way, I'll have to tweak some of the files and wording to make that clearer in these instructions.

kevin218 (approved):

Issue #621, opened 2024-02-20
[Bug]: emcee old_chain doesn't work for white-light curve fitting


evamariaa:
### FAQ check

- [X] Yes, I checked the FAQ and my question has not been addressed.

### Instrument

Light curve fitting (Stages 4-6)

### What happened?

When using `old_chain` in the Stage 5 light curve fitting settings, the function `start_from_oldchain_emcee` in `fitters.py` looks for `ch0` instead of `white` as `channel_key` to find the previous emcee run.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #622, opened 2024-02-20
Fix emcee old_chain for white-light curve fit bug

evamariaa:
Quick fix of issue #621 so that the emcee run can also continue from a white-light emcee chain run.
I used the same nomenclature (_channel_tag_) as in the `save_fit` function.


taylorbell57 (approved):

taylorbell57:
Thanks for finding and patching this!

PR #624, opened 2024-02-22
Adding a PyMC3 compatible GP using celerite2

taylorbell57:
This PR adds a GP to the PyMC3 models which is possible thanks to the recent switch to celerite2 over celerite. Celerite2 will also eventually allow us to use jax or PyMC>3 code if/when we implement it. I've been using this in production for a while now and it has been working great!


src/eureka/S5_lightcurve_fitting/differentiable_models/GPModel.py
```diff
@@ -0,0 +1,278 @@
...
+            The requested kernel.
+        """
+        # get metric and amplitude for the current kernel and channel
+        amp = lib.exp(coeffs[c][k][0])
+        ls = lib.exp(coeffs[c][k][1])
+
+        if kernel_name == 'Matern32':
+            kernel = celerite2.terms.Matern32Term(sigma=1, rho=ls)
+        else:
+            raise AssertionError('Our celerite2 implementation currently only '
```
kevin218, line 272: Should we consider implementing a few other kernels in this PR?

taylorbell57: The small number of available celerite2 kernels are listed here: https://celerite2.readthedocs.io/en/latest/api/python/#python-recommended

I don't expect that the SHO kernel would be overly useful for our kind of data

kevin218: Ok, thanks!

PR #625, opened 2024-02-22
Ms no longer used in starry models; batman `w` parameter no long must be between (0,360); minor tweaks

taylorbell57:
The requirement for Ms was always a nuisance since it could potentially crash the code if Ms was larger than allowed by Kepler's 3rd law. The only reason it was added was so that starry would be able to accept both a/Rs and Porb, but Everett Schlawin pointed out to me that we can just set Mp=0 and then solve for Ms which is way easier and removes that nuisance parameter.

Some folks allow negative omega values, and sometimes its useful to use a prior like Intial Guess +/- 180 so that your fits don't end up bi-modal if they're near 0 or 360.

Finally, sometimes curve_fit will fail to converge with the default maxfev value, so I just increased it by 10x which solved my problems. Might be worth changing this to scipy.optimize.minimize and using Powell instead since that's been working so well for us elsewhere.

### DIFF
--- a/docs/source/ecf.rst
+++ b/docs/source/ecf.rst
@@ -1038,13 +1038,8 @@ This file describes the transit/eclipse and systematics parameters and their pri
 
          This parameter is recommended for batman_ecl fits as it allows for a conversion of a/R* to physical units in order to account for light travel time.
          If not provided for batman_ecl fits, the finite speed of light will not be accounted for.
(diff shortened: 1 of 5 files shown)


kevin218 (changes requested):
Check the flake8 issue

kevin218 (approved):

Issue #605, opened 2024-01-24
[Enhancement]: Integrate Fleck for starspot modelling 


kevin218:
### Instrument

Light curve fitting (Stages 4-6)

### What is your suggestion?

We should add [Fleck](https://fleck.readthedocs.io/en/stable/index.html) to the LC fitting stage so that users can model starspots.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #627, opened 2024-02-27
Enable modeling of starspots with both Fleck and Starry

erinmmay:
Exactly what the title says - Adds fleck for starspot modeling, modifies starry code to allow starspots (not recommended, very slow...).  Adds a plot to show starspots for fleck. This addresses #605


taylorbell57:
@erinmmay, once this PR is ready for review, please mark it as such near the bottom of the page and request a review from Kevin and/or I.

erinmmay:
Resolved conflicts and set fleck's "fast" option as the default (assumes stellar rotation is much longer than transit time and spots are stationary. user still has option to supply stellar rotation and calculate in "slow" mode). Pulling down the new version of the branch with conflicts resolved to test and then will remove "draft" status so it can be reviewed

erinmmay:
I'm removing the draft status from this to expedite the process so it can be reviewed now, but I still have a couple tests to run (e.g. making sure this works with dynesty) before it should be merged

src/eureka/S5_lightcurve_fitting/models/FleckModel.py
```diff
@@ -0,0 +1,217 @@
...
+                if len(index) != 0:
+                    item = self.longparamlist[chan][index[0]]
+                    uarray.append(self.parameters.dict[item][0])
+            bm_params.u = uarray
+
+            # Enforce physicality to avoid crashes from batman by returning
+            # something that should be a horrible fit
+            if not ((0 < bm_params.per) and (0 < bm_params.inc < 90) and
+                    (1 < bm_params.a) and (0 <= bm_params.ecc < 1) and
+                    (0 <= bm_params.w <= 360)):
```
kevin218, line 177: We recently removed the constraint on omega (see BatmanModels.py).  You should check for other differences in BatmanModels.py

kevin218 (changes requested):
Here's is some preliminary feedback on this PR.

erinmmay:
addressed comments and updated FleckModel.py to use new PlanetParams class. Fully tested with emcee and dynesty.

demos/JWST/S5_fit_par_template.epf
```diff
@@ -70,6 +70,25 @@ u2          0.84            'free'          0           1           U
...
+# spotrot      10       'fixed'		      10		    1		    N
+# spotnpts     100      'independent
+# spotcon0     0.9			'free'			    0		      1		    U
+# spotrad0     0.2			'free'          0    	    360     U
+# spotlat0     0			  'free'          -90    	  90      U
+# spotlon0     0			  'free'          -180    	180     U
+# spotcon1     0.1			'free'			    0		      1		    U
+# spotrad1	   0.2		  'free'          0    	    1       U
+# spotlat1	   0		    'free'          -90    	  90      U
+# spotlon1	   0			  'free'          -180    	180     U
```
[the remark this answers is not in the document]
erinmmay: sigh I addressed this but they're still not aligned. one moment.

docs/source/ecf.rst
```diff
@@ -1117,6 +1117,19 @@ This file describes the transit/eclipse and systematics parameters and their pri
       - ``osc_t1`` - This offset determined the start phase of the sinusoidal function.
 
          The time-dependent amplitude and period are defined as follows: ``amp = osc_amp * np.exp(-osc_amp_decay * (time - osc_t0))`` and ``per = osc_per * np.exp(-osc_per_decay * (time - osc_t0))``.  The damped oscillator model is then defined as: ``osc = 1 + amp * np.sin(2 * np.pi * (time - osc_t1) / per)``.  Note that ``osc[time < osc_t0] = 1``.
+         
+   - Star Spot Parameters
+      - ``spotstari`` - The stellar inclination in degrees.
```
taylorbell57, line 1122: Does this parameter only matter if `spotrot` >> T_14? It would probably be helpful to add a small word of guidance to users on how they might set priors on this parameter or when they might choose to fix it to 90 degrees or something.

demos/JWST/S5_fit_par_template.epf
```diff
@@ -70,6 +70,25 @@ u2          0.84            'free'          0           1           U
...
 # osc_t1          0.3163  'free'          0.3          0.4            U
 
+# ------------------
+# ** Star Spot parameters **
+# each spot needs rad, lat, lon. see docs for details!
+# for fleck: can only have one spot contrast (spotcon0), spotnpts is the number of temporal points to evaluate at, spotrad is relative to star
+# for starry: assign one spotcon per spot (spotcon0, spotcon1, etc), spotres is the ydeg of the star map, spotrad is in degrees
+# NOTE: fleck contrast = 1 - starry contrast
+# ------------------
+# spotstari    90       'fixed'         90        1       N
```
taylorbell57, line 80: While this parameter is set to fixed in the template, the demonstrated prior (should users choose to set the parameter to free) is 90 +/- 1 degree which seems excessively constraining. Wouldn't a prior of U(0,90) or something be more fitting if someone chose to fit for this parameter?

demos/JWST/S5_fit_par_template.epf
```diff
@@ -70,6 +70,25 @@ u2          0.84            'free'          0           1           U
...
+# spotrot      10       'fixed'         10        1       N
+# spotnpts     100      'independent
+# spotcon0     0.9      'free'          0         1       U
+# spotrad0     0.2      'free'          0         360     U
+# spotlat0     0        'free'          -90       90      U
+# spotlon0     0        'free'          -180      180     U
+# spotcon1     0.1      'free'          0         1       U
```
taylorbell57, line 87: Probably worth adding a comment at the end of this line to remind users that you should only add a `spotcon1` parameter if you were using starry to model two spots since fleck only permits one spot contrast for all spots.

demos/JWST/S5_fit_par_template.epf
```diff
@@ -70,6 +70,25 @@ u2          0.84            'free'          0           1           U
...
+# ------------------
+# ** Star Spot parameters **
+# each spot needs rad, lat, lon. see docs for details!
+# for fleck: can only have one spot contrast (spotcon0), spotnpts is the number of temporal points to evaluate at, spotrad is relative to star
+# for starry: assign one spotcon per spot (spotcon0, spotcon1, etc), spotres is the ydeg of the star map, spotrad is in degrees
+# NOTE: fleck contrast = 1 - starry contrast
+# ------------------
+# spotstari    90       'fixed'         90        1       N
+# spotrot      10       'fixed'         10        1       N
+# spotnpts     100      'independent
```
taylorbell57, line 82: In your notes in `ecf.rst`, you mention that this should be ~200-500 for fleck - I recommend you stick with that for this template

demos/JWST/S5_fit_par_template.epf
```diff
@@ -70,6 +70,25 @@ u2          0.84            'free'          0           1           U
 # osc_t0          0.3163  'fixed'         0            1              U
 # osc_t1          0.3163  'free'          0.3          0.4            U
 
+# ------------------
+# ** Star Spot parameters **
```
taylorbell57, line 74: Please also add these changes to the `docs/media/S5_fit_par_template.epf` file which is shown on the https://eurekadocs.readthedocs.io/en/latest/ecf.html page

src/eureka/S5_lightcurve_fitting/differentiable_models/StarryModel.py
```diff
@@ -132,9 +132,43 @@ def setup(self):
...
+                        elif 'lon' in key:
+                            spotlon = np.append(spotlon, 
+                                                self.parameters.dict[key][0])
+                        elif 'con' in key:
+                            spotcon = np.append(spotcon, 
+                                                self.parameters.dict[key][0])
+                        elif 'rot' in key:
+                            starrot = self.parameters.dict[key][0]
+                        else:
+                            spotres = self.parameters.dict[key][0]
```
taylorbell57, line 158: This code is going to break for `mult_white` or shared fits. I recommend you make use of the `temp` object I defined above which should only include the parameters for the current channel

src/eureka/S5_lightcurve_fitting/differentiable_models/StarryModel.py
```diff
@@ -313,9 +347,43 @@ def update(self, newparams, **kwargs):
...
```
taylorbell57, line 373: See my comment above about the use of the `temp` object to make sure this code works for shared fits and `mult_white` fits

src/eureka/S5_lightcurve_fitting/models/FleckModel.py
```diff
@@ -0,0 +1,328 @@
...
+import inspect
+
+import fleck
+
+from .Model import Model
+from ..limb_darkening_fit import ld_profile
+from ...lib.split_channels import split
+
+
+class PlanetParams():
```
taylorbell57, line 12: Rather than copying this class, just import it from `BatmanModels` as was done in `PoetModels.py`

src/eureka/S5_lightcurve_fitting/models/FleckModel.py
```diff
@@ -0,0 +1,328 @@
...
+                fleck_times = np.linspace(time[0], time[-1], npoints)
+                fleck_transit = star.light_curve(
+                    self.spotlon[chan][:nspots, None]*unit.deg, 
+                    self.spotlat[chan][:nspots, None]*unit.deg, 
+                    self.spotrad[chan][:nspots, None],
+                    star_inc*unit.deg, 
+                    planet=pl_params, times=fleck_times).flatten()
+                m_transit = np.interp(time, fleck_times, fleck_transit)
+                
+                light_curve *= m_transit/m_transit[0]
```
taylorbell57, line 322: The choice to normalize by the first integration rather than the mean or something else seems non-trivial to me. I guess it probably doesn't matter much though since that should presumably be taken care of by the `c0` parameter.

taylorbell57 (changes requested):
If the fleck model isn't super slow, could you also add it to either the NIRCam or NIRSpec test by adding the model to the relevant S5 ECF inside `tests/NIR*_ecfs/` and adding the relevant parameters to the corresponding EPF file?

src/eureka/S5_lightcurve_fitting/plots_s5.py
```diff
@@ -794,3 +797,51 @@ def plot_GP_components(lc, model, meta, fitter, isTitle=True):
...
+            elif 'rot' in attr:
+                star_rotation = model.parameters.dict[attr][0]
+                fleck_fast = True
+            elif 'stari' in attr:
+                star_inc = model.parameters.dict[attr][0]
+        elif attr.startswith('u'):
+            uarray = np.append(uarray, model.parameters.dict[attr][0])
+        else:
+            setattr(bm_params, attr, model.parameters.dict[attr][0])
+    bm_params.u = uarray
```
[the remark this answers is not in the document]
erinmmay: I'm not sure I would recommend joint fits with star spots at this point. I can add a check at the beginning for now if a multwhite or shared fit is attempted with fleck to warn that it's not currently supported?

src/eureka/S5_lightcurve_fitting/plots_s5.py
```diff
@@ -794,3 +797,51 @@ def plot_GP_components(lc, model, meta, fitter, isTitle=True):
         fig.savefig(meta.outputdir+fname, bbox_inches='tight', dpi=300)
         if not meta.hide_plots:
             plt.pause(0.2)
+
+
+def plot_fleck_star(lc, model, meta, fitter):
```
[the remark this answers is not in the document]
erinmmay: starry is prohibitively slow with star spots right now, it's technically enabled but practically useless. I can add the plots but not sure they're actually useful.

erinmmay:
re: adding fleck to tests/NIR*_ecfs/ - it is too slow in my opinion to be a test model.

src/eureka/S5_lightcurve_fitting/models/FleckModel.py
erinmmay: yeah, doesn't matter what you normalize by, just needs to be normalized because fleck changes the out of transit baseline based on the spot coverage. c0 takes care of it.

src/eureka/S5_lightcurve_fitting/plots_s5.py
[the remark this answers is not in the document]
taylorbell57: While using fleck for visits from different epochs simultaneously doesn't really make sense (or would be prohibitively computationally expensive), I can imagine someone wanting to fit NIRSpec NRS1 and NRS2 at the same time while needing to model a star spot. There should be enough examples from the other models on how to support mult_white or shared fits, and hopefully you'd be able to largely copy-paste some code from another model

src/eureka/S5_lightcurve_fitting/plots_s5.py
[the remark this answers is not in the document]
taylorbell57: If we're enabling starry spots for v1.0, I think it'd be important to include a plot for them. Hopefully it should require minimal work from you since starry has most of the plotting utilities built-in

src/eureka/S5_lightcurve_fitting/s5_fit.py
```diff
@@ -647,6 +647,25 @@ def fit_channel(meta, time, flux, chan, flux_err, eventlabel, params,
                                          nints=lc_model.nints,
                                          num_planets=meta.num_planets)
         modellist.append(t_eclipse)
+    if 'fleck_tr' in meta.run_myfuncs:
```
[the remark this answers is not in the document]
erinmmay: Looks like with the new implementation of AstroModel there's nothing I need to do to integrate the option to use fleck_tr with the PC models, I just need to add the model name to the fleck transit model so that the AstroModel code knows it's a transit

src/eureka/S5_lightcurve_fitting/models/FleckModel.py
erinmmay: Changing to BatmanModels for now so I can run local tests, but leaving a note to myself to check when resolving conflicts that I change it to import from AstroModels instead.

environment.yml
```diff
@@ -12,6 +12,7 @@ dependencies:
     - dynesty[version='>1.0'] # Lower limit needed for specific arguments
     - emcee[version='>3.0.0'] # Lower limit needed for specific arguments
     - flake8 # Needed for testing
+    - fleck
```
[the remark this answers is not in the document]
erinmmay: added

src/eureka/S5_lightcurve_fitting/plots_s5.py
```diff
@@ -794,3 +797,51 @@ def plot_GP_components(lc, model, meta, fitter, isTitle=True):
...
+    fig = plt.figure(5306, figsize=(8, 6))
+    plt.clf()
+    star = fleck.Star(spot_contrast=spot_contrast, 
+                      u_ld=bm_params.u,
+                      rotation_period=star_rotation)
+    star.plot(spotlon[:, None]*unit.deg, spotlat[:, None]*unit.deg, 
+              spotrad[:, None], star_inc*unit.deg, 
+              planet=bm_params, time=0)
+    fname = (f'figs{os.sep}fig5306_fleck_star_{fitter}')
+    fig.savefig(meta.outputdir+fname, bbox_inches='tight', dpi=300)
```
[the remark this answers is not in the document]
erinmmay: done

src/eureka/S5_lightcurve_fitting/fitters.py
```diff
@@ -162,6 +162,10 @@ def callback(theta):
     if meta.isplots_S5 >= 1:
         plots.plot_fit(lc, model, meta, fitter=calling_function)
 
+    # Plot star spots
+    if 'fleck_tr' in meta.run_myfuncs and meta.isplots_S5:
```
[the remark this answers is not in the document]
erinmmay: fixed

docs/source/ecf.rst
erinmmay: only matters when the star is rotating, so that the spots move appropriately. Will add text about this

demos/JWST/S5_fit_par_template.epf
erinmmay: This parameter only matters if the star is rotating (fleck's "slow" mode) otherwise the inclination is irrelevant because the lat/lon of spot can change to accommodate it. I don't know that people would fit for this since information from R-M measurements would be more informative than spots in transits.

demos/JWST/S5_fit_par_template.epf
erinmmay: done

demos/JWST/S5_fit_par_template.epf
erinmmay: done

demos/JWST/S5_fit_par_template.epf
erinmmay: done

src/eureka/S5_lightcurve_fitting/plots_s5.py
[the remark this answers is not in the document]
erinmmay: apparently I already did this in fleck.py and just forgot to put it in starry/plotting code. done, running tests and will request review later

src/eureka/S5_lightcurve_fitting/differentiable_models/StarryModel.py
erinmmay: apparently I already did this in fleck.py and just forgot to put it in starry/plotting code. done, running tests and will request review later

kevin218 (commented):
The new changes look good to me.  I'll let @taylorbell57 give final approval.

kevin218:
What’s going on with this PR?  It looks like I signed off on it 3 weeks ago and nothing has happened since, but now there a bunch of conflicts.

taylorbell57:
@erinmmay still needed to address several things and then was going to request a review from me once it was ready

erinmmay:
I've merged the changes in main into my starspots branch, have to go in and make a few more edits/tests. I will update and request a review when it's ready

erinmmay:
okay, all good now. ready for review.

src/eureka/S5_lightcurve_fitting/differentiable_models/StarryModel.py
```diff
@@ -114,8 +114,80 @@ def setup(self, newparams):
...
+                    if self.parameters.dict[item0][1] == 'free':
+                        item0 += channel_id
+                    value = self.parameters.dict[item0][0]
+                    spotlon[chan, n] = value
+                # read contrasts (same for all spots)
+                item0 = 'spotcon0'
+                if self.parameters.dict[item0][1] == 'free':
+                    item0 += channel_id
+                value = self.parameters.dict[item0][0]
+                spotcon[chan] = value
```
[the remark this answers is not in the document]
taylorbell57: You also setup this array to be 2D and access it below as being 2D

src/eureka/S5_lightcurve_fitting/differentiable_models/StarryModel.py
```diff
@@ -114,8 +114,80 @@ def setup(self, newparams):
...
+                # read stellar inclination
+                item0 = 'spotstari'
+                if self.parameters.dict[item0][1] == 'free':
+                    item0 += channel_id
+                value = self.parameters.dict[item0][0]
+                starinc[chan] = value
```
taylorbell57, line 169: You initialize this `starinc` array, but you never pass the values into a starry object

src/eureka/S5_lightcurve_fitting/differentiable_models/StarryModel.py
```diff
@@ -323,9 +399,81 @@ def update(self, newparams, **kwargs):
...
+                             lat=spotlat[chan, si], 
+                             lon=spotlon[chan, si])
+
+                # Initialize star object
+                star = starry.Primary(map, m=0, r=pl_params.Rs, 
+                                      prot=starrot[chan])
+            else:           
+                # Initialize star object
+                star = starry.Primary(starry.Map(udeg=self.udeg),
+                                      m=0, r=pl_params.Rs)
```
taylorbell57, line 475: Almost all of my comments for the `setup` function above also apply here (except for the small number where I'd explicitly said they don't apply here)

src/eureka/S5_lightcurve_fitting/models/FleckModel.py
```diff
@@ -0,0 +1,245 @@
...
+        else:
+            nchan = 1
+            channels = [channel, ]
+            
+        if pid is None or pid == 0:
+            pid_iter = range(self.num_planets)
+            pid_id = ''
+        else:
+            pid_iter = [pid,]
+            pid_id = f'_pl{pid}'
```
taylorbell57, line 101: `pid_id` should not be set here and should instead be set down below and updated for each `pid` in `pid_iter`. Look at the `BatmanTransitModel` as an example

src/eureka/S5_lightcurve_fitting/plots_s5.py
```diff
@@ -793,3 +797,198 @@ def plot_GP_components(lc, model, meta, fitter, isTitle=True):
...
         if not meta.hide_plots:
             plt.pause(0.2)
+
+
+def plot_fleck_star(lc, model, meta, fitter):
+    
+    # find number of spots
+    spotinds = [s for s in model.parameters.dict.keys() if 'spotrad' in s]   
+    counter = [s for s in spotinds if '_' in s]
+    nspots = len(spotinds) - len(counter)
```
taylorbell57, line 807: This bit of code (and the one below for `limbdarks`) is quite unintuitive and will be hard to maintain

src/eureka/S5_lightcurve_fitting/plots_s5.py
```diff
@@ -793,3 +797,198 @@ def plot_GP_components(lc, model, meta, fitter, isTitle=True):
...
+
+    # find number of limb darkening parametesr
+    limbdarks = [s for s in model.parameters.dict.keys() if 'u' in s] 
+    counter = [s for s in limbdarks if '_' in s]
+    nus = len(limbdarks) - len(counter)
```
taylorbell57, line 812: We already have code implemented to nicely handle limb-darkening parameters for the batman transit model - no need to do something different here

src/eureka/S5_lightcurve_fitting/plots_s5.py
```diff
@@ -793,3 +797,198 @@ def plot_GP_components(lc, model, meta, fitter, isTitle=True):
...
+    
+    spotrad = np.zeros((lc.nchannel_fitted, nspots))
+    spotlat = np.zeros((lc.nchannel_fitted, nspots))
+    spotlon = np.zeros((lc.nchannel_fitted, nspots))
+    spotcon = np.zeros((lc.nchannel_fitted))
+    starrot = np.zeros((lc.nchannel_fitted))
+    starinc = np.zeros((lc.nchannel_fitted))
```
taylorbell57, line 819: Similarly to my comments on the starry model, I don't think it makes sense to have different stellar rotation periods and inclinations for each channel, and I think there should only be a single value that's shared between all channels.

taylorbell57, line 819: See my comments on the starry model about better defaults for some of these parameters

src/eureka/S5_lightcurve_fitting/plots_s5.py
```diff
@@ -793,3 +797,198 @@ def plot_GP_components(lc, model, meta, fitter, isTitle=True):
...
+
+    uarray = np.zeros((lc.nchannel_fitted, nus))
+
+    # Set spot/fleck parameters
+    # set a star rotation for the star object
+    # not actually used in fast mode. 
+    # overwritten if user supplies and runs slow mode
+    starrot[:] = 100
```
taylorbell57, line 827: Can just do `starrot = 100*np.ones((lc.nchannel_fitted))` above instead of doing this over multiple lines

src/eureka/S5_lightcurve_fitting/models/FleckModel.py
```diff
@@ -0,0 +1,245 @@
...
+        counter = [s for s in spotinds if '_' in s]
+        nspots = len(spotinds) - len(counter)
+        
+        # Get relevant spot parameters
+        spotrad = np.zeros((self.nchannel_fitted, nspots))
+        spotlat = np.zeros((self.nchannel_fitted, nspots))
+        spotlon = np.zeros((self.nchannel_fitted, nspots))
+        spotcon = np.zeros((self.nchannel_fitted))
+        starrot = np.zeros((self.nchannel_fitted))
+        starinc = np.zeros((self.nchannel_fitted))
```
taylorbell57, line 118: Similarly to my comments on the starry model, I don't think it makes sense to have different stellar rotation periods and inclinations for each channel, and I think there should only be a single value that's shared between all channels.

taylorbell57, line 118: See my comments on the starry model about better defaults for some of these parameters

src/eureka/S5_lightcurve_fitting/plots_s5.py
```diff
@@ -793,3 +797,198 @@ def plot_GP_components(lc, model, meta, fitter, isTitle=True):
...
+
+    uarray = np.zeros((lc.nchannel_fitted, nus))
+
+    # Set spot/fleck parameters
+    # set a star rotation for the star object
+    # not actually used in fast mode. 
+    # overwritten if user supplies and runs slow mode
+    starrot[:] = 100 
+
+    pid_id = ''
```
taylorbell57, line 829: No need to make this empty string if it is never updated below (and none of the plots depend on the planet number)

src/eureka/S5_lightcurve_fitting/plots_s5.py
```diff
@@ -793,3 +797,198 @@ def plot_GP_components(lc, model, meta, fitter, isTitle=True):
...
+    for channel in lc.fitted_channels:
+
+        # Initialize planet
+        pl_params = PlanetParams(model, 0, channel)
+
+        # channel ID
+        if channel == 0:
+            channel_id = ''
+        else:
+            channel_id = f'_{channel}'
```
taylorbell57, line 840: See comment above on the change to `channel_id`

src/eureka/S5_lightcurve_fitting/plots_s5.py
```diff
@@ -793,3 +797,198 @@ def plot_GP_components(lc, model, meta, fitter, isTitle=True):
...
+        spotcon[channel] = value
+        # read stellar inclination
+        item0 = 'spotstari' + pid_id
+        if model.parameters.dict[item0][1] == 'free':
+            item0 += channel_id
+        value = model.parameters.dict[item0][0]
+        starinc[channel] = value
+        # read stellar rotation (if provided)
+        if 'spotrot' in model.parameters.dict.keys():
+            starrot[channel] = model.parameters.dict['spotrot'][0]
```
taylorbell57, line 875: Rather than having this code duplicated across the starry, fleck, and plotting codes, this should be consolidated into a function that gets re-used somewhere.

src/eureka/S5_lightcurve_fitting/differentiable_models/StarryModel.py
```diff
@@ -114,8 +114,80 @@ def setup(self, newparams):
...
+                counter = [s for s in spotinds if '_' in s]
+                nspots = len(spotinds) - len(counter)
+
+                # create arrays to hold values
+                spotrad = np.zeros((self.nchannel_fitted, nspots))
+                spotlat = np.zeros((self.nchannel_fitted, nspots))
+                spotlon = np.zeros((self.nchannel_fitted, nspots))
+                spotcon = np.zeros((self.nchannel_fitted, nspots))
+                starrot = np.zeros((self.nchannel_fitted))
+                starinc = np.zeros((self.nchannel_fitted))
```
taylorbell57, line 131: At this point in the code, the "values" are theano tensors (since you're pre-compiling things here) and you cannot set elements of a numpy float array with theano tensors. You'll need to instead use `tt.zeros`. But in the `update` function down below, the "values" are floats and you can stick with `np.zeros` there

src/eureka/S5_lightcurve_fitting/differentiable_models/StarryModel.py
```diff
@@ -114,8 +114,80 @@ def setup(self, newparams):
...
+                item0 = 'spotstari'
+                if self.parameters.dict[item0][1] == 'free':
+                    item0 += channel_id
+                value = self.parameters.dict[item0][0]
+                starinc[chan] = value
+                # read number of points
+                starres = self.parameters.dict['spotnpts'][0]
+                # read stellar rotation (if provided)
+                if 'spotrot' in self.parameters.dict.keys():
+                    starrot[chan] = self.parameters.dict['spotrot'][0]
```
taylorbell57, line 174: In the `setup` function we're pre-compiling the code to be evaluated analytically by theano/pymc3, so we need to use the theano tensor variables in the object `self.model` (using `getattr` since it is an object instead of a dictionary) instead of using the dictionary `self.parameters.dict` (which only contains the initial `float` guesses and would result in the `pymc3` fits silently failing since the star spot parameters were never used in the code)

taylorbell57: But in the `update` function down below, we're using the updated "best-fit" values after the fitting process which are stored in the `self.parameters.dict` dictionary, so your code in the `update` function is fine.

src/eureka/S5_lightcurve_fitting/models/FleckModel.py
```diff
@@ -0,0 +1,245 @@
...
+            light_curve = np.ma.ones(time.shape)
+            for pid in pid_iter:
+                # Initialize planet
+                pl_params = PlanetParams(self, pid, chan)
+
+                # Set spot/fleck parameters
+                # set a star rotation for the star object
+                # not actually used in fast mode. 
+                # overwritten if user supplies and runs slow mode
+                starrot[:] = 100
```
taylorbell57, line 148: Can just do `starrot = 100*np.ones((lc.nchannel_fitted))` above instead of doing this over multiple lines

src/eureka/S5_lightcurve_fitting/models/FleckModel.py
```diff
@@ -0,0 +1,245 @@
...
+                pl_params.u = uarray
+                pl_params.limb_dark = self.parameters.dict['limb_dark'][0]
+
+                # Enforce physicality to avoid crashes from batman by returning
+                # something that should be a horrible fit
+                if not ((0 < pl_params.per) and (0 < pl_params.inc < 90) and
+                        (1 < pl_params.a) and (0 <= pl_params.ecc < 1)):
+                    # Returning nans or infs breaks the fits, so this was the
+                    # best I could think of
+                    light_curve = 1e12*np.ma.ones(time.shape)
```
taylorbell57, line 205: Please change `1e12` to `1e6` as was done in the `BatmanTransitModel` - with `1e12` I was sometimes getting failed fits because the jump in log-likelihood was too sharp

src/eureka/S5_lightcurve_fitting/models/FleckModel.py
```diff
@@ -0,0 +1,245 @@
...
+                    u1 = 2*np.sqrt(pl_params.u[0])*pl_params.u[1]
+                    u2 = np.sqrt(pl_params.u[0])*(1-2*pl_params.u[1])
+                    pl_params.u = np.array([u1, u2])
+
+                # Make the star object
+                star = fleck.Star(spot_contrast=spotcon[chan], 
+                                  u_ld=pl_params.u, 
+                                  rotation_period=starrot[chan])
+                # Make the transit model
+                fleck_times = np.linspace(time[0], time[-1], npoints)
```
taylorbell57, line 229: This will fail if the first or last time values are masked (which does sometimes happen)

src/eureka/S5_lightcurve_fitting/plots_s5.py
```diff
@@ -793,3 +797,198 @@ def plot_GP_components(lc, model, meta, fitter, isTitle=True):
...
+
+        for ui in range(nus):   
+            item0 = 'u' + str(ui+1) + pid_id
+            if model.parameters.dict[item0][1] == 'free':
+                item0 += channel_id
+            uarray[channel, ui] = model.parameters.dict[item0][0]
+
+        pl_params.u = uarray[channel]
+    
+        fig = plt.figure(5306, figsize=(8, 6))
```
taylorbell57, line 885: Figure number 5306 is already taken by `Lightcurve.plot` - please update this to 5307

src/eureka/S5_lightcurve_fitting/plots_s5.py
```diff
@@ -793,3 +797,198 @@ def plot_GP_components(lc, model, meta, fitter, isTitle=True):
...
+            item0 += channel_id
+        value = model.parameters.dict[item0][0]
+        starinc[channel] = value
+        # read number of points
+        starres = model.parameters.dict['spotnpts'][0]
+        # read stellar rotation (if provided)
+        if 'spotrot' in model.parameters.dict.keys():
+            starrot[channel] = model.parameters.dict['spotrot'][0]
+    
+        fig = plt.figure(5306, figsize=(8, 6))
```
taylorbell57, line 976: Please update this to 5308

src/eureka/S5_lightcurve_fitting/plots_s5.py
```diff
@@ -793,3 +797,198 @@ def plot_GP_components(lc, model, meta, fitter, isTitle=True):
...
+        for n in range(nspots):
+            map.spot(contrast=spotcon[channel, n], radius=spotrad[channel, n], 
+                     lat=spotlat[channel, n], lon=spotlon[channel, n])
+        map.show(ax=ax)
+        if lc.white:
+            fname_tag = 'white'
+        else:
+            ch_number = str(channel).zfill(len(str(lc.nchannel)))
+            fname_tag = f'ch{ch_number}'
+        fname = (f'figs{os.sep}fig5307_{fname_tag}_starry_star_{fitter}'
```
taylorbell57, line 989: Please update this to 5308

src/eureka/S5_lightcurve_fitting/plots_s5.py
```diff
@@ -793,3 +797,198 @@ def plot_GP_components(lc, model, meta, fitter, isTitle=True):
...
+                  spotrad[channel][:nspots, None], 
+                  starinc[channel]*unit.deg, 
+                  planet=pl_params, time=0.0)
+        
+        if lc.white:
+            fname_tag = 'white'
+        else:
+            ch_number = str(channel).zfill(len(str(lc.nchannel)))
+            fname_tag = f'ch{ch_number}'
+        fname = (f'figs{os.sep}fig5306_{fname_tag}_fleck_star_{fitter}'
```
taylorbell57, line 901: Please update this to 5307

taylorbell57 (changes requested):

setup.cfg
```diff
@@ -36,6 +36,7 @@ install_requires =
     dynesty>1.0 # Lower limit needed for specific arguments
     emcee>3.0.0 # Lower limit needed for specific arguments
     exotic-ld==3.2.0 # Lower limit needed for updated JWST sensitivity files, upper limit needed for breaking changes
+    fleck
```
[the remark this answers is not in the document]
erinmmay: no limit on version to my knowledge

src/eureka/S5_lightcurve_fitting/differentiable_models/StarryModel.py
```diff
@@ -114,8 +114,80 @@ def setup(self, newparams):
...
+                spotlon = np.zeros((self.nchannel_fitted, nspots))
+                spotcon = np.zeros((self.nchannel_fitted, nspots))
+                starrot = np.zeros((self.nchannel_fitted))
+                starinc = np.zeros((self.nchannel_fitted))
+
+                # channel ID
+                if chan == 0:
+                    channel_id = ''
+                else:
+                    channel_id = f'_{chan}'
```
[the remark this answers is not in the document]
erinmmay: oop, thanks for catching. was on my list to update when I merged main into my branch and this got forgotten

src/eureka/S5_lightcurve_fitting/differentiable_models/StarryModel.py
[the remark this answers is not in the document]
erinmmay: True, but the way the multi channel fitting is set up it's going to create a copy of the parameter for each channel unless the user sets them to shared. I guess this catches that situation?

src/eureka/S5_lightcurve_fitting/differentiable_models/StarryModel.py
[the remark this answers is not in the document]
erinmmay: updated these arrays to be the defaults instead of just blank arrays of zeros in all locations (I assumed they'd always be overwritten, but if someone forgets to set one it makes sense to have it be the default)

src/eureka/S5_lightcurve_fitting/differentiable_models/StarryModel.py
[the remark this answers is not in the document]
erinmmay: yep, thought I changed this but missed it. Fixed

src/eureka/S5_lightcurve_fitting/differentiable_models/StarryModel.py
erinmmay: ahh good catch, added

src/eureka/S5_lightcurve_fitting/models/FleckModel.py
erinmmay: unless I'm missing something pid_id is never set in the BatmanTransitModel? I need it here since I have to read in additional parameters, whereas BatmanTransitModel just uses AstroModel's PlanetParams which does set the pid_id

erinmmay: Do you mean just move setting the pid_id inside the for `pid in pid_iter` loop?

src/eureka/S5_lightcurve_fitting/plots_s5.py
erinmmay: yeah I struggled on how to do this quickly. I need to know the number of spots and since the names of parameters can have things appended for multiple planets and multiple channels, I can't just count how many "spotrad"s are in the parameters and needed to account for that.

src/eureka/S5_lightcurve_fitting/plots_s5.py
erinmmay: Ah yeah, I wrote this before the AstroModel PlanetParams class became a thing and needed to find the limb darkening values (annoying that fleck needs them to plot up the star...). Will update

src/eureka/S5_lightcurve_fitting/plots_s5.py
erinmmay: It is different between fleck/starry (fleck only has one contrast, starry can have multiple), so there would be two versions of it regardless

src/eureka/S5_lightcurve_fitting/models/FleckModel.py
erinmmay: ah yes, at some point I had changed this to be the unmasked array and apparently that got lost along the way. Will fix

demos/JWST/S5_fit_par_template.epf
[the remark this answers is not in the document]
taylorbell57: @erinmmay, also note that this comment is still relevant; please indent using spaces rather than tabs

src/eureka/S5_lightcurve_fitting/differentiable_models/StarryModel.py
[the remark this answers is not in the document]
taylorbell57: Oh yeah, fair point

src/eureka/S5_lightcurve_fitting/models/FleckModel.py
taylorbell57: Ah yeah, I just meant that the `pid_id` never gets updated in the current version of your code. Thinking about it more clearly though, you shouldn't be using `pid_id` at all since the spot properties should not change between different planets. If the multi-planet fitting code is automatically making `_pl#` versions of these spot parameters, let me know I can help avoid that. I guess similarly, I should check that multi-planet fits don't duplicate the Lorentzian or Quasi-Periodic Pulsation parameters.

src/eureka/S5_lightcurve_fitting/plots_s5.py
taylorbell57: Maybe just tweak or add comments so that the future versions of ourselves can remember what is going on here?

docs/source/ecf.rst
```diff
@@ -1154,6 +1154,19 @@ This file describes the transit/eclipse and systematics parameters and their pri
       - ``osc_t1`` - This offset determined the start phase of the sinusoidal function.
 
          The time-dependent amplitude and period are defined as follows: ``amp = osc_amp * np.exp(-osc_amp_decay * (time - osc_t0))`` and ``per = osc_per * np.exp(-osc_per_decay * (time - osc_t0))``.  The damped oscillator model is then defined as: ``osc = 1 + amp * np.sin(2 * np.pi * (time - osc_t1) / per)``.  Note that ``osc[time < osc_t0] = 1``.
+         
+   - Star Spot Parameters
+      - ``spotstari`` - The stellar inclination in degrees. Only matters when the star is rotating so that spots move appropriately relative to the planet, fix to 90 unless you're accounting for stellar rotation (fleck slow mode, see spotrot below), then place priors based on inclination measurements of the stellar inclination. Recommend using fast mode and fixing to 90 if this is not known! 
+      - ``spotrot`` - The stellar rotation rate in days. For fleck, only assign if you'd like to run in slow mode! (In slow mode the star rotates and spots move appropriately. Otherwise Eureka! will use fleck's fast mode which assumes the stellar rotation is >> transit time and spots are stationary)
```
taylorbell57, line 1160: Can you add comments to these two variables, telling users that they should set them to be shared if doing multwhite or shared fits?

demos/JWST/S5_fit_par_template.epf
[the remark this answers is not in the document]
erinmmay: ah weird, I know I did this - it's showing as spaces on my computer. If you look at the most recent version it is aligned, so this might just be displaying an outdated file

src/eureka/S5_lightcurve_fitting/differentiable_models/StarryModel.py
erinmmay: not gonna lie, a little lost on this one. For some reason I no longer can successfully fit a spot-free light curve with starry so it's hard to test the implementation of spots.

taylorbell57 (changes requested):

src/eureka/S5_lightcurve_fitting/differentiable_models/StarryModel.py
```diff
@@ -114,8 +114,80 @@ def setup(self, newparams):
             pl_params = PlanetParams(self, 0, chan, eval=False)
 
             # Initialize star object
-            star = starry.Primary(starry.Map(udeg=self.udeg),
-                                  m=0, r=pl_params.Rs)
+            # check for spots and set parameters
+            if hasattr(self.parameters, 'spotrad0'):
```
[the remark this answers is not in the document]
kevin218: We should be consistent with our naming convention, which means no 0.

taylorbell57:
Alright, I've made several changes to try and make things simpler to maintain and simpler to use. The fleck transit model now heavily re-uses the `BatmanTransitModel` in a similar fashion to what we did with the POET transit model. I fixed a crash that'd occur if `rp<0`, removed the (now implicit) 0 from the 0th spot parameter names (e.g., now just `spotrad` instead of `spotrad0`). I added a bit of code so that the units of `spotcon` and `spotrad` in the starry model are the same as those in the fleck model. I moved some of the duplicated code to AstroModel to tidy things up. I added an initial plot before fitting begins to help show users where their initial spot guess is (which helped me initialize the spot within the transit chord). I added some starry vs fleck specific defaults to `spotnpts`. I greatly increased the default `spotrot` value (to 10,000 years) since the `spotlat` and `spotlon` parameters are with respect to t=0 while the times being used in the fit are in BMJD where t>>>0, so the fitted `spotlat` and `spotlon` parameters were very hard to interpret with the previous `spotrot` default of only 10 days. I checked that the models seemed to work well with multi-planet and multi-channel fits. The starry spot model is indeed excruciatingly slow, so I didn't test it near as fully as I did the fleck model. I've tested the code on a transit observation that I have which has a spot crossing, and it seems to work quite well.

@kevin218, since I've made so many edits to this PR now (and am likely less impartial now), you should probably give this another review

src/eureka/S5_lightcurve_fitting/differentiable_models/StarryModel.py
```diff
@@ -113,99 +113,65 @@ def setup(self, newparams):
...
+                    spotcon = tt.concatenate([
+                        spotcon, [getattr(pl_params, f'spotcon{spot_id}'),]])
+
+                # Apply some conversions since inputs are in fleck units
+                spotrad *= 90
+                spotcon = 1-spotcon
+
+                if pl_params.spotnpts is None:
+                    # Have a default spotnpts for starry
+                    pl_params.spotnpts = 30
```
kevin218, line 145: Should we assign a default value here or in AstroModel.py?

src/eureka/S5_lightcurve_fitting/models/FleckModel.py
```diff
@@ -23,223 +19,108 @@ def __init__(self, **kwargs):
...
+            spotrad = np.concatenate([
+                spotrad, [getattr(pl_params, f'spotrad{spot_id}'),]])
+            spotlat = np.concatenate([
+                spotlat, [getattr(pl_params, f'spotlat{spot_id}'),]])
+            spotlon = np.concatenate([
+                spotlon, [getattr(pl_params, f'spotlon{spot_id}'),]])
+
+        if pl_params.spotnpts is None:
+            # Have a default spotnpts for fleck
+            pl_params.spotnpts = 300
```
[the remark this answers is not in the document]
taylorbell57: AstroModel.py can't tell whether you're fitting with Fleck or Starry, which is why I put these package-specific default values in their respective model files

src/eureka/S5_lightcurve_fitting/differentiable_models/StarryModel.py
taylorbell57: AstroModel.py can't tell whether you're fitting with Fleck or Starry, which is why I put these package-specific default values in their respective model files

kevin218 (approved):

PR #640, opened 2024-04-01
Upgrading to `jwst==1.14.0` and `Python>=3.10`, and bug fixes

taylorbell57:
This gives us access to the EMI correction step for MIRI data in Stage 1 (introduced in `jwst` v1.13.4) but requires us to use `Python>=3.10`. After initial testing on my end, the change in Python version seems to have no consequences, but someone with an Apple Silicon computer should probably run a check. I've also added an extra testing suite: now the code will be tested with and without the PyMC3 optional install which is important as that option continues to fall behind because of that hard upper-limit on the `jwst` version caused by `theano`'s deprecation and the resulting `numpy` upper limit.

I also added bug fixes in a couple places. Fixed an incorrect self-import style used in S1, fixed the chi-squared calculation in the case of no masked points (in which case `~lc.flux.mask` just gives `True` which gets treated as a `1` which in turn gives a deeply negative reduced chi-squared value), and added the ability to continue an emcee chain with a subset of the parameters that had been used before.


src/eureka/S1_detector_processing/ramp_fitting.py
```diff
@@ -369,6 +369,8 @@ def calc_opt_sums_uniform_weight(rn_sect, gain_sect, data_masked, mask_2d,
 
     Parameters
     ----------
+    ramp_data : RampData
+        Input data necessary for computing ramp fitting. (Unused)
```
kevin218, line 373: Ok, you've got me stumped.  Why are you including a variable that goes unused?

taylorbell57: Because their function now has the `ramp_data` as a required input (see [here](https://github.com/spacetelescope/stcal/blob/a57fe3474291b855d937e920ce8ebfbced4b990d/src/stcal/ramp_fitting/ols_fit.py#L3833)), so without making this change we get a crash because of mismatched arguments

kevin218 (approved):

taylorbell57:
@kevin218, just had to deal with a merge conflict, and now this is good to go again

kevin218 (approved):

Issue #628, opened 2024-02-27
[Bug]: segmentation fault MC3 package after running light curves


evamariaa:
### FAQ check

- [X] Yes, I checked the FAQ and my question has not been addressed.

### Instrument

Light curve fitting (Stages 4-6)

### What happened?

I get a segmentation fault after running light curve fit (same for dynesty and emcee) and I traced down to MC3 package. 
It's the function from MC3 `from mc3.stats import time_avg` that causes it, previous Allan plot looks fine so I'm not sure what the issue is that MC3 crashes - is it a Eureka! or a MC3 problem? Either way it's flagged this way. 
I created a new conda environment and new installation and it still pops up. 

### Error traceback output

When running with gdb:

EMCEE RESULTS:
rp: 0.09314383900504214 (+0.00023746832716464716, -0.00024390231091873937)
t0: 60336.16629344476 (+0.0001366072174278088, -0.00014108874165685847)
inc: 86.14041037419365 (+0.03564806637537288, -0.03959720206140105)
a: 7.5178380816512576 (+0.012216578350765062, -0.012060384627009313)
u2: 0.0596876616903001 (+0.01720078126260373, -0.016858445839115672)
c0: 1.0037328901325697 (+3.041406496073762e-05, -2.8784564057193762e-05)
c1: -0.00018125230315689576 (+0.00021138404946226995, -0.00022132731635158144)
scatter_mult: 4.604809762772209 (+0.1430173800715835, -0.12412840317554785); 489.802873463214 (+15.212425121351847, -13.203248708624503) ppm

Program received signal SIGSEGV, Segmentation fault.
0x00007ffe987d7c1e in invgamma ()
   from /home/ahrer/.conda/envs/eureka-dev/lib/python3.9/site-packages/mc3/lib/_time_averaging.cpython-39-x86_64-linux-gnu.so

### What operating system are you using?

Linux

### What version of Python are you running?

Python 3.9.7

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

taylorbell57:
@evamariaa, can you see if Kevin's tweak in PR #609 for the file `src/eureka/S5_lightcurve_fitting/plots_s5.py` fixes this issue for you?

evamariaa:
Ok, just installed from PR https://github.com/kevin218/Eureka/pull/609 and running through Stage 5 now, so far I didn't get a segmentation fault yet so I'm cautiously optimistic and will let you know if it stays like this! Thanks!

[...]

PR #641, opened 2024-04-05
Create exception for MC3 failed import

kevin218:
This is a temporary workaround to Issue #628.  If MC3 fails to import, the user is alerted with a print statement and no Allan variance plots will get produced in Stage 5.

Enough people are having this problem that I wanted to get this workaround into main ASAP.


environment.yml
```diff
@@ -39,7 +39,7 @@ dependencies:
         - crds
         - exotic-ld==3 # Lower limit needed for updated JWST sensitivity files, upper limit needed for breaking changes
         - image_registration@git+https://github.com/keflavich/image_registration@master # Need GitHub version to avoid np.float issue
-        - jwst==1.11.4
+        - jwst==1.12.2
```
taylorbell57, line 42: Why are we updating this only here?

taylorbell57 (changes requested):

environment.yml
kevin218: It wasn't updated here.  Now it matches `setup.cfg`.

taylorbell57 (approved):

Issue #649, opened 2024-05-13
[Bug]: Setting meta.centroids fails when nreads is not constant in each WFC3 exposure.


jbrande:
### FAQ check

- [X] Yes, I checked the FAQ and my question has not been addressed.

### Instrument

Other (any stage)

### What happened?

Some WFC3 datasets don't have consistent numbers of reads in each exposure. When this is the case, setting `meta.centroids = np.array(meta.centroids)` in the WFC `conclusion_step()` will fail, due to very so slightly ragged array with an error `ValueError: setting an array element with a sequence. The requested array has an inhomogeneous shape after 1 dimensions. The detected shape was (85,) + inhomogeneous part.`

I guess that step should have a `dtype=object` parameter to preserve all the read-level information for future steps rather than trying to trim or pad the number of reads.

### Error traceback output

```
Traceback (most recent call last):
  File "/home/jbrande/miniconda3/envs/eureka/lib/python3.10/pdb.py", line 1723, in main
    pdb._runscript(mainpyfile)
  File "/home/jbrande/miniconda3/envs/eureka/lib/python3.10/pdb.py", line 1583, in _runscript
    self.run(statement)
  File "/home/jbrande/miniconda3/envs/eureka/lib/python3.10/bdb.py", line 598, in run
    exec(cmd, globals, locals)
  File "<string>", line 1, in <module>
  File "/mnt/c/Users/yonib/OneDrive - University of Kansas/ExoLab/hst/TOI1759/ecfs_laptop/run_eureka_wfc3.py", line 16, in <module>
    s3_spec, s3_meta = s3.reduce(eventlabel, ecf_path=ecf_path)
  File "/home/jbrande/Planetary/Eureka/src/eureka/S3_data_reduction/s3_reduce.py", line 755, in reduce
    spec, meta, log = inst.conclusion_step(spec, meta, log)
  File "/home/jbrande/Planetary/Eureka/src/eureka/S3_data_reduction/wfc3.py", line 162, in conclusion_step
    meta.centroids = np.array(meta.centroids)
ValueError: setting an array element with a sequence. The requested array has an inhomogeneous shape after 1 dimensions. The detected shape was (85,) + inhomogeneous part.
```

### What operating system are you using?

Ubuntu 22.04.03 on WSLg on Win10

### What version of Python are you running?

3.10.14

### What Python packages do you have installed?

```
_libgcc_mutex             0.1                        main
_openmp_mutex             5.1                       1_gnu
[...]

PR #650, opened 2024-05-13
Addressing #649

jbrande:
Holding off on merging this until I can test more thoroughly, but this should allow for WFC3 exposures of inconsistent read lengths to be properly handled.

### DIFF
--- a/src/eureka/S3_data_reduction/wfc3.py
+++ b/src/eureka/S3_data_reduction/wfc3.py
@@ -159,7 +159,7 @@ def conclusion_step(data, meta, log):
     log : logedit.Logedit
         The updated log.
     """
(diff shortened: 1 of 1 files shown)


jbrande:
@taylorbell57, maybe there's a more intelligent place for this check earlier in the step, but this is the first failure point as a result.

taylorbell57:
@jbrande, can you remind me what the status of this PR is? I think the recommendation I gave during the team meeting was to pad arrays with NaN values so that you don't have to deal with ragged arrays since I think that was causing an issue for you later on in the code; have you tried that?

taylorbell57:
Closing as this code was included in PR #681

Issue #597, opened 2023-12-20
[Bug]: If a user marks a column as saturated in every group in S1, the column is still used in S3 


erinmmay:
### FAQ check

- [X] Yes, I checked the FAQ and my question has not been addressed.

### Instrument

NIRSpec (Stages 1-3)

### What happened?

From a chat with @gianninapr - Currently Stage 1 has the option for a user to define which columns are saturated in which group so they can be ignored for GLBS and ramp fitting. However in some cases a user may want to manually flag a column as saturated in every group. In this case, the column isn't ignored in Stage 3 because there we only ignore pixels with an odd DQ (or *bad* pixels) and saturated pixels are not necessarily flagged as bad pixels. This means the Stage 3 code tries to use them for the median frame etc. We should add a check in the Stage 1 saturation flagging code so that columns flagged as saturated in *every* group get an updated DQ value so that they are also flagged as bad pixels. 

Along the same lines: allow a user to supply a custom DQ mask in Stage 1. I'll include that option in a PR to address this bug.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

gianninapr:
Some notes on this issue:

After my conversation with Erin, I tried running a quick fix, but the results were still not ideal. After going over the definitions of the masks and a conversation with Caleb Cañas we both noticed the issue as it stands. 

In [`update_saturation.py`](https://github.com/kevin218/Eureka/blob/main/src/eureka/S1_detector_processing/update_saturation.py) around line 46 the user-provided custom saturation columns are set to the `sat_flag` value, which is defined as the 'SATURATED' jwst dq flag. When the flags are passed in [`s3_reduce.py`](https://github.com/kevin218/Eureka/blob/main/src/eureka/S3_data_reduction/s3_reduce.py), the code marks all odd entries as do not use, yet the dq flag for 'SATURATED' is an even number (2), so it isn't marking the user-defined saturated columns as do not use, which is why we still see them on the median frame in the end. 

I'm happy to take care of this if there's a straightforward fix (like adding a condition to also catch the saturation flag in the if statement in `s3_reduce.py`) that would be sufficient, but just wanted to check in and see if there was more thoughts people had and what they thought an elegant solution is.

erinmmay:
[...]

Issue #546, opened 2023-07-06
[Enhancement]: Automatically skip completely masked lightcurves in S5


taylorbell57:
### Instrument

Light curve fitting (Stages 4-6)

### What is your suggestion?

At present S5 will crash if it gets to a lightcurve that is completely masked (something that happens more often now that some S3 bug has been introduced). We should just automatically skip any channels that are completely masked.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #651, opened 2024-05-20
Flags fully saturated pixel columns as DO_NOT_USE

gianninapr:
Issue Addressed
-----------------
This PR addresses Issue #597 in which columns that were fully saturated (the first group already saturated) were still being used in Stage 3 and saturation was being manually removed from the first group if found.

Fix Introduced
----------------
From conversations with @erinmay the best fix seemed to be to get rid of the portion of the code that removed saturation from the first group, as well as mark any columns that are fully saturated as DO_NOT_USE. And with @cicanas help to understand JWST flagging I was able to brainstorm a path forward for this by establishing a similar condition check to what was already established for partial saturation but for full saturation. 

Elegance
-----------
I tried to keep the same format that was being used for most of the script, and I did wonder if maybe a more elegant approach for the conditions was an if/else loop, but I was worried that separating the conditions would omit actual saturation flags that might still be needed in favor of just DO_NOT_USE flags. But happy to work on a more elegant solution if anyone sees any obvious ways to update it.

Testing
------------
I was able to test this on WASP 39b data with full saturation in columns 60-144 and also tested it with only partial saturation in the second group to make sure the condition was not catching partial saturation. If others want to test it feel free! I'll post a side by side comparison below of one of our Stage 3 results that showcase the bug fix.

old
new


taylorbell57:
This is not something that needs to be addressed in this PR, but this new (helpful!) feature will further increase the need to address issue #546. Just noting that here and linking that issue to remind myself and others why resolving that other issue is important

taylorbell57 (dismissed):
Looks good to me - unless @kevin218 wants to review this as well, I say we go ahead and merge! Great work @gianninapr!

taylorbell57 (approved):
Approving again after resolving merge conflict

PR #652, opened 2024-05-22
Bug Patch: S6 with similar variable names and S3 undefined aperture_shape

taylorbell57:
This little PR resolves the issue where Stage 6 plots malfunctioned for the case where some variables were getting confused by the S6 code. Additionally, I added a small patch to make sure that the default `meta.aperture_shape` was applied early enough in the code

### DIFF
--- a/src/eureka/S6_planet_spectra/s6_spectra.py
+++ b/src/eureka/S6_planet_spectra/s6_spectra.py
@@ -423,7 +423,9 @@ def parse_s5_saves(meta, log, fit_methods, channel_key='shared'):
 
         fname = f'S5_{fitter}_samples_{channel_key}'
 
-        keys = [key for key in full_keys if y_param == key[:len(y_param)]]
+        temp_keys = [y_param+f'_{c}' if c > 0 else y_param
+                     for c in range(meta.nspecchan)]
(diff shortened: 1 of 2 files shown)


src/eureka/S6_planet_spectra/s6_spectra.py
```diff
@@ -423,7 +423,9 @@ def parse_s5_saves(meta, log, fit_methods, channel_key='shared'):
 
         fname = f'S5_{fitter}_samples_{channel_key}'
 
-        keys = [key for key in full_keys if y_param == key[:len(y_param)]]
+        temp_keys = [y_param+f'_{c}' if c > 0 else y_param
+                     for c in range(meta.nspecchan)]
+        keys = [key for key in temp_keys if key in full_keys]
```
kevin218, line 428: I think we need to apply a similar fix on Line 451

kevin218 (changes requested):

src/eureka/S6_planet_spectra/s6_spectra.py
taylorbell57: Good point!

kevin218 (approved):

PR #657, opened 2024-06-04
Removing trailing spaces/tabs from all files

taylorbell57:
I found a way to remove all trailing spaces from files automatically - I know your editor often does this @kevin218 automatically, and I just set mine up to do so as well. All this PR does is apply that change to all files to reduce the frequency with which we'll see changes like that in future PRs, hopefully making future PRs a bit easier.

If you don't think this is worthwhile, I'm totally fine just deleting this PR - took like 5 seconds, so no real lost time on my end


kevin218 (approved):
Nice, this should save us some time in the future.

FYI, I didn't actually look at all of the files and trust that there are no meaningful changes in this PR.

PR #663, opened 2024-06-18
Adding an astrophysical model to combine all physical models, adding quasi-Lambertian phase curve

taylorbell57:
_Note that this PR is still a draft and should not yet be reviewed._

_Note that this PR builds on PRs #632 and #659 and should only be reviewed after those PRs are merged into main._

This PR adds an AstroModel to both `models` and `differentiable_models` to deal with safely combining different astrophysical models. This was in part necessitated by another addition in this PR which adds a quasi-Lambertian phase curve model (based on Agol+2007) which is useful for modelling thermal phase curves of airless worlds. I also spent some time trying to make the PyMC3 models closer to the non-differentiable models to reduce duplicated code.

In addition, this PR allows linear extrapolation of limb-darkening coefficients beyond the edge of the nominal filter range of ExoTiC-LD - I regularly encounter issues with using fixed LD coeffs from ExoTiC-LD because the nominal filter coverage does not extend as far into the edges of the filter as I'd like to use. This works okay in many cases, but in some edge cases it can result in some very strange behaviour.

Things to do before this PR is ready for review include:
- [x] Add quasi-Lambertian phase curve model to `differentiable_models`
- [x] Merge this PR with the code from the branch `metaClass`
- [x] Merge this PR with the code from the branch `reformatting_S5_varnames`
- [x] Resolve weird edge-cases with exotic-ld extrapolation


kevin218:
Can the quasi-Lambertian also be used for reflected light?

taylorbell57:
> Can the quasi-Lambertian also be used for reflected light?

Yeah, I believe you'd just need to fix gamma=4

taylorbell57:
Correction, apparently the best-fit gamma coefficient to approximate a Lambertian phase function was 3.285 - Eric just used 4 in his 2007 paper since it was close-ish and had some nice mathematical properties for his application. If desired, we can add an actual Lambertian phase function later (perhaps with an albedo term as well, so we can handle mixtures of emitted and reflected light with NIRSpec PRISM and NIRISS phase curves) - this quasi-Lambertian was just useful for the TRAPPIST-1 phase curve data (since the 3D modelling folks found that different airless scenarios could be well approximated with the quasi-Lambertian function).

kevin218:
Yeah, I'm in the position of needing to fit both thermal and reflected
light.

With that said, no planet is a perfect Lambertian, and usually far from it,
so fitting for gamma may be reasonable. I'll have to run some tests when I
get back.

taylorbell57:
@kevin218, there have been numerous circumstances in which I've wanted to compute model limb-darkening coefficients for wavelengths that are outside of the wavelengths for which ExoTiC-LD has throughput curves at the short end of NIRSpec G395H NRS1 and for the long end of NIRCam F444W. I propose we use the polynomial extrapolations I've computed (and plotted below) for those two filters to allow approximate computation of the LD coefficients outside the official throughput curve regime. A warning is raised in the log when this is done, and Fig 4303 will be made (if `meta.isplots_S4 >= 3`) which will show figures similar to the ones below.

JWST_NIRCam_F444
JWST_NIRSpec_G395H

This change resolves the last to-do item for this PR on my end, but because this PR builds on PRs #632 and #659, I will keep this PR as a draft and recommend you do not review this PR until we've merged those two other PRs.

kevin218:
@taylorbell57 There are a lot of changes here and I'd like to get a head start on reviewing them.  Is it ok if I do so?

src/eureka/S5_lightcurve_fitting/models/QuasiLambertianPhaseCurve.py
```diff
@@ -0,0 +1,102 @@
...
+                    # the planet is on a circular orbit
+                    t = time - pl_params.t_secondary
+                    phi = 2*np.pi/pl_params.per*t
+                else:
+                    # the planet is on an eccentric orbit
+                    anom = true_anomaly(pl_params, time)
+                    phi = anom + pl_params.w*np.pi/180 + np.pi/2
+
+                # calculate the phase variations
+                phaseVars = np.abs(np.cos(phi/2))**pl_params.gamma
```
kevin218, line 98: Let's make this function a little more generic to allow for reflected light signals due to clouds in Hot Jupiters.  This involves adding an offset parameter (e.g., `phi-phi0` here or 't-t0' above) that allows the peak to occur before or after secondary eclipse.

taylorbell57:
> @taylorbell57 There are a lot of changes here and I'd like to get a head start on reviewing them. Is it ok if I do so (once #632 is merged)?

Sure thing, I just wanted you to hold off since the PR looks bigger than it really is since those other two PRs of mine are nested inside this one

src/eureka/S5_lightcurve_fitting/differentiable_models/QuasiLambertianPhaseCurve.py
```diff
@@ -0,0 +1,120 @@
...
+                    # the planet is on a circular orbit
+                    t = time - pl_params.t_secondary
+                    phi = 2*np.pi/pl_params.per*t
+                else:
+                    # the planet is on an eccentric orbit
+                    anom = true_anomaly(pl_params, time, lib)
+                    phi = anom + pl_params.w*np.pi/180 + np.pi/2
+
+                # calculate the phase variations
+                phaseVars = abs(lib.cos(phi/2))**pl_params.gamma
```
kevin218, line 113: See comment in other quasi-Lambertian function

kevin218 (changes requested):
Nice job in devising and implementing the AstroModel class

src/eureka/S5_lightcurve_fitting/models/AstroModel.py
```diff
@@ -86,6 +86,7 @@ def __init__(self, model, pid=0, channel=0, eval=True):
         self.AmpCos2 = 0.
         self.AmpSin2 = 0.
         self.gamma = 0.
+        self.quasi_offset = 0.
```
kevin218, line 89: In the interest of trying to group related parameters from the same systematic model, what do you think of using the name `quasi_gamma`?

kevin218 (approved):

PR #665, opened 2024-06-21
Allow setting of matplotlib backend and layout engine

### LINKED ISSUE #647: [Bug]: Matplotlib layout engine only defined if users run `eureka.lib.plots.set_rc(style='eureka')`
Without running `eureka.lib.plots.set_rc(style='eureka')` before running Eureka! (e.g. in the `run_eureka.py` file), the matplotlib layout engine is not defined which causes a crash. You should be allowed to not run that function, but we need to make sure the matplotlib layout engine is always initialized one way or another.

_Originally posted by @taylorbell57 in https://github.com/kevin218/Eureka/issues/646#issuecomment-2086216180_

taylorbell57:
This PR adds support for inputting matplotlib backend to `lib.plots.set_rc()`.

Additionally, this PR ensures that the matplotlib layout engine is always set, which resolves #647

I've tested this on my computer and importing eureka is enough to make sure all the plots are made successfully, but users should still run the `eureka.lib.plots.set_rc(style='eureka')` step. I didn't want to force the eureka style on everyone who ever imports eureka, but I needed to make sure the layout engine and figure filetype were defined, so I tried to strike a minimally intrusive balance.

### DIFF
--- a/src/eureka/lib/plots.py
+++ b/src/eureka/lib/plots.py
@@ -1,10 +1,11 @@
+import matplotlib
 from matplotlib import rcdefaults, rcParams
 
 # Default figure file type
 figure_filetype = '.png'
 
 
-def set_rc(style='preserve', usetex=False, filetype='.png',
+def set_rc(style='preserve', usetex=False, backend=None, filetype='.png',
            from_scratch=False, **kwargs):
     """Function to adjust matplotlib rcParams for plotting procedures.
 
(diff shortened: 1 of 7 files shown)


src/eureka/lib/plots.py
```diff
@@ -19,7 +20,17 @@ def set_rc(style='preserve', usetex=False, filetype='.png',
...
         Do you want to use LaTeX fonts (which requires LaTeX to be
-        installed), by default False
+        installed), by default False. Can also set to None to not change
+        the current matplotlib setting.
+    layout : str; optional
+        Specifies the matplotlib.layout_engine that you want to use.
+        Defaults to 'constrained' which is known to make nice plots.
+        Can also try 'tight' (which uses the tight layout engine)
+        or None, which uses the default layout engine, neither of
+        which are untested and not recommended.
```
kevin218, line 30: "neither of which are tested..."

kevin218 (changes requested):

kevin218 (approved):

PR #671, opened 2024-07-11
Fixing our readthedocs pages

taylorbell57:
This tiny little PR fixes our Read the Docs pages. Since requiring Python >= 3.10 our docs page has apparently been failing to build since the .readthedocs.yml file was setup to use Python 3.9. I've tested this change and can confirm it resolves the issue.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -7,9 +7,9 @@ version: 2
 
 # Set the version of Python and other tools you might need
 build:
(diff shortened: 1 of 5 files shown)


kevin218 (dismissed):
There was a discussion in an issue where someone notes that the documentation didn't mention the user needs to source their .bash_profile after making changes.  Should we make that change here?

kevin218 (approved):

PR #681, opened 2024-08-13
Minor documentation edits to reflect pre-1.0 testing

jbrande:
Minor edits, warning users of a dependency conflict that will cause installation to fail when specifying both `jwst` and `pymc3`, as well as one warning users about potential issues with HST spatial scan observations of inconsistent numbers of exposures.

### DIFF
--- a/docs/source/quickstart.rst
+++ b/docs/source/quickstart.rst
@@ -253,9 +253,8 @@ Note that HST operations start at Stage 3, unlike JWST operations. HST systemati
 can also be slightly more complicated to fit, so it may be worthwhile to read some
 papers discussing these, such as `Knutson et al. (2014) <knutson2014_>`_ or
 `Brande et al. (2022) <brande2022_>`_. Exponential ramps are available as systematics
-models in ``Eureka!`` Stage 5, but the spatial scan model described in
-`Brande et al. (2022) <brande2022_>`_ is forthcoming, so currently only a single
-scan direction will be fit.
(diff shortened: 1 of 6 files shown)


docs/source/quickstart.rst
```diff
@@ -241,6 +241,12 @@ where ``[60]`` can be replaced with a list of the visit numbers you want to
 download, e.g. ``[60, 61, 62, 63, 64]`` if you want the entire transit
 observation.
 
+.. warning::
+	HST planning is usually consistent between orbits, but may sometimes modify 
+	the number of exposures in a scan for observation timing purposes. Ensure 
+	that your datasets have consistent numbers of exposures, or exclude 
+	trailing exposures at the end of orbits if needed.
+
```
taylorbell57, line 249: I'm not sure this is the best place for this warning. Thoughts @kevin218?

kevin218: I thought the issue was with a change in the number of up-the-ramp samples (i.e. non-destructive reads).

jbrande: Sorry, yes, it was non-destructive reads.

kevin218: Ok, great. 

Now going back to Taylor's concern, we could put it in the faq.

jbrande: Sounds good. I've moved that warning to the FAQ, and made sure the appropriate errors get thrown to inform the user of the issue before they try to go further in the analysis and get stuck.

taylorbell57 (approved):

PR #715, opened 2024-11-05
Update download script

kevin218:
This tiny PR fixes the JWST download script because it was broken and also fixes a few minor typos.

### DIFF
--- a/src/eureka/S3_data_reduction/nirspec.py
+++ b/src/eureka/S3_data_reduction/nirspec.py
@@ -7,7 +7,7 @@ from ..lib.util import read_time, supersample
 
 
 def read(filename, data, meta, log):
-    '''Reads single FITS file from JWST's NIRCam instrument.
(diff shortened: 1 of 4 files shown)


taylorbell57 (approved):

taylorbell57:
The failed pytest can't be related to your changes, but the test does keep failing with the message:

```
ImportError: cannot import name 'MMMBackground' from 'photutils'
```

From the install log, the installed `photutils` version is 2.0.2. Looking at their GitHub, it looks like we should be doing

```
from photutils.background import MMMBackground, MedianBackground, Background2D
```

while we're currently doing
```
from photutils import MMMBackground, MedianBackground, Background2D
```

I think that change likely took place sometime around 2.0.

taylorbell57:
Ah, we fixed that a while ago, and your branch is just based on an old version of Eureka. Should be all good once merged into main

PR #716, opened 2024-11-05
S5 bug patches

taylorbell57:
This little PR fixes some S5 bugs I've noticed in the past week.

There was an error thrown if the user didn't specify `w` which would result in `w` being set to `None` which isn't allowed.

There was also a bug in S5 for fleck/starry spot modells when doing non-shared spectroscopic fits that only popped up once the channel number was >0.

I also added a safety uniform prior to the fleck model to ensure that some key parameters don't end up crashing the code


kevin218 (approved):

taylorbell57:
@kevin218, I had to resolve a merge conflict with the catwoman PR, but this is now ready again for approval

kevin218 (dismissed):

kevin218:
@taylorbell57 Before merging, can you also update the installation instructions to reference v1.1?  There have been a number of important updates and bug fixes that people should be using in their version of Eureka!.

kevin218 (approved):

PR #726, opened 2024-12-03
Save background model at source position and stddev of background counts in S3, create background lightcurves in S4

erinmmay:
This functionality enables users to examine the background for trends without saving the full 2D background array (large file sizes!). Also makes plots of the background light curves in S4 for quick inspection.

### DIFF
--- a/src/eureka/S3_data_reduction/background.py
+++ b/src/eureka/S3_data_reduction/background.py
@@ -152,7 +152,19 @@ def BGsubtraction(data, meta, log, m, isplots=0):
     data['flux'] -= data.bg
     if hasattr(data, 'medflux'):
         data['medflux'] -= np.median(data.bg, axis=0)
-
+        
+    # Save BG value at source position and BG stddev (no outlier rejection)
+    data['skylev'] = (['time', 'x'], np.zeros([data.flux.shape[0],data.flux.shape[2]]))
+    data['skylev'].attrs['flux_units'] = data['flux'].attrs['flux_units']
+    data['skylev'] = data.bg[:, meta.src_ypos, :]
+    
+    data['skyerr'] = (['time', 'x'], np.zeros([data.flux.shape[0],data.flux.shape[2]]))
+    data['skyerr'].attrs['flux_units'] = data['flux'].attrs['flux_units']
+    y_inds = np.zeros([data.flux.shape[1]], dtype = bool)
+    y_inds[:meta.bg_y1] = True
+    y_inds[meta.bg_y2:] = True
+    data['skyerr'] = np.std(data.flux[:,y_inds,:], axis = 1)
+    
     # Make image+background plots
     if isplots >= 3:
         plots_s3.image_and_background(data, meta, log, m)
--- a/src/eureka/S3_data_reduction/background.py
+++ b/src/eureka/S3_data_reduction/background.py
@@ -154,13 +154,13 @@ def BGsubtraction(data, meta, log, m, isplots=0):
         data['medflux'] -= np.median(data.bg, axis=0)
         
     # Save BG value at source position and BG stddev (no outlier rejection)
-    data['skylev'] = (['time', 'x'], 
-                      np.zeros([data.flux.shape[0], data.flux.shape[2]]))
+    coords = list(data.coords.keys())
+    coords.remove('y')
+    data['skylev'] = (coords, np.zeros_like(data.flux[:, 0]))
     data['skylev'].attrs['flux_units'] = data['flux'].attrs['flux_units']
     data['skylev'] = data.bg[:, meta.src_ypos, :]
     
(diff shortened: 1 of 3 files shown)


src/eureka/S3_data_reduction/background.py
```diff
@@ -152,7 +152,21 @@ def writeBG_WFC3(arg):
...
+    data['skylev'].attrs['flux_units'] = data['flux'].attrs['flux_units']
+    data['skylev'] = data.bg[:, meta.src_ypos, :]
+    
+    data['skyerr'] = (['time', 'x'], 
+                      np.zeros([data.flux.shape[0], data.flux.shape[2]]))
+    data['skyerr'].attrs['flux_units'] = data['flux'].attrs['flux_units']
+    y_inds = np.zeros([data.flux.shape[1]], dtype=bool)
+    y_inds[:meta.bg_y1] = True
+    y_inds[meta.bg_y2:] = True
+    data['skyerr'] = np.std(data.flux[:, y_inds, :], axis=1)
```
kevin218, line 168: I think we should start with data.mask here, instead of all zeros.  Something like `y_inds = deepcopy(data.mask)`.  Since the masks in Eureka! now set bad values to True, we would set `y_inds[:, meta.bg_y1:meta.bg_y2] = True`.  

The std dev calculation may also require an update based on the new mask implementation, but I'm not certain.  Maybe it would be easiest to create a masked array and then do the calculation?

src/eureka/S3_data_reduction/background.py
```diff
@@ -152,7 +152,21 @@ def writeBG_WFC3(arg):
     data['flux'] -= data.bg
     if hasattr(data, 'medflux'):
         data['medflux'] -= np.median(data.bg, axis=0)
-
+        
+    # Save BG value at source position and BG stddev (no outlier rejection)
+    data['skylev'] = (['time', 'x'], 
+                      np.zeros([data.flux.shape[0], data.flux.shape[2]]))
```
kevin218, line 158: This declaration is going to be problematic for NIRISS because it has an extra `order` dimension.  Instead, do the following...
`coords = list(data.coords.keys())`
`coords.remove('y')`
`data['skylev'] = (coords, np.zeros_like(data.flux[:, 0]))`

Then apply similar code for `skyerr` below.
`data['skyerr'] = (coords, np.zeros_like(data.skylev))`

src/eureka/S4_generate_lightcurves/plots_s4.py
```diff
@@ -84,6 +84,87 @@ def binned_lightcurve(meta, log, lc, i, white=False):
     fig.savefig(meta.outputdir+fname, bbox_inches='tight', dpi=300)
     if not meta.hide_plots:
         plt.pause(0.2)
+        
+
+def binned_background(meta, log, lc, i, white=False):
+    '''Plot each spectroscopic light curve. (Figs 4102)
```
kevin218, line 90: Update the description and figure number

src/eureka/S4_generate_lightcurves/s4_genLC.py
```diff
@@ -463,6 +494,8 @@ def genlc(eventlabel, ecf_path=None, s3_meta=None, input_meta=None):
                 # Plot each spectroscopic light curve
                 if meta.isplots_S4 >= 3:
```
kevin218, line 495: Wow, I'm surprised this bug hasn't been spotted sooner.  This should be `if meta.isplots_S4 >= 1:`

src/eureka/S4_generate_lightcurves/s4_genLC.py
```diff
@@ -529,6 +587,9 @@ def genlc(eventlabel, ecf_path=None, s3_meta=None, input_meta=None):
                 # Plot the white-light light curve
                 if meta.isplots_S4 >= 3:
```
kevin218, line 588: This should also be `>= 1`

kevin218 (changes requested):

src/eureka/S3_data_reduction/background.py
erinmmay: I thought about this and came to the conclusion that I didn't want to mask bad data before the stddev here because then larger a stddev would indicate there were bad pixels in the column, resulting in fewer pixels for measuring background, resulting in maybe less accurate background removal. Happy to discuss this though.

taylorbell57: Rather than using just the standard deviation, I'd think instead using the standard error of the mean (aka `np.ma.std(values)/np.sqrt(np.sum(~values.mask))`) would strike a good balance. With the current setup, users might be concerned about seemingly noisy background measurements even though the bad pixels driving that noise were actually masked. And by dividing by the square-root of the number of unmasked points, you still capture the increased uncertainty in columns where the background measurement was likely a bit noisier since it was based on fewer measurements

PR #727, opened 2024-12-10
Allowing fixing/constraining spotcon to an input model file

taylorbell57:
This PR allows users to fix or constrain the starspot contrast parameter, `spotcon`, to values from an input model file in the same way that we allow the input of limb-darkening coefficients. I've found that the spot parameters can be fairly well constrained by broadband fits, but for spectroscopic fits I fix the spot size and location parameters to the results from the broadband fit and I use a ratio of PHOENIX models that matches the fitted `spotcon` value from my broadband fit. Since `fleck` only allows a single contrast for all spots, the `spotcon_file` for `fleck` will only use the first column, but if the user uses `starry` then there can be a separate column for each spot (if the user wants the same contrast for all spots, they can provide just a single column file and the code will just duplicate that for the other spots).

### DIFF
--- a/docs/source/ecf.rst
+++ b/docs/source/ecf.rst
@@ -933,15 +933,31 @@ The following three parameters control the use of pre-generated limb darkening c
 
 use_generate_ld
 ^^^^^^^^^^^^^^^
-If you want to use the generated limb-darkening coefficients from Stage 4, use exotic-ld or spam. Otherwise, use None. Important: limb-darkening coefficients are not automatically fixed, change the limb darkening parameters to 'fixed' in the .epf file if they should be fixed instead of fitted! The limb-darkening laws available to exotic-ld and spam are linear, quadratic, 3-parameter and 4-parameter non-linear.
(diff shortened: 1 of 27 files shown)


docs/source/ecf.rst
```diff
@@ -945,15 +945,31 @@ The following three parameters control the use of pre-generated limb darkening c
...
 
 ld_file_white
 ^^^^^^^^^^^^^
-The same type of parameter as ld_file, but for the limb-darkening coefficients to be used for the white-light fit. This parameter is required if ld_file is not None and any of your EPF parameters are set to white_free or white_fixed. If no parameter is set to white_free or white_fixed, then this parameter is ignored.
+The same type of parameter as ``ld_file``, but for the limb-darkening coefficients to be used for the white-light fit. This parameter is required if ``ld_file`` is not None and any of your EPF parameters are set to ``white_free`` or ``white_fixed``. If no parameter is set to ``white_free`` or ``white_fixed``, then this parameter is ignored.
+
+recenter_ld_prior
```
kevin218, line 958: Is there any reason we can't re-center priors when using the Exotic-ld coefficients?

kevin218 (changes requested):

docs/source/ecf.rst
taylorbell57: That's exactly what this parameter was made to do. If you're fixing your LD parameters to any LD model, then you don't need this parameter. But if you want to use a Gaussian prior around any LD model, then you need to set this parameter to True

taylorbell57: Oh I see what you're saying now - let me fix that

taylorbell57: Okay, should be good now. Sorry about the confusion - reading/writing this from my phone on the airplane

kevin218 (approved):
Enjoy your trip. Happy holidays!

PR #728, opened 2024-12-16
Stage 4cal (Calibrated Stellar Spectra)

kevin218:
This PR adds a new step to the Eureka! pipeline.  Stage 4cal produces calibrated stellar spectra over two regions (typically in and out of transit/eclipse) with uncertainties.  In addition, I've added the following:
1. A new folder to Demos that steps through Stages 2, 3, and 4cal using NIRSpec G395H as an example
2. A suitable description of the new ECF parameters for the website
3. New tests that include a NIRSpec file from actual data, thus replacing the simulated FITS file.

PR #736, opened 2025-01-13
Fixing bug in light travel time calculation for circular orbit

### LINKED ISSUE #735: [Bug]: Incorrect light travel time correction for circular orbits
### FAQ check

### Instrument

### What happened?

Bug discovered in juliet, but the relevant code is functionally equivalent in Eureka! For an eccentric orbit, the code returns a light travel time delay of +51 seconds for eclipses of WASP-17b and HAT-P-26b, as expected (2*a*Rs/c). For a circular orbit, the code returns an LTT delay of _minus_ 10 seconds for HAT-P-26b and _minus_ 30 seconds for WASP-17b, which is unphysical and inconsistent with the eccentric case.

In `AstroModel.correct_light_travel_time`, the code path for a circular orbit has a `sin(pl_params.inc)`, which assumes inclination is in radians. But I _believe_ `pl_params.inc` is in degrees. `batman` and `KeplerOrbit` assume inclination is in degrees and multiply by `np.pi/180` before use in trigonometric functions. Feeding degrees rather than radians to the `cos` function multiplies the LTT delay by a factor between -1 and 1. Changing the circular orbit code to `sin(pl_params.inc*np.pi/180)` yields an LTT delay of +51 seconds, consistent with expectations and the eccentric orbit case.

### Error traceback output

_No response_

_No response_

_No response_

_No response_

### Code of Conduct

taylorbell57:
Resolves #735

There was previously an error in the units used within a sin function call, where degrees were accidentally passed in instead of radians.

### DIFF
--- a/src/eureka/S5_lightcurve_fitting/models/AstroModel.py
+++ b/src/eureka/S5_lightcurve_fitting/models/AstroModel.py
@@ -537,13 +537,6 @@ def true_anomaly(model, t, lib=np, xtol=1e-10):
     -------
     ndarray
         The true anomaly in radians.
-
-    Notes
-    -----
-    History:
-
-    - March 2023 Taylor Bell
-        Based on Bell_EBM code, but modified to enable theano code.
     """
     return 2.*lib.arctan(lib.sqrt((1.+model.ecc)/(1.-model.ecc)) *
                          lib.tan(eccentric_anomaly(model, t, lib,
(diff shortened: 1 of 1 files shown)


kevin218 (approved):

kevin218:
I think `pytest` (without PyMC3) will continue to fail until #728 is merged into main.

I'm still working to resolve the ongoing [dependency issue with PASTASOSS](https://github.com/spacetelescope/pastasoss/issues/13).

Issue #743, opened 2025-02-11
[Bug]: Limb darkening for joint fits not using different values for the different visits/channels


erinmmay:
### FAQ check

- [x] Yes, I checked the FAQ and my question has not been addressed.

### Instrument

Light curve fitting (Stages 4-6)

### What happened?

In AstroModel limb darkening does not account for the channel number when doing multwhite fits. As a result, when fixing to the exotic-ld calculated values, both "visits" end up using the same values. This impacts joint fits between G395H NRS1/NRS2 and NIRISS SOSS Order 1 and Order 2. 

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #744, opened 2025-02-11
fixing LD bug for joint fits

erinmmay:
addressing bug #743 

I believe this is how we want to do it, limb darkening should have channel id appended regardless of if it's a free fit since we can fix them to different values.

### DIFF
--- a/src/eureka/S5_lightcurve_fitting/models/AstroModel.py
+++ b/src/eureka/S5_lightcurve_fitting/models/AstroModel.py
@@ -306,7 +306,10 @@ class PlanetParams():
         ld_func = ld_profile(self.limb_dark)
         len_params = len(inspect.signature(ld_func).parameters)
         coeffs = ['u{}'.format(n) for n in range(1, len_params)]
-        self.u = [getattr(self, coeff) for coeff in coeffs]
+        for i, item0 in enumerate(coeffs):
+            item0 += self.channel_id
+            coeffs[i] = item0
+        self.u = [getattr(parameterObject, coeff).value for coeff in coeffs]
         if self.limb_dark == '4-parameter':
             self.limb_dark = 'nonlinear'
         elif self.limb_dark == 'kipping2013':


erinmmay:
also adding a minor fix for fleck. Currently if you set the star rotation (spotrot) in the s5 fit params epf file then fleck_fast is never set to anything and the code crashes, this adds the lines to set fleck_fast when spotrot is defined in the epf

taylorbell57:
I haven't yet confirmed where this bug arises, but as it stands I don't think this patch is quite right. If there's a bug (which I'm certainly willing to believe, though it hasn't arisen for non-multwhite fits), it must be in the interactions between these snippets of code:

https://github.com/kevin218/Eureka/blob/a8b3a8a7554c31fa66ee3a39b1c233581a23ed02/src/eureka/S5_lightcurve_fitting/models/AstroModel.py#L146-L171

https://github.com/kevin218/Eureka/blob/a8b3a8a7554c31fa66ee3a39b1c233581a23ed02/src/eureka/S5_lightcurve_fitting/models/AstroModel.py#L306-L309

https://github.com/kevin218/Eureka/blob/a8b3a8a7554c31fa66ee3a39b1c233581a23ed02/src/eureka/S5_lightcurve_fitting/models/BatmanModels.py#L38-L73

If the bug is specifically with loading-in exotic-ld results (and/or custom ld files), then my pretty strong suspicion is that the bug is in the last of those three snippets (from `BatmanModels.py` or the equivalent chunks of code for `fleck` or `starry`). I won't have time to look into this more today, but I can try to investigate further tomorrow if you don't beat me to it.

erinmmay:
I add tons of print statements to confirm that the code you reference in BatmanModels sets things up in properly when reading from the exotic values and that the bug arises when BatmanModels.py calls AstroModel/PlanetParams to initialize the planet. AstroModel line 308/309 was previously only looking for u1/u2, never u1_ch1/u2_ch1 and the result was that both visits used the same limb darkening (confirmed with the print statements, visually with the plots, and by switching the order of the visits to see which LD values were used in that case)

taylorbell57:
I really would've thought that the ld parameters would be populated with the values for ch1 based on the snippet below which is used to set (in the pretend case of some parameter called `par`) the value of `PlanetParams.par` to the value of `parameterObject.par_ch#`:

https://github.com/kevin218/Eureka/blob/a8b3a8a7554c31fa66ee3a39b1c233581a23ed02/src/eureka/S5_lightcurve_fitting/models/AstroModel.py#L149-L155

It is the same snippet that is used for every other variable, with the only exception being that lower down there's a snippet that allows ld parameters and spot parameters to be the same between different planets.

taylorbell57:
The lines to which you are referring:

https://github.com/kevin218/Eureka/blob/a8b3a8a7554c31fa66ee3a39b1c233581a23ed02/src/eureka/S5_lightcurve_fitting/models/AstroModel.py#L308-L309

use `self.u1` which should've been set to the value of `parameterObject.u1_ch1` by those lines I just linked above.

erinmmay:
yeah unfortunately they are not set to parameterObject.u1_ch1. self.u1 returns the actual u1 value. If there's a different way you want to fix this happy to do that instead, but as it stands it's not working as expected.

taylorbell57:
Ahh yeah, that'd do it! Honestly maybe we don't need that `'free'` check at all since its already wrapped in a `KeyError` check, though maybe we still do (worth investigating).

erinmmay:
On my end it says that's already enabled.

taylorbell57:
Okay thanks, must just be an issue with the online GitHub Codespaces then - I'll try using my local git installation later

taylorbell57:
@erinmmay, have you had the chance to test my edits to this PR? I'd like to get this patch merged if it's good-to-go

taylorbell57:
@kevin218, do you want to review this PR - it works for me, but would be good to get a second set of eyes on this

Issue #746, opened 2025-02-11
[Bug]: GP Packages not properly specified


jbrande:
### FAQ check

- [x] Yes, I checked the FAQ and my question has not been addressed.

### Instrument

Light curve fitting (Stages 4-6)

### What happened?

It looks like the GPModel "gp_code_name" parameter isn't properly specified in s5_fit.py (line 919, gp_code instead of gp_code_name), so the GP package always gets defaulted to "celerite", rather than the actual specified package in the S5 ECF (e.g. "george"). 

Renaming the parameter in s5_fit.py to gp_code_name doesn't appear to break celerite, but I haven't tested george yet.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #747, opened 2025-02-11
Fix GP model specification in s5_fit.py

jbrande:
Handling issue #746 and further GP testing, if needed.


jbrande:
I couldn't get george to work at first, but I think I've diagnosed that now. There was another issue in the GPModel eval(), where george GPs would end up with evaluated models of different lengths than the time axis, caused by a minor typo neglecting to pass the mask through to the kernel input in the predict() step. George is still painfully slow, though, so this might not actually help anyone, lol.

PR #757, opened 2025-04-16
Initial upload of optimizers.

astroRez:
Initial upload of MIRI-LRS, NIRSpec-PRISM, NIRSpec-G395H, and NIRISS-SOSS parametric optimizers.


demos/JWST/run_eureka_optimization_niriss_SOSS.py
```diff
@@ -0,0 +1,1177 @@
...
+        s4_meta_GA.bg_thresh = best["bg_thresh"]
+        s4_meta_GA.bg_hw = best["bg_hw"]
+        s4_meta_GA.spec_hw = best["spec_hw"]
+        s4_meta_GA.bg_deg = best["bg_deg"]
+        s4_meta_GA.bg_method = best["bg_method"]
+        s4_meta_GA.p3thresh = best["p3thresh"]
+        s4_meta_GA.median_thresh = best["median_thresh"]
+        s4_meta_GA.window_len = best["window_len"]
+        s4_meta_GA.p7thresh = best["p7thresh"]
+        s4_meta_GA.sigma = best["sigma"]
```
kevin218, line 927: I think we can use a `for` loop to shrink this file from 1000 lines to 100 lines.  Assuming we have a list of parameters that we want to loop over (e.g., params = ["dqmask", "expand", ...]), we can do the following:
`for i, param in enumerate(params):`
`    s3_meta_GA = S3MetaClass(ecf_path, s3_ecffile, eventlabel)`
`    s4_meta_GA = S4MetaClass(ecf_path, s4_ecffile, eventlabel)`
`    s3_meta_GA.xwindow = [best["xwindow_LB"], best["xwindow_UB"]]`
`    s3_meta_GA.ywindow = [best["ywindow_LB"], best["ywindow_UB"]]`
`    s4_meta_GA.nspecchan = 1`
`    for j in range(i):`
`        setattr(s3_meta_GA, params[j], best[params[j]])`
`        setattr(s4_meta_GA, params[j], best[params[j]])`
`    # Perform parametric sweep`
`    best_param, best_fitness_value = optimizers.parametric_sweep_S4(`
`                objective_funcs.generic, param, bounds_box_width, ...)`
`    best[param] = best_param[0]`

demos/JWST/run_eureka_optimization_niriss_SOSS.py
```diff
@@ -0,0 +1,1177 @@
...
+    directories = [(outputdir_WLC_S3, "S3"), (outputdir_WLC_S4, "S4")]
+    best_directory_white = os.path.join(outputdir_optimization, "White")
+
+    copy_files_and_subfolders_to_new_directory(directories, best_directory_white)
+
+    # Plot Optimization History
+    keys_plot = list(history_fitness_score.keys())
+    values_plot = list(history_fitness_score.values())
+
+    plt.figure(figsize=(12, 6))
```
kevin218, line 1166: Use a specific figure number (e.g., plt.figure(100, figsize=(12, 6)) and then use that in the file naming scheme ("Fig100_OptimizationHistory.png").

kevin218 (changes requested):
I've reviewed a few files and have made suggestions on how to economize the code (i.e. reduce duplication).  I'm happy to chat more after you review my suggestions.

kevin218:
Here are some additional suggestions based on my experiences in playing with the code:

- Move the optimizer demo files into their own folder within `demo`.
- Convert the .txt files to .ecf files and use the built-in ECF reader.
- Create a new folder (e.g., src/eureka/optimizer) and place files there.
- Create `opt_meta.py` and set all default values there.  See `s3_meta.py` as an example for how to define global default values and instrument-specific default values.
- I think we can delete `spec_hw_selection` and `bg_hw_selection` in favor of using default values.
- Make sure to define the functions of both `manual` and `auto` in the comments.
- In the ECF, allow users to specify a list of parameters that they wish to optimize.  If set to None, the default will be to optimize all possible parameters.
- Should we start by optimizing `bg_hw` and `spec_hw` first?
- Try to create as many generic functions as possible to cut down on repeating code.
- Use Eureka!'s log capabilities instead of print statements.  This will mean creating a log file in a folder that doesn't get deleted.
- When reporting the best parameter and fitness values, be sure to specify which parameter was being optimized.
- Move all plotting functions to `plots_opt.py` and follow the figure naming scheme (ask me for details).

taylorbell57:
This is on my to-do list and I haven't forgotten it, FYI - I've just been swamped with research travel, a week of sick leave, and a week of personal vacation

src/eureka/S1_detector_processing/s1_meta.py
```diff
@@ -36,6 +36,10 @@ def __init__(self, folder=None, file=None, eventlabel=None, **kwargs):
         # Set a default data file suffix
         self.suffix = getattr(self, 'suffix', 'uncal')
 
+        # Set optimization flag
+        self.isopt_S1 = getattr(self, 'isopt_S1', False)
+        self.isopt_S3 = getattr(self, 'isopt_S3', False)
```
taylorbell57: Why set this default in the S1 meta?

src/eureka/lib/manageevent.py
```diff
@@ -147,7 +147,7 @@ def findevent(meta, stage, allowFail=False):
     meta : eureka.lib.readECF.MetaClass
         The new meta object for the current processing.
     stage : str
-        The previous stage (e.g. "S2" for Stage 3).
+        The previous stage (e.g. "S3" for Stage 3).
```
taylorbell57: No, S2 would be the previous stage for Stage 3

taylorbell57 (changes requested):
I haven't gotten into reviewing the major edits of this PR yet, but here's some preliminary feedback

demos/JWST/Optimizer/run_eureka_optimization.py
```diff
@@ -0,0 +1,40 @@
...
+import eureka.optimizer.S1opt_optimizer as s1opt
+import eureka.optimizer.S3opt_optimizer as s3opt
+
+"""
+Eureka! Optimization
+--------------------
+
+Description:
+This script is designed to optimize the Stage 1,3, and 4 ECF parameters for
+JWST time-series observations.
```
taylorbell57: How time consuming would it be to add support for running/skipping flat-field in Stage 2?

taylorbell57 (changes requested):
Here are a couple more comments on some basic aspects before I dive into the bigger parts of this PR on Thursday/Friday this week.

Please also note that there will need to be a bunch of additions to the docs to describe these new features and plots.

src/eureka/lib/manageevent.py
```diff
@@ -174,13 +174,13 @@ def findevent(meta, stage, allowFail=False):
...
     for file_suffix in ['*_Meta_Save.dat', '*SpecData.h5']:
-        newfnames = glob.glob(meta.inputdir+stage+'_'+meta.eventlabel +
+        newfnames = glob.glob(meta.inputdir+os.sep+stage+'_'+meta.eventlabel +
                               file_suffix)
 
         if len(newfnames) == 0:
             # There were no metadata files in that folder, so let's see if
             # there are in children folders
-            newfnames = glob.glob(meta.inputdir+'**'+os.sep+stage+'_' +
+            newfnames = glob.glob(meta.inputdir+os.sep+'**'+os.sep+stage+'_' +
```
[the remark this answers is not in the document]
kevin218: `inputdir` is either user defined or as follows in s3_meta.py: `self.inputdir = getattr(self, 'inputdir', 'Stage2')`. The formatting is similar for other stages.  While I support minimizing the use of `os.sep`, I feel this proposed change is outside the scope of this PR.

src/eureka/lib/util.py
```diff
@@ -73,8 +78,6 @@ def readfiles(meta):
                              f'{meta.filename} to point to the folder '
                              f'containing the "{meta.suffix}.fits" files.')
 
-    meta.segment_list = np.array(sn.sort_nicely(meta.segment_list))
-
```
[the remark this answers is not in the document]
kevin218: The deleted line was a duplicate from Line 68.  There's no need to call the sort function twice on the same list.

demos/JWST/Optimizer/run_eureka_optimization.py
kevin218: Adding a Stage 2 optimization would be quite a bit of work for this PR, but we could explore that option in a subsequent PR.

src/eureka/lib/manageevent.py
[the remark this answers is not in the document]
taylorbell57: `inputdir` is regularized within readECF here:
https://github.com/kevin218/Eureka/blob/764a234e3a965e8a936df4b1e5f1a2328fec5635/src/eureka/lib/readECF.py#L239-L253

demos/JWST/Optimizer/run_eureka_optimization.py
taylorbell57: Can you open an Issue to look into that so we don't forget

demos/JWST/Optimizer/run_eureka_optimization.py
```diff
@@ -0,0 +1,40 @@
...
+
+Inputs:
+- Loaded from an input text file, 'optimizer_inputs_nirspec_PRISM.txt'.
+- Parameters not specified for optimization in the optimizer input text file will assume ECF values by default.
```
taylorbell57: This appears to be outdated/confusing

demos/JWST/Optimizer/S1opt_template.ecf
```diff
@@ -0,0 +1,33 @@
+# Eureka! Control File for Optimizer stage
+
+# Optimizer Documentation: https://eurekadocs.readthedocs.io/en/latest/ecf.html#optimizer-ecf
```
taylorbell57: There is no docs source for this page yet

demos/JWST/Optimizer/S1opt_template.ecf
```diff
@@ -0,0 +1,33 @@
+# Eureka! Control File for Optimizer stage
```
taylorbell57: Need to document somewhere that this file needs to have the same event label as all the other ECFs (took me a moment to figure that out)

demos/JWST/Optimizer/run_eureka_optimization.py
```diff
@@ -0,0 +1,40 @@
+import os
+import eureka.optimizer.S1opt_optimizer as s1opt
+import eureka.optimizer.S3opt_optimizer as s3opt
+
```
taylorbell57, line 5: We should still include lines like these which are at the top of our other run_eureka.py files:
```
import eureka.lib.plots

# Set up some parameters to make plots look nicer.
# You can set usetex=True if you have LaTeX installed
eureka.lib.plots.set_rc(style='eureka', usetex=False, filetype='.png')
```

demos/JWST/Optimizer/S1opt_template.ecf
```diff
@@ -0,0 +1,33 @@
...
+# Commenting out this line will use all single parameters listed below.
+# Single parameter options: jump_rejection_threshold, expand_mask, bg_deg, bg_method, p3thresh, window_len
+# Double parameter options: any combination of the above, joined by two underscores (e.g., expand_mask__p3thresh)
+params_to_optimize  ['jump_rejection_threshold', 'expand_mask', 'p3thresh']
+
+# Range for parameter values to optimize
+# Only specify if overriding the default ranges in S1opt_meta.py
+# For single parameters, use format: bounds_<parameter_name> = range(min, max+1) or np.arange(min, max+step, step) if step != 1
+# For double parameters, use format: bounds_<parameter1>__<parameter2> = [range(min1, max1+1), range(min2, max2+1)]
+bounds_expand_mask  range(5,10)         # Example: optimize BG inner edge between 5 and 9 pixels
```
taylorbell57: Missing values for the other parameters listed in the `params_to_optimize`

src/eureka/lib/manageevent.py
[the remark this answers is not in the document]
taylorbell57: I made some edits directly to this PR to simplify the handling of topdir, inputdir, and outputdir while I was trying to test the code locally

src/eureka/optimizer/S1opt_optimizer.py
```diff
@@ -0,0 +1,315 @@
...
+
+    for p in s1opt_meta.params_to_optimize:
+        s1opt_meta, log, history, best = optimize(s1opt_meta, log, history,
+                                                  best, p, eventlabel,
+                                                  ecf_path, 1)
+
+    # Save the best dictionary to a pickle file
+    with open(os.path.join(s1opt_meta.outputdir, "best_params.pkl"),
+              "wb") as f:
+        pickle.dump(best, f)
```
taylorbell57: Can we use a better file format than a .pkl file

src/eureka/optimizer/S3opt_optimizer.py
```diff
@@ -0,0 +1,312 @@
...
+                                                     best_s4, p, eventlabel,
+                                                     ecf_path, 4,
+                                                     best_s3=best_s3)
+
+    # Combine best parameters from both stages
+    best = {**best_s3, **best_s4}
+    # Save the best dictionary to a pickle file
+    with open(os.path.join(s3opt_meta.outputdir, "best_params.pkl"),
+              "wb") as f:
+        pickle.dump(best, f)
```
taylorbell57: Can we use a better file format than a .pkl file

demos/JWST/Optimizer/S1opt_template.ecf
taylorbell57: I don't like the name `bounds_...` since these aren't bounds that we're specifying but rather all the values that will be tried

demos/JWST/Optimizer/S1opt_template.ecf
```diff
@@ -0,0 +1,33 @@
+# Eureka! Control File for Optimizer stage
+
+# Optimizer Documentation: https://eurekadocs.readthedocs.io/en/latest/ecf.html#optimizer-ecf
+
+# Fitness Scaling
+# Use these settings to override the default fitness scaling factors.
+scaling_MAD_spec    0.01            # Optimizer scaling for 2D MAD value
+scaling_MAD_white   1.0             # Optimizer scaling for WLC MAD value
```
taylorbell57, line 10: Do we need to define both, or can we just assume the white MAD is always 1.0?

demos/JWST/Optimizer/S1opt_template.ecf
taylorbell57: Also, note that this utility also works for boolean variables or other non-discreet values by just passing in any iterable (e.g., a list)

taylorbell57: Should one include or exclude the value(s) in the S1 ECF file? If so, it would seem that that setup would get called twice (including the initial run), while if not then I'd worry whether the S1 ECF value could be "selected" by the current algorithm

src/eureka/optimizer/S1opt_optimizer.py
```diff
@@ -0,0 +1,315 @@
...
+        s3_spec, s3_meta = s3.reduce(eventlabel, input_meta=s3_meta)
+        s4_spec, s4_lc, s4_meta = s4.genlc(eventlabel, input_meta=s4_meta,
+                                           s3_meta=s3_meta)
+        # Record initial fitness score
+        history["final_run"] = (
+            s1opt_meta.scaling_MAD_spec * s4_meta.mad_s4 +
+            s1opt_meta.scaling_MAD_white * s4_meta.mad_s4_binned[0])
+        log.writelog(f"Final fitness value: {history["final_run"]}")
+        log.writelog(f"Final white MAD: {s4_meta.mad_s4_binned[0]}")
+        log.writelog(f"Final spec MAD: {s4_meta.mad_s4}\n")
```
taylorbell57, line 129: Shouldn't the final run be one of the runs that were already completed? If so, can't we just cp/mv it rather than re-running the whole analysis?

src/eureka/optimizer/S1opt_optimizer.py
```diff
@@ -0,0 +1,315 @@
...
+    Parameters
+    ----------
+    eventlabel : str
+        The unique identifier for these data.
+    ecf_path : str; optional
+        The absolute or relative path to where ecfs are stored. Defaults to
+        None which resolves to './'.
+    initial_run : boolean; optional
+        Set to True to perform an initial run with default ECF parameters
+        before starting the optimization. Defaults to False.
```
taylorbell57: Defaults to `True` right now

src/eureka/optimizer/S1opt_optimizer.py
```diff
@@ -0,0 +1,315 @@
...
+                log.writelog(f"Could not create bounds for parameter {p}. " +
+                             "Please manually specify bounds in ECF. " +
+                             "Skipping...")
+                return s1opt_meta, log, history, best
+        log.writelog(f"Optimizing parameters {p} over bounds: {bounds}")
+        log.writelog(f"Initial parameter values: {init_vals}")
+    else:
+        log.writelog(f"No default bounds exist for parameter {p}. " +
+                     "Please manually specify bounds in ECF. Skipping...")
+        return s1opt_meta, log, history, best
```
taylorbell57: Similarly to the user-facing side of things, I don't like the name `bounds` since they're not bounds but a collection of parameter values to try

src/eureka/optimizer/objective_funcs.py
```diff
@@ -0,0 +1,167 @@
...
+    ----------
+    val : float
+        The variable value to be evaluated.
+    meta : eureka.lib.readECF.MetaClass
+        The metadata object.
+    stage : int
+        The stage number indicating which stage's parameters to
+        optimize.
+    run_S3 : boolean; optional
+        If True, run Stage 3.
```
taylorbell57: What's the motivation for treating S3 specially here?

src/eureka/optimizer/optimizers.py
```diff
@@ -0,0 +1,251 @@
+import numpy as np
+import eureka.optimizer.objective_funcs as of
+
+
+def sweep_range_single(bounds_var, meta, log, stage, **kwargs):
+    """
+    Optimize single parameter using parametric sweep.  This function returns
+    the best parameter and resulting fitness score after sweeping through
+    every value within the specified bounds .
```
taylorbell57: This function is never used anywhere - can we just remove it?

src/eureka/optimizer/S1opt_optimizer.py
```diff
@@ -0,0 +1,315 @@
...
+        s1_meta.params[key] = value
+        setattr(s1_meta, key, value)
+
+    # Perform parametric sweep
+    if p == "spec_hw__bg_hw":
+        # Optimize both spec_hw and bg_hw simultaneously
+        # Require that spec_hw < bg_hw
+        best_param_value, best_fitness_value = optimizers.sweep_list_lt(
+            bounds, meta, log, stage, s1_meta=s1_meta, s2_meta=s2_meta,
+            s3_meta=s3_meta, s4_meta=s4_meta)
```
taylorbell57: We shouldn't be optimizing `spec_hw__bg_hw` in S1, right? Can we remove this?

src/eureka/optimizer/S1opt_optimizer.py
```diff
@@ -0,0 +1,315 @@
...
+            meta.scaling_MAD_spec * s4_meta.mad_s4 +
+            meta.scaling_MAD_white * s4_meta.mad_s4_binned[0])
+        log.writelog(f"Initial fitness value: {history["initial_run"]}")
+        log.writelog(f"Initial white MAD: {s4_meta.mad_s4_binned[0]}")
+        log.writelog(f"Initial spec MAD: {s4_meta.mad_s4}\n")
+
+    for p in s1opt_meta.params_to_optimize:
+        s1opt_meta, log, history, best = optimize(s1opt_meta, log, history,
+                                                  best, p, eventlabel,
+                                                  ecf_path, 1)
```
taylorbell57: Can we add some logging here so we can know how far along the optimization is?

src/eureka/optimizer/S3opt_optimizer.py
```diff
@@ -0,0 +1,312 @@
...
+                                                     best_s3, p, eventlabel,
+                                                     ecf_path, 3)
+
+    # Run optimization loop for Stage 4 parameters
+    best_s4 = {}
+    for i, p in enumerate(s3opt_meta.params_to_optimize_s4):
+        s3opt_meta, log, history, best_s4 = optimize(s3opt_meta, log, history,
+                                                     best_s4, p, eventlabel,
+                                                     ecf_path, 4,
+                                                     best_s3=best_s3)
```
taylorbell57, line 115: Can we add some logging here so we can know how far along the optimization is?

src/eureka/optimizer/optimizers.py
```diff
@@ -0,0 +1,251 @@
...
+        # Need to rerun Stage 3 first time around on Stage 4 optimization
+        # Otherwise, save time by not rerunning Stage 3 again
+        run_s3 = False if (i > 0 and stage == 4) else True
+        try:
+            fitness_value = of.single(val, meta, stage, run_s3, **kwargs)
+            if fitness_value < best_fitness_value:
+                # Update best fitness and parameters if current is better
+                best_fitness_value = fitness_value
+                best_param_value = val
+        except Exception as e:
```
taylorbell57, line 58: This is going to catch Ctrl+C keyboard interrupts. We should either just catch the expected exception(s), or not catch anything

taylorbell57: Same goes for all the other functions in this file

src/eureka/optimizer/objective_funcs.py
```diff
@@ -0,0 +1,167 @@
...
+
+    if meta.delete_intermediate:
+        if run_stage[1]:
+            shutil.rmtree(s1_meta.outputdir)
+        if run_stage[2]:
+            shutil.rmtree(s2_meta.outputdir)
+        if run_stage[3]:
+            shutil.rmtree(s3_meta.outputdir)
+        if run_stage[4]:
+            shutil.rmtree(s4_meta.outputdir)
```
taylorbell57: Not sure why, but these aren't currently deleting the intermediate folders for me

taylorbell57 (changes requested):

src/eureka/optimizer/S3opt_optimizer.py
taylorbell57, line 115: With how this is setup, if `delete_intermediate` is `True`, then the Stage 3 outputs are deleted before the Stage 4 optimization is run and the Stage 4 optimization fails

src/eureka/optimizer/S1opt_meta.py
```diff
@@ -0,0 +1,124 @@
...
+            if self.inst == 'miri':
+                self.set_MIRI_defaults()
+            elif self.inst == 'nircam':
+                self.set_NIRCam_defaults()
+            elif self.inst == 'nirspec':
+                self.set_NIRSpec_defaults()
+            elif self.inst == 'niriss':
+                self.set_NIRISS_defaults()
+            elif self.inst == 'wfc3':
+                self.set_WFC3_defaults()
```
taylorbell57, line 70: At present, only `set_NIRSpec_defaults` is defined and all the others result in a crash

src/eureka/optimizer/S3opt_meta.py
```diff
@@ -0,0 +1,223 @@
...
+        }
+
+        for key, default in defaults.items():
+            setattr(self, key, getattr(self, key, default))
+        # self.bounds_spec_hw__bg_hw = getattr(self, 'bounds_spec_hw__bg_hw',
+        #                                      [self.bounds_spec_hw,
+        #                                       self.bounds_bg_hw])
+        # self.bounds_mad_sigma__mad_box_width = getattr(self,
+        #         'bounds_mad_sigma__bg_hw',[self.bounds_mad_sigma,
+        #         self.bounds_mad_box_width])
```
taylorbell57: Remove the commented-out code if no longer needed

demos/JWST/Optimizer/S1opt_template.ecf
kevin218: I think we want both to allow for one of the values to be set to zero.

demos/JWST/Optimizer/S1opt_template.ecf
kevin218: We only need to specify if overriding the default ranges in S1opt_meta.py

kevin218: I went with `sweep_`

kevin218: When performing parameter sweeps, it's expected that one of the runs will be identical to the initial run.

Right now, there are no guard rails against exploring a range of parameter values that doesn't include the default value in the ECF.  This could yield a strange result, where the "optimal" settings are worse than the initial settings, but that will be obvious in the final figure.  I've experience this issue for parameters like `mad_box_width` that  default to `range(11, 52, 10)` and the ECF value is, for example, 15.  In this case, the responsibility falls on the user to perform a second optimization with a smaller step size (e.g., `range(11, 21, 1)`).

src/eureka/optimizer/S1opt_optimizer.py
kevin218: Yes and no.  Optimizer sweeps may not store all of the desired outputs in order to reduce storage on the user's HD.  This final run produces all requested figures and save files.

demos/JWST/Optimizer/run_eureka_optimization.py
kevin218: I can't think of anything to optimize in Stage 2.  Is this really necessary?

src/eureka/optimizer/S3opt_optimizer.py
kevin218: The log is already being updated inside `optimize()` with text that reads: `Optimizing parameter jump_rejection_threshold over bounds: range(4, 15)`

src/eureka/optimizer/optimizers.py
kevin218: Not catching anything isn't a viable solution.  We can't have a single error (like `spec_hw` being too small for a Gaussian fit) break the entire optimization process.  At the same time, we can't predict every type of error that might arise. I'm open to suggestions for improvements.

src/eureka/optimizer/objective_funcs.py
kevin218: Strange.  I'm not sure why it's not working for you.  Are the files inside deleted?  Maybe there's an issue with permissions.

src/eureka/optimizer/S3opt_optimizer.py
kevin218: Yeah, that was an annoying bug to resolve.  The solution I came up with is to have the Stage 4 optimizer run Stage 3 using the best Stage 3 parameters and then optimize the Stage 4 parameters.

src/eureka/optimizer/objective_funcs.py
kevin218: When optimizing Stage 4 parameters, we don't need to rerun Stage 3 at each step.  So this boolean effectively speeds up the optimizer.

src/eureka/optimizer/objective_funcs.py
kevin218: Maybe there's an issue with `meta.delete_intermediate` being defined in Stage 1opt and not getting overwritten in Stage 3.

demos/JWST/Optimizer/run_eureka_optimization.py
taylorbell57: Turning on/off Stage 2's flat field step is the only thing I can think of, so it'd be a very small addition to allow full S1-4 optimization

demos/JWST/Optimizer/S1opt_template.ecf
taylorbell57: I'm not sure that I have a clean code change request here then, but noting this in the documentation and/or template files would be a good idea

src/eureka/optimizer/S1opt_optimizer.py
taylorbell57: Fair enough

src/eureka/optimizer/objective_funcs.py
taylorbell57: Probably worth noting that in the docstring to remind the future versions of ourselves and others

src/eureka/optimizer/S3opt_optimizer.py
taylorbell57: With some more trial runs I did start to notice some of the logging, but some concerns I still had were:

1. It's not always clear which parameter value the loop is currently on (either isn't logged or is lost in the deluge of other logs)
2. It's not always clear what fraction of the values for that parameter have been tested so far (is it half-way though, starting the first value, etc.)

src/eureka/optimizer/optimizers.py
taylorbell57: I've just learned that `except Exception as e:` doesn't actually catch Ctrl+C interrupts (which inherits from `BaseException` and not `Exception`), so my concerns are reduced. We should definitely avoid bare `except:` statements though which catch everything.

src/eureka/optimizer/objective_funcs.py
taylorbell57: From what I recall the files weren't deleted inside, but it has been a while so I'm not totally sure (let me know if you want me to test something locally again)

src/eureka/optimizer/S3opt_optimizer.py
kevin218: I've added additional log information that tracks the current value being evaluated.  Given the deluge of information being output in the terminal, I don't think there's an easy way to track progress that way.

src/eureka/optimizer/optimizers.py
kevin218: Good to know! There are a fair number of `except:` cases in Eureka!, but none in this PR.  Feel free to create an issue to track this proposed change.

demos/JWST/Optimizer/run_eureka_optimization.py
```diff
@@ -0,0 +1,45 @@
...
+# Set up some parameters to make plots look nicer.
+# You can set usetex=True if you have LaTeX installed
+eureka.lib.plots.set_rc(style='eureka', usetex=False, filetype='.png')
```
taylorbell57, line 8: We're missing an import statement needed to be able to call this line

demos/JWST/Optimizer/run_eureka_optimization.py
```diff
@@ -0,0 +1,45 @@
...
+--------------------
+
+Description:
+This script is designed to optimize the Stage 1, 3, and 4 ECF parameters for
+JWST time-series observations.
+
+Inputs:
+- Loaded from a Eureka! control file, see S1opt_template.ecf/S3opt_template.ecf.
+- Parameters not specified for optimization will adopt the values listed in the
+    corresponding ECF.
```
taylorbell57: This could use a bit more clarification since there's two types of S1 ECFs for example (S1opt and S1)

demos/JWST/Optimizer/S3opt_template.ecf
```diff
@@ -0,0 +1,42 @@
...
+
+# Range for parameter values to optimize
+# Only specify if overriding the default values in S3opt_meta.py
+# For single parameters, use format: sweep_<parameter_name> = range(min, max+1) or np.arange(min, max+step, step) if step != 1
+# For double parameters, use format: sweep_<parameter1>__<parameter2> = [range(min1, max1+1), range(min2, max2+1)]
+sweep_bg_thresh    np.arange(3,5.5,0.5)  # Example: optimize BG threshold between 3 and 5 sigma, in steps of 0.5
+
+# Diagnostics
+isplots_S3opt       1               # Generate none (0), few (1), some (3), or many (5) figures (Options: 1 - 5)
+delete_intermediate False           # If True, intermediate directories will be deleted after each optimization step
```
taylorbell57, line 32: Shouldn't this also default to True

tests/S3opt_ecfs/S3opt_NIRSpec.ecf
```diff
@@ -0,0 +1,42 @@
...
+# For single parameters, use format: sweep_<parameter_name> = range(min, max+1) or np.arange(min, max+step, step) if step != 1
+# For double parameters, use format: sweep_<parameter1>__<parameter2> = [range(min1, max1+1), range(min2, max2+1)]
+sweep_spec_hw__bg_hw      [range(4,6), range(4,7)]
+sweep_bg_thresh            np.arange(4,5.5,0.5)
+sweep_mad_sigma            range(6,8)
+sweep_sigma__box_width     [range(3,7,2), range(7,16,6)]
+
+# Diagnostics
+isplots_S3opt       1               # Generate none (0), few (1), some (3), or many (5) figures (Options: 1 - 5)
+delete_intermediate False           # If True, intermediate directories will be deleted after each optimization step
```
taylorbell57: Shouldn't this be True?

tests/S3opt_ecfs/S3opt_NIRSpec.ecf
```diff
@@ -0,0 +1,42 @@
...
+# Use these settings to override the default fitness scaling factors.
+scaling_MAD_spec    0.01            # Optimizer scaling for 2D MAD value
+scaling_MAD_white   1.0             # Optimizer scaling for WLC MAD value
+
+# List of single parameter options to optimize in Stage 3:
+#   dqmask, bg_thresh, bg_hw, bg_deg, bg_method, p3thresh, spec_hw, median_thresh, window_len, p7thresh
+# Double parameter options: any combination of the above, joined by two underscores (e.g., bg_hw__p3thresh)
+# Special cases: spec_hw__bg_hw (requires spec_hw < bg_hw)
+params_to_optimize_s3  ['spec_hw__bg_hw', 'bg_thresh']
+# params_to_optimize_s3  []
```
taylorbell57: Remove old, commented out line

docs/source/ecf.rst
```diff
@@ -257,6 +257,79 @@ In short this weights each pixel, :math:`i`, within a slope following :math:`w_i
 ``custom``: As with default, except a custom SNR to :math:`P` lookup table can be defined through the ``default_ramp_fit_custom_snr_bounds`` and ``default_ramp_fit_custom_exponents`` (see example .ecf file).
 
 
+
+Stages 1/3/4 Optimizer
+---------------------
```
taylorbell57: This underline isn't the right length

docs/source/ecf.rst
```diff
@@ -257,6 +257,79 @@ In short this weights each pixel, :math:`i`, within a slope following :math:`w_i
...
+Scaling factor applied to the white light curve MAD value in the fitness function. Higher values prioritize band-integrated quality.
+
+params_to_optimize
+''''''''''''''''''
+List of parameters to optimize in Stage 1. Commenting out this line will use all single parameters listed below.  Single parameter options: jump_rejection_threshold, expand_mask, bg_deg, bg_method, p3thresh, window_len. Double parameter options: any combination of the above, joined by two underscores (e.g., expand_mask__p3thresh). Example:
+.. code-block:: python
+   params_to_optimize = ['jump_rejection_threshold', 'expand_mask', 'p3thresh']
+
+params_to_optimize_s3
+''''''''''''''''''''''''''
```
taylorbell57: This underline isn't the right length

docs/source/ecf.rst
```diff
@@ -257,6 +257,79 @@ In short this weights each pixel, :math:`i`, within a slope following :math:`w_i
...
+to optimize the background threshold between 3 and 5 sigma, in steps of 0.5
+
+params_to_optimize_s4
+'''''''''''''''''''''
+List of parameters to optimize in Stage 4. Single parameter options: mad_sigma, mad_box_width, sigma, box_width. Suggested double parameter options: mad_sigma__mad_box_width, sigma__box_width. Example:
+.. code-block:: python
+    params_to_optimize_s4 = ['mad_sigma__mad_box_width', 'sigma__box_width']
+
+sweep_<parameter_name>
+''''''''''''''''''''
```
taylorbell57: This underline isn't the right length

docs/source/ecf.rst
```diff
@@ -257,6 +257,79 @@ In short this weights each pixel, :math:`i`, within a slope following :math:`w_i
...
+   sweep_<parameter_name> = [3, 5, 6, 7]
+
+sweep_<parameter1>__<parameter2>
+''''''''''''''''''''''''''''''''
+Range for parameter values to optimize. Only specify if overriding the default values in the meta Python files. For double parameters, use the following format:
+.. code-block:: python
+   sweep_<parameter1>__<parameter2> = [range(min1, max1+1), range(min2, max2+1)]
+
+isplots_S1opt and isplots_S3opt
+''''''''''''''''''''''''''''''''
```
taylorbell57: This underline isn't the right length

docs/source/ecf.rst
```diff
@@ -257,6 +257,79 @@ In short this weights each pixel, :math:`i`, within a slope following :math:`w_i
 ``custom``: As with default, except a custom SNR to :math:`P` lookup table can be defined through the ``default_ramp_fit_custom_snr_bounds`` and ``default_ramp_fit_custom_exponents`` (see example .ecf file).
 
 
+
+Stages 1/3/4 Optimizer
+---------------------
+This tool allows users to optimize the parameter values of Stages 1, 3, and 4 by performing a parametric sweep over a specified range and evaluating the resulting spectra with a user-defined fitness function.  The optimizer starts with an intial run of a stage using the default parameter values, performs a sweep over the specified parameter values, and then completes a final run with the best parameter values from the sweep.
```
taylorbell57: Not a "user-defined fitness function", but fair to say that there's a user-defined weighting

docs/source/ecf.rst
```diff
@@ -257,6 +257,79 @@ In short this weights each pixel, :math:`i`, within a slope following :math:`w_i
...
+
+The fitness function is defined in the meta Python files and can be customized by the user.  The default fitness function is a weighted combination of the 2D MAD and the white light curve MAD.  The optimizer will generate a plot showing how the fitness function improves with each parameter sweep.  In the rare event that the final fitness value is worse than the initial value, users should consider trying smaller step sizes or a different range that include the default parameter values.
```
taylorbell57: Again, the fitness function isn't really defined or really that customizable from the "meta Python files" (which should probably be renamed to ECFs), since it's hard-coded that:
```
fitness_value = (
    meta.scaling_MAD_spec * s4_meta.mad_s4 +
    meta.scaling_MAD_white * s4_meta.mad_s4_binned[0])
```

It'd be cool to eventually allow hot-swappable fitness functions, but that's definitely not needed for this PR. That'd be something we could add when we eventually re-org the whole pipeline to be more like the jwst pipeline (in that you can override steps with custom functions or skip steps)

docs/source/ecf.rst
```diff
@@ -257,6 +257,79 @@ In short this weights each pixel, :math:`i`, within a slope following :math:`w_i
...
+
+In general, the optimizer should be used as a tool to explore the parameter space and understand how different parameters impact the resulting spectra.  Users may want to review the spectra generated from different parameter values in the sweep to get a better understanding of how each parameter impacts the results.  It is also important to keep in mind that the optimal parameter values may differ for different datasets, so it is recommended to use the optimizer for each new dataset to find the best parameter values.
```
taylorbell57: I'd make this wording/warning much stronger and very explicitly caution folks that this helps with parameter exploration and fine-tuning, but users still shouldn't treat Eureka! like a black-box, and they should critically investigate all of their outputs before and after optimization to make sure that they agree with the final tunings. And make it clear that the fitness metric is imperfect and could result in undesirable final outcomes in cases where there are nearby binary stars or other things like that.

The example I always give people is that just minimizing the Stage 4 MAD is risky because you can just set the Stage 4 sigma clipping to be 0.0001 sigma and throw away >99.99% of your data and get a super tiny MAD value, but that'll obviously give you horrible final results. So it's crucial that they consider what they're optimizing and whether the "optimal" results really are optimal

docs/source/ecf.rst
```diff
@@ -257,6 +257,79 @@ In short this weights each pixel, :math:`i`, within a slope following :math:`w_i
...
+
+scaling_MAD_spec
+''''''''''''''''
+Scaling factor applied to the 2D Median Absolute Difference (MAD) value in the fitness function. Higher values prioritize spectral quality.
```
taylorbell57: Make it clearer that this is applied to the un-binned, pixel-level spectra

docs/source/ecf.rst
```diff
@@ -257,6 +257,79 @@ In short this weights each pixel, :math:`i`, within a slope following :math:`w_i
...
+
+scaling_MAD_white
+'''''''''''''''''
+Scaling factor applied to the white light curve MAD value in the fitness function. Higher values prioritize band-integrated quality.
+
+params_to_optimize
```
taylorbell57: Nit-picky, but should this be `params_to_optimize_s1` since there's no obvious reason to favour S1 vs S3/S4 with a more basic name?

docs/source/ecf.rst
```diff
@@ -257,6 +257,79 @@ In short this weights each pixel, :math:`i`, within a slope following :math:`w_i
...
+''''''''''''''''''
+List of parameters to optimize in Stage 1. Commenting out this line will use all single parameters listed below.  Single parameter options: jump_rejection_threshold, expand_mask, bg_deg, bg_method, p3thresh, window_len. Double parameter options: any combination of the above, joined by two underscores (e.g., expand_mask__p3thresh). Example:
+.. code-block:: python
+   params_to_optimize = ['jump_rejection_threshold', 'expand_mask', 'p3thresh']
```
taylorbell57, line 283: If you're not doing GLBS, will it still iterate over 'expand_mask' and 'p3thresh'? This is a decent default for NIRSpec probably, but not a decent default for LRS.

docs/source/ecf.rst
```diff
@@ -257,6 +257,79 @@ In short this weights each pixel, :math:`i`, within a slope following :math:`w_i
...
+
+params_to_optimize_s3
+''''''''''''''''''''''''''
+List of parameters to optimize in Stage 3. Commenting out this line will use all single parameters listed below. Single parameter options: dqmask, bg_deg, bg_thresh, bg_hw, bg_method, p3thresh, spec_hw, median_thresh, window_len, p7thresh. Double parameter options: any combination of the above, joined by two underscores (e.g., bg_hw__p3thresh). Special cases: spec_hw__bg_hw (requires spec_hw < bg_hw). Example:
+.. code-block:: python
+   params_to_optimize_s3 = ['spec_hw__bg_hw', 'dqmask', 'bg_deg', 'bg_thresh', 'bg_method', 'p3thresh', 'median_thresh__window_len', 'p7thresh']
```
taylorbell57: Is this an example or the default? Wording is a bit confusing

docs/source/ecf.rst
```diff
@@ -257,6 +257,79 @@ In short this weights each pixel, :math:`i`, within a slope following :math:`w_i
...
+
+to optimize the background threshold between 3 and 5 sigma, in steps of 0.5
```
taylorbell57: This is missing its example code and probably belongs further down in this file

docs/source/ecf.rst
```diff
@@ -257,6 +257,79 @@ In short this weights each pixel, :math:`i`, within a slope following :math:`w_i
...
+
+params_to_optimize_s4
+'''''''''''''''''''''
+List of parameters to optimize in Stage 4. Single parameter options: mad_sigma, mad_box_width, sigma, box_width. Suggested double parameter options: mad_sigma__mad_box_width, sigma__box_width. Example:
+.. code-block:: python
+    params_to_optimize_s4 = ['mad_sigma__mad_box_width', 'sigma__box_width']
+
+sweep_<parameter_name>
+''''''''''''''''''''
+Range for parameter values to optimize. Only specify if overriding the default values in the meta Python files. For single parameters, use any of the following formats:
```
taylorbell57: Probably worth replacing "range" with "collection" or something like that here and elsewhere

docs/source/ecf.rst
```diff
@@ -257,6 +257,79 @@ In short this weights each pixel, :math:`i`, within a slope following :math:`w_i
...
+.. code-block:: python
+   sweep_<parameter_name> = range(min, max+1)
+   sweep_<parameter_name> = np.arange(min, max+step, step)
+   sweep_<parameter_name> = [3, 5, 6, 7]
```
taylorbell57, line 309: Worth adding a boolean example?

taylorbell57 (changes requested):
This current version isn't working properly for me for a MIRI/LRS dataset. I get a reasonable initial value:
```
Initial fitness value: 410.6930600610409
Initial white MAD: 330.1672891564689
Initial spec MAD: 8052.577090457203
```

And then it starts the first step in the S1opt optimization:
```
Optimizing parameter skip_firstframe over range: [False, True]
Initial parameter value: False
  Evaluating skip_firstframe = False...
```
and then eventually outputs (with some outputs redacted for posting publicly here)
```
Stage 4 MAD = 8052.58 ppm
Generating light curves
  Bandpass 0 = 5.000 - 12.000
    indices 4 - 250, 247 in total
  Sigma clipped 9 outliers in time series
    MAD = 330 ppm
Saving results...
Finished writing to ..._SpecData.h5
Finished writing to ..._LCData.h5

Total time (min): 0.06
Could not calculate fitness score for skip_firstframe = False.
Error: [Errno 39] Directory not empty: '.../S1opt_2026-05-08_..._run1/Stage3/S3_2026-05-08_..._run2/ap4_bg9/'
  Evaluating skip_firstframe = True...
```

And then it continues to finish Stage 4 successfully for each of the S1opt sweeps but says it "Could not calculate fitness score".

For reference, I have:
```
scaling_MAD_spec = 0.01
scaling_MAD_white = 1.0
```
and:
```
delete_intermediate = True
delete_final = True
```

Also, all of the `tests/test_lightcurve_fitting.py` tests seem to be failing right now

taylorbell57:
Please also merge in the latest main branch

docs/source/ecf.rst
kevin218: Agreed on being able to customize the fitness function.

docs/source/ecf.rst
kevin218: Changed to "recommended"

docs/source/ecf.rst
kevin218: I wanted to keep the optimization section all in one place, rather than having the parameters specific to Stages 3 and 4 further down in this file.

I'm not sure what you mean by "missing its example code"

taylorbell57: Line 289 doesn't seem to correspond to lines 283-288, as 289 seems more related to `sweep_<parameter_name>` and not `params_to_optimize_s3`

src/eureka/optimizer/S3opt_meta.py
```diff
@@ -0,0 +1,217 @@
...
+        '''
+        return
+
+    def set_MIRI_defaults(self):
+        '''
+        Set Optimizer specific defaults for MIRI.
+        '''
+        defaults = {
+            "sweep_bg_hw": range(5, 16),
+            "sweep_spec_hw": range(1, 10),
```
taylorbell57: spec_hw=1 gives me the following error:
```
  Locating source position...                                                                                                                                                                                                                                                
Could not calculate fitness score for spec_hw = 1 & bg_hw = 11.                                                                                                                                                                                                              
Error: The number of func parameters=4 must not exceed the number of data points=3
```
which I think likely comes from the centroiding step where it can't fit a Gaussian to only 3 points

taylorbell57:
I think I figured out that the `Could not calculate fitness score for ...` errors were because the intermediate files were being deleted *before* the scores were being computed which caused the error.

And for the inability to delete the intermediate files, ChatGPT suggested switching `rmtree` to:
```
def _safe_rmtree(path, *, retries=3, delay=0.25):
    """Remove a directory tree, tolerating transient ENOTEMPTY errors."""
    if path is None or not os.path.exists(path):
        return

    for attempt in range(retries + 1):
        try:
            shutil.rmtree(path)
            return
        except OSError as exc:
            if exc.errno == errno.ENOENT:
                return
            if exc.errno == errno.ENOTEMPTY and attempt < retries:
                gc.collect()
                time.sleep(delay)
                continue
            raise
```

taylorbell57:
I also don't get any intermediate fitness scores logged to the console when running S3opt unlike S1opt

taylorbell57:
Also, the optimizer's log file should log the fitness for every considered parameter value, while it is currently very sparse with just:
```
Optimizing parameters median_thresh__window_len over sweep: [range(3, 10), range(1, 22, 2)]
Initial parameter values: [5, 7]
  Evaluating median_thresh__window_len = 3 & 1...
  Evaluating median_thresh__window_len = 3 & 3...
  Evaluating median_thresh__window_len = 3 & 5...
  Evaluating median_thresh__window_len = 3 & 7...
  Evaluating median_thresh__window_len = 3 & 9...
  Evaluating median_thresh__window_len = 3 & 11...
  Evaluating median_thresh__window_len = 3 & 13...
  Evaluating median_thresh__window_len = 3 & 15...
  Evaluating median_thresh__window_len = 3 & 17...
  Evaluating median_thresh__window_len = 3 & 19...
  Evaluating median_thresh__window_len = 3 & 21...
...(this goes on for a long time)...
Best parameter value(s): ...
Best fitness value: ...
```

The main reason for this is that one may have (and I currently do have) cases where the "optimized" value is too aggressive and I want to be able to manually choose a different option based on the fitness values from the other options considered.

src/eureka/optimizer/S3opt_meta.py
```diff
@@ -0,0 +1,217 @@
...
+            "sweep_bg_deg": range(0, 2),  # 0 → 1
+            "sweep_p3thresh": range(3, 8),  # 3 → 7
+            "sweep_median_thresh": range(3, 10),  # 3 → 9
+            "sweep_window_len": range(1, 22, 2),  # 1 → 21, odd only
+            "sweep_p7thresh": range(5, 61, 5),  # 5 → 60, step 5
+            # Stage 4
+            "sweep_sigma": range(3, 8),  # 3 → 7
+            "sweep_box_width": range(11, 52, 10),  # 11 → 51, step 10
+            "sweep_mad_sigma": range(4, 8),  # 4 → 7
+            "sweep_mad_box_width": range(11, 52, 10),  # 11 → 51, step 10
```
taylorbell57, line 121: 51 is far too broad for LRS which has a pretty steep drop-off in the SNR at long wavelengths, and this results in the optimizer throwing out most of the long-wavelength data. I'd suggest capping this at < 32 or something in the MIRI-specific defaults

kevin218: I changed to `range(11, 32, 5)`

kevin218:
> I think I figured out that the `Could not calculate fitness score for ...` errors were because the intermediate files were being deleted _before_ the scores were being computed which caused the error.

The problem was a missing "/" at the end of the input directory when not running the stages sequentially and searching for save files in the children directories.

kevin218:
> I also don't get any intermediate fitness scores logged to the console when running S3opt unlike S1opt

Can you verify that `verbose = True` in the S3opt ECF? If that isn't the issue, you'll need to give more information so that I can diagnose the problem.

taylorbell57:
> > I also don't get any intermediate fitness scores logged to the console when running S3opt unlike S1opt
> 
> Can you verify that `verbose = True` in the S3opt ECF? If that isn't the issue, you'll need to give more information so that I can diagnose the problem.

Yup, my S3opt ECF had:
```
# Eureka! Control File for Optimizer Stage 3

# Optimizer Documentation: https://eurekadocs.readthedocs.io/en/latest/ecf.html#stages-1-3-4-optimizer
# Rename this file to match the event label given in run_eureka_optimization.py
# and run_eureka.py (e.g., S1opt_nirspec_fs_template.ecf).

# Fitness Scaling
# Use these settings to override the default fitness scaling factors.
scaling_MAD_spec    0.01            # Optimizer scaling for 2D MAD value
scaling_MAD_white   1.0             # Optimizer scaling for WLC MAD value

# List of single parameter options to optimize in Stage 3:
#   dqmask, bg_deg, bg_thresh, bg_hw, bg_method, p3thresh, spec_hw, median_thresh, window_len, p7thresh
# Double parameter options: any combination of the above, joined by two underscores (e.g., bg_hw__p3thresh)
# Special cases: spec_hw__bg_hw (requires spec_hw < bg_hw)
params_to_optimize_s3   ['spec_hw__bg_hw', 'dqmask', 'bg_deg', 'bg_thresh', 'bg_method', 'p3thresh', 'median_thresh__window_len', 'p7thresh']

# List of single parameter options to optimize in Stage 4:
#   mad_sigma, mad_box_width, sigma, box_width
# Suggested double parameter options:
#   mad_sigma__mad_box_width, sigma__box_width
params_to_optimize_s4   ['mad_sigma__mad_box_width', 'sigma__box_width']

# Range for parameter values to optimize
# Only specify if overriding the default values in S3opt_meta.py
# For single parameters, use format: sweep_<parameter_name> = range(min, max+1) or np.arange(min, max+step, step) if step != 1
# For double parameters, use format: sweep_<parameter1>__<parameter2> = [range(min1, max1+1), range(min2, max2+1)]
# sweep_bg_thresh    np.arange(3,5.5,0.5)  # Example: optimize BG threshold between 3 and 5 sigma, in steps of 0.5

# Diagnostics
isplots_S3opt       1               # Generate none (0), few (1), some (3), or many (5) figures (Options: 1 - 5)
delete_intermediate False           # If True, intermediate directories will be deleted after each optimization step
delete_final        False           # If True, intermediate directories will be deleted upon completion of optimization run
verbose             True            # If True, more details will be printed about steps in the optimization process
hide_plots          True            # If True, plots will automatically be closed rather than popping up

# Project directory
topdir              REDACTED

# Directories relative to topdir
inputdir            Optimized/Stage2
outputdir           Stage3opt
```

And I was using version `1.4.dev154+g34ab8ed14` (which is from [this commit](https://github.com/kevin218/Eureka/pull/757/commits/34ab8ed14d5435099881a9d36a901dbe0168a0d2)).

And the S3opt log file had the following (truncated for readability here):
```
Copying S3opt control file
Initial fitness value: 366.5190359763011
Initial white MAD: 292.3796907510212
Initial spec MAD: 7413.934522527989

Optimizing parameters spec_hw__bg_hw over sweep: [range(2, 10), range(5, 16)]
Initial parameter values: [4, 9]
  Evaluating spec_hw__bg_hw = 2 & 5...
  Evaluating spec_hw__bg_hw = 2 & 6...
  Evaluating spec_hw__bg_hw = 2 & 7...
  Evaluating spec_hw__bg_hw = 2 & 8...
  Evaluating spec_hw__bg_hw = 2 & 9...
  Evaluating spec_hw__bg_hw = 2 & 10...
  Evaluating spec_hw__bg_hw = 2 & 11...
  Evaluating spec_hw__bg_hw = 2 & 12...
  Evaluating spec_hw__bg_hw = 2 & 13...
  Evaluating spec_hw__bg_hw = 2 & 14...
  Evaluating spec_hw__bg_hw = 2 & 15...
  Evaluating spec_hw__bg_hw = 3 & 5...
  Evaluating spec_hw__bg_hw = 3 & 6...
  Evaluating spec_hw__bg_hw = 3 & 7...
  Evaluating spec_hw__bg_hw = 3 & 8...
  Evaluating spec_hw__bg_hw = 3 & 9...
  Evaluating spec_hw__bg_hw = 3 & 10...
  Evaluating spec_hw__bg_hw = 3 & 11...
  Evaluating spec_hw__bg_hw = 3 & 12...
  Evaluating spec_hw__bg_hw = 3 & 13...
  Evaluating spec_hw__bg_hw = 3 & 14...
  Evaluating spec_hw__bg_hw = 3 & 15...
  Evaluating spec_hw__bg_hw = 4 & 5...
  Evaluating spec_hw__bg_hw = 4 & 6...
  Evaluating spec_hw__bg_hw = 4 & 7...
  Evaluating spec_hw__bg_hw = 4 & 8...
  Evaluating spec_hw__bg_hw = 4 & 9...
  Evaluating spec_hw__bg_hw = 4 & 10...
  Evaluating spec_hw__bg_hw = 4 & 11...
  Evaluating spec_hw__bg_hw = 4 & 12...
  Evaluating spec_hw__bg_hw = 4 & 13...
  Evaluating spec_hw__bg_hw = 4 & 14...
  Evaluating spec_hw__bg_hw = 4 & 15...
  Evaluating spec_hw__bg_hw = 5 & 5...
  Evaluating spec_hw__bg_hw = 5 & 6...
  Evaluating spec_hw__bg_hw = 5 & 7...
  Evaluating spec_hw__bg_hw = 5 & 8...
  Evaluating spec_hw__bg_hw = 5 & 9...
  Evaluating spec_hw__bg_hw = 5 & 10...
  Evaluating spec_hw__bg_hw = 5 & 11...
  Evaluating spec_hw__bg_hw = 5 & 12...
  Evaluating spec_hw__bg_hw = 5 & 13...
  Evaluating spec_hw__bg_hw = 5 & 14...
  Evaluating spec_hw__bg_hw = 5 & 15...
  Evaluating spec_hw__bg_hw = 6 & 5...
  Evaluating spec_hw__bg_hw = 6 & 6...
  Evaluating spec_hw__bg_hw = 6 & 7...
  Evaluating spec_hw__bg_hw = 6 & 8...
  Evaluating spec_hw__bg_hw = 6 & 9...
  Evaluating spec_hw__bg_hw = 6 & 10...
  Evaluating spec_hw__bg_hw = 6 & 11...
  Evaluating spec_hw__bg_hw = 6 & 12...
  Evaluating spec_hw__bg_hw = 6 & 13...
  Evaluating spec_hw__bg_hw = 6 & 14...
  Evaluating spec_hw__bg_hw = 6 & 15...
  Evaluating spec_hw__bg_hw = 7 & 5...
  Evaluating spec_hw__bg_hw = 7 & 6...
  Evaluating spec_hw__bg_hw = 7 & 7...
  Evaluating spec_hw__bg_hw = 7 & 8...
  Evaluating spec_hw__bg_hw = 7 & 9...
  Evaluating spec_hw__bg_hw = 7 & 10...
  Evaluating spec_hw__bg_hw = 7 & 11...
  Evaluating spec_hw__bg_hw = 7 & 12...
  Evaluating spec_hw__bg_hw = 7 & 13...
  Evaluating spec_hw__bg_hw = 7 & 14...
  Evaluating spec_hw__bg_hw = 7 & 15...
  Evaluating spec_hw__bg_hw = 8 & 5...
  Evaluating spec_hw__bg_hw = 8 & 6...
  Evaluating spec_hw__bg_hw = 8 & 7...
  Evaluating spec_hw__bg_hw = 8 & 8...
  Evaluating spec_hw__bg_hw = 8 & 9...
  Evaluating spec_hw__bg_hw = 8 & 10...
  Evaluating spec_hw__bg_hw = 8 & 11...
  Evaluating spec_hw__bg_hw = 8 & 12...
  Evaluating spec_hw__bg_hw = 8 & 13...
  Evaluating spec_hw__bg_hw = 8 & 14...
  Evaluating spec_hw__bg_hw = 8 & 15...
  Evaluating spec_hw__bg_hw = 9 & 5...
  Evaluating spec_hw__bg_hw = 9 & 6...
  Evaluating spec_hw__bg_hw = 9 & 7...
  Evaluating spec_hw__bg_hw = 9 & 8...
  Evaluating spec_hw__bg_hw = 9 & 9...
  Evaluating spec_hw__bg_hw = 9 & 10...
  Evaluating spec_hw__bg_hw = 9 & 11...
  Evaluating spec_hw__bg_hw = 9 & 12...
  Evaluating spec_hw__bg_hw = 9 & 13...
  Evaluating spec_hw__bg_hw = 9 & 14...
  Evaluating spec_hw__bg_hw = 9 & 15...
Best parameter value(s): [3 5]
Best fitness value: 362.5290288816884

Optimizing parameter dqmask over sweep: [True, False]
Initial parameter value: True
  Evaluating dqmask = True...
  Evaluating dqmask = False...
Best parameter value(s): [True]
Best fitness value: 362.5290288816884
```
and then proceeds through many more parameter sweeps in a similar manner (none of those `...` are something I wrote but are what is actually logged).

taylorbell57:
@kevin218, is this ready for my review again? No rush - just don't want to be accidentally holding this back

kevin218:
> @kevin218, is this ready for my review again? No rush - just don't want to be accidentally holding this back

@taylorbell57 Yes! Sorry, I thought I had indicated that it was ready a couple weeks ago, but apparently not. Thanks for checking.

src/eureka/optimizer/S1opt_meta.py
taylorbell57: I'm going to add here and in the MIRI spectra section the added default of tuning a few skip parameters which are non-trivial to guess what they should be

docs/source/ecf.rst
```diff
@@ -257,6 +257,88 @@ In short this weights each pixel, :math:`i`, within a slope following :math:`w_i
...
+'''''''''''''''''''''''''''''''
+Sets how many plots should be saved during optimization. Can generate none (0), few (1), some (3), or many (5) figures (Options: 1 - 5).
+
+delete_intermediate
+'''''''''''''''''''
+If True, intermediate directories will be deleted after each optimization step.  This may be useful for Stage 1 optimization since the outputs can be quite large.
+
+delete_final
+''''''''''''
+If True, intermediate directories will be deleted upon completion of optimization run.
```
taylorbell57: Can you please tweak this a bit to make it clearer how this differs from `delete_intermediate`

src/eureka/lib/util.py
```diff
@@ -73,13 +75,43 @@ def readfiles(meta):
                              f'{meta.filename} to point to the folder '
                              f'containing the "{meta.suffix}.fits" files.')
 
-    meta.segment_list = np.array(sn.sort_nicely(meta.segment_list))
+    if meta.isopt_S1 or meta.isopt_S3:
```
taylorbell57: Should we perhaps add a separate flag to control this in case folks want to optimize against all of the files for some reason?

kevin218: Yeah, it should be trivial to add that option. 

It looks like the sole use of `isopt_S1` and `isopt_S3` is right here, so we could rename the variable to something more obvious and update the docs.

src/eureka/optimizer/S3opt_optimizer.py
kevin218: Looks like I forgot to address this one.

taylorbell57:
@kevin218, aside from the small remaining open comments, this PR looks good-to-go to me

kevin218 (dismissed):

kevin218 (dismissed):

taylorbell57 (approved):

Issue #760, opened 2025-04-21
[Bug]: Multwhite fitting pulls earliest save file from inputdirlist, not latest


jbrande:
### FAQ check

- [x] Yes, I checked the FAQ and my question has not been addressed.

### Instrument

Light curve fitting (Stages 4-6)

### What happened?

When fitting multiple white lightcurves, S5 will look in the directory specified in `meta.inputdirlist` for alternate save files, but it always loads the first filename in the list. This is in contrast to manageevent.findevent(), which always loads the latest file in the various subdirectories. This can lead to unintended behavior when working with multiple datasets, (e.g. testing extraction parameters for NRS1/NRS2 datasets) where the white lightcurves may not have corresponding Stage4 results.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #761, opened 2025-04-21
Editing s5_fit to load latest saved white LC, like in manageevent.findevent()

jbrande:
Handles #760.


taylorbell57:
Thanks for finding this scary bug and patching it!!

jbrande:
I ran across it last week and got very confused for a couple hours, because the warning message at the top of the Stage 5 run immediately gets quashed by a bunch of sampling run information! Easy fix, though.

Issue #591, opened 2023-12-13
[Enhancement]: Merge differentiable and standard models if possible


taylorbell57:
### Instrument

Light curve fitting (Stages 4-6)

### What is your suggestion?

At present, there are separate implementations of the numpy-based and theano-based astrophysical and detector models which requires twice as much work if someone wants to add a new model or wants to change how a current model works. It will be a bit tricky to write things in a way that is fast for numpy-based models but also works for theano-based models (for-loops aren't slow for the latter since it gets compiled, but for-loops can be quite slow for the former), but I think for most cases it should be quite doable. Once we've worked out how to do it well for one of them, it should be very easy to extrapolate to the other functions

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #766, opened 2025-05-15
Adding support for jax and jaxoplanet

### LINKED ISSUE #170: S5: Look into using jax based packages for both GP and nested sampling
To improve speed of light curve fitting, especially using Gaussian Processes (GPs), we have already discussed adding [tinygp package](https://tinygp.readthedocs.io/en/latest/index.html) to the stage 5 fitting. I haven't finished coding that yet though.
When looking into tinygp, I realised that it's based on [jax](https://jax.readthedocs.io/en/latest/notebooks/quickstart.html), which is kind of like numpy but for high-performance computing, and that there's also a fast nested sampling algorithm using jax called [jaxns](https://github.com/Joshuaalbert/jaxns), which is quite a bit faster than dynesty or multinest. Would be useful to include that in the future.

### LINKED ISSUE #310: speed up GPs (try JAAX, polychord, or implementing slow parts in C/C++)

taylorbell57:
Because starry+PyMC3 are now so outdated that we can no longer maintain support for them, I am quickly working to add jax+jaxoplanet support.

This is just a draft PR to help make sure others know this feature is getting added soon, and I'll mark it as ready for review once it is.

This PR will resolve #170 and resolve #310


taylorbell57:
Reminding myself here that I can/should make progress on #591 in this PR

PR #768, opened 2025-05-19
Adds number of pixel bins as option in S4

kevin218:
I've been running some tests using MIRI that required specifying spectroscopic channels using a fixed number of pixel rows.  The tests were successful, so I'm adding this new functionality to the `main` branch.

I've also enabled RxR plus CxC BG subtraction for MIRI.  Similar to the 390hz feature, this effort didn't improve the resulting spectrum.

Finally, I fixed a small (but important) bug in Stage 6.  After relabeling `a0` to `rp` in the Harmonica Stage 5 code, I forgot to update Stage 6.

### DIFF
--- a/src/eureka/S4_generate_lightcurves/s4_genLC.py
+++ b/src/eureka/S4_generate_lightcurves/s4_genLC.py
@@ -453,10 +453,10 @@ def genlc(eventlabel, ecf_path=None, s3_meta=None, input_meta=None):
                     log.writelog(f"  Bandpass {i} = "
                                  f"{lc.wave_low.values[i]:.3f} - "
                                  f"{lc.wave_hi.values[i]:.3f}")
-                    # Compute valid indeces within wavelength range
+                    # Compute valid indices within wavelength range
                     index = np.where((spec.wave_1d >= lc.wave_low.values[i]) *
                                      (spec.wave_1d < lc.wave_hi.values[i]))[0]
-                    log.writelog(f"    Indeces {index[0]} - {index[-1]}, "
+                    log.writelog(f"    indices {index[0]} - {index[-1]}, "
                                  f"{len(index)} in total")
                     # Make masked arrays for easy summing
                     optspec_ma = np.ma.masked_where(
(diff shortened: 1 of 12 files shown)


src/eureka/S4_generate_lightcurves/s4_genLC.py
```diff
@@ -200,7 +200,32 @@ def genlc(eventlabel, ecf_path=None, s3_meta=None, input_meta=None):
...
+            elif meta.nspecchan is None and meta.npixelbins is not None:
+                # User wants bins defined by the given number of pixels
+                istart = np.where(wave_1d >= meta.wave_min)[0][0]
+                iend = np.where(wave_1d >= meta.wave_max)[0][0]
+                # Shift bins by some number of pixels (only useful for MIRI)
+                istart += meta.npixelshift
+                iend += meta.npixelshift
+                edges = wave_1d[istart:iend+meta.npixelbins:meta.npixelbins]
+                dwav = np.ediff1d(
+                    wave_1d[istart:iend+meta.npixelbins])[::meta.npixelbins]/2
```
taylorbell57, line 219: Please double-check that this is correct. At first glance, this doesn't look like it would work and it seems to me like you would want to move the `[::meta.npixelbins]` inside of the parentheses; otherwise, the ediff1d would be the difference in wavelength from one pixel to the next, rather than from one bin to the next

src/eureka/S4_generate_lightcurves/s4_genLC.py
```diff
@@ -433,6 +456,8 @@ def genlc(eventlabel, ecf_path=None, s3_meta=None, input_meta=None):
                     # Compute valid indeces within wavelength range
                     index = np.where((spec.wave_1d >= lc.wave_low.values[i]) *
                                      (spec.wave_1d < lc.wave_hi.values[i]))[0]
+                    log.writelog(f"    Indeces {index[0]} - {index[-1]}, "
+                                 f"{len(index)} in total")
```
taylorbell57, line 460: Old and new typo: indeces -> indices

taylorbell57 (changes requested):
Please also document the new ECF settings in the documentation pages and ECF templates

taylorbell57 (approved):

Issue #769, opened 2025-05-27
[Bug]: Stage 3 'tuple' object does not support item assignment


meganmansfield:
### FAQ check

- [x] Yes, I checked the FAQ and my question has not been addressed.

### Instrument

NIRSpec (Stages 1-3)

### What happened?

Last week I upgraded to Eureka v1.2.1 from v1.1 and I can no longer get NIRSpec Stage 3 to run. The issue seems to be appearing in the "straighten curvature" section. I originally thought this was an issue related to an attempt to extract a calibrated stellar spectrum, but I'm now getting the error even when I run a fresh Stages 1-3 using the exact same .ecf file set-ups that were working for me last week with Eureka v1.1.
I installed Eureka v1.2.1 in a fresh environment, which also meant I upgraded from python 3.10.14 to 3.11.0, but have not upgraded my OS or anything else in the time before first experiencing the bug.

### Error traceback output

```
  Reading file 1...
  Masking NaNs/infs in data arrays...
    FLUX has 238858 NaNs/infs, which is 0.44% of all pixels.
    ERR has 238858 NaNs/infs, which is 0.44% of all pixels.
    V0 has 238858 NaNs/infs, which is 0.44% of all pixels.
  Locating source position...
    Source position on detector is row 12.
  Converting from data numbers per second (DN/s) to electrons...
  Automatically getting reference files...
  Computing clean median frame...
  Correcting curvature and bringing the trace to the center of the detector...
    Correcting the wavelength solution...
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
File ~/Documents/Projects/W76_NIRSpec_pc/NRS1/run_eureka.py:28
     18 ecf_path = '.'+os.sep
     20 if __name__ == '__main__':
     21     # To skip one or more stages that were already run,
     22     # just comment them out below
   (...)     25 
     26     # meta = s2.calibrateJWST(eventlabel, ecf_path=ecf_path)
---> 28     spec, meta = s3.reduce(eventlabel, ecf_path=ecf_path)
     30     # spec, lc, meta = s4.genlc(eventlabel, ecf_path=ecf_path)
     31     # # meta, spec, ds = s4cal.medianCalSpec(eventlabel, ecf_path=ecf_path)
     32 
     33     # meta = s5.fitlc(eventlabel, ecf_path=ecf_path)
     34 
     35     # meta = s6.plot_spectra(eventlabel, ecf_path=ecf_path)

[...]

PR #770, opened 2025-05-29
fix error in straightening curvature

meganmansfield:
Not sure if this is the most efficient way to fix this bug, but I've tested that it leads to the same Stage 3 outputs as before the bug cropped up. It would be great for someone with a functioning Python 3.13 to test it though to make sure it works as expected! Addresses issue #769

### DIFF
--- a/src/eureka/S3_data_reduction/straighten.py
+++ b/src/eureka/S3_data_reduction/straighten.py
@@ -109,7 +109,7 @@ def roll_columns(data, shifts):
         # in each column
 
         arr = np.swapaxes(data[i], 0, -1)
-        all_idcs = np.ogrid[[slice(0, n) for n in arr.shape]]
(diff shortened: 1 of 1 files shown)


meganmansfield:
Not sure if I need to open another PR for this, but I just made one more small change. In my Python 3.11 at least, when you try and do src_pos_type = max, it interprets the word "max" as the built-in function max() and throws an error. I fixed this by just changing it to "maximum".

taylorbell57:
The general recommendation is to use `'max'` and similarly explicit strings. In practice, I and many others get lazy and just do things like `gaussian` which works but then there's some risk of folks getting confused as to why `max` doesn't work. I think this is opening a larger discussion though and this would break backwards compatibility, so I'm going to undo your tweak to this PR, and you can open a separate issue if you want so that we can have a broader discussion about whether or not / how we want to protect against confusion with built-in functions.

PR #771, opened 2025-05-30
Enables NIRISS SUBSTRIP96 mode

kevin218:
PASTASOSS only computes the trace for SUBSTRIP256, so we currently need to supply an offset.  The expected value is -10 pixels, but this can vary by a pixel or two.

I've made sure that Order 2 gets masked, even when only requesting the Order 1 spectrum.  Additionally, I've limited the Order 2 mask so that it doesn't interfere with the spectral extraction of Order 1.

Finally, I (hopefully) fixed the NIRISS MAD calculation.

### DIFF
--- a/docs/source/ecf.rst
+++ b/docs/source/ecf.rst
@@ -440,11 +440,11 @@ Only used for NIRISS. List of spectral orders to be reduced.
 
 trace_offset
 ''''''''''''
-Only used for NIRISS. Manually adjust trace position for SUBSTRIP96 mode.  Default is -10 pixels.
+Only used for NIRISS. PASTASOSS v1.2 doesn't correctly compute the trace position for SUBSTRIP96 mode; therefore, we have to apply a manual offset in the cross-dispersion direction.  The default is -12 pixels for SUBSTRIP96 and should be good to within a pixel or two.  If you see in Fig. 3304 that the spectrum is not quite centered, you should adjust the ``trace_offset`` accordingly.
 
 src_ypos
 ''''''''
-Only used for NIRISS. Spectral trace will be shifted to given vertical position.  Must be same size as meta.orders.
+The user should only specify this parameter for NIRISS data. Spectral trace will be shifted to given vertical position.  Must be same size as meta.orders.
 
 src_pos_type
 ''''''''''''
--- a/docs/source/ecf.rst
+++ b/docs/source/ecf.rst
@@ -434,6 +434,18 @@ subarray_halfwidth
 ''''''''''''''''''
 Only used if any element of xwindow or ywindow is set to None, and only used for MIRI photometry data. This sets the half-width of the xwindow, ywindow subarray in pixels and is centered on the approximate centroid position. For MIRI photometry, the default is 75 pixels, which is a good value for most datasets since the MIRI full frame images are very large and it is generally helpful and faster to zoom-in on the science target.
 
+orders
+''''''
+Only used for NIRISS. List of spectral orders to be reduced.
+
+trace_offset
+''''''''''''
+Only used for NIRISS. Manually adjust trace position for SUBSTRIP96 mode.  Default is -10 pixels.
+
+src_ypos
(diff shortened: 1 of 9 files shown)


src/eureka/S3_data_reduction/s3_meta.py
```diff
@@ -274,7 +274,9 @@ def set_NIRISS_defaults(self):
         self.curvature = getattr(self, 'curvature', True)
         self.src_ypos = getattr(self, 'src_ypos', [35, 80])
         self.orders = getattr(self, 'orders', [1, 2])
+        self.all_orders = getattr(self, 'all_orders', [1, 2])
         self.record_ypos = getattr(self, 'record_ypos', False)
+        self.trace_offset = getattr(self, 'trace_offset', None)
```
taylorbell57, line 279: Please document all new ECF settings in the template ECFs and on the readthedocs pages

src/eureka/S3_data_reduction/niriss.py
```diff
@@ -133,17 +133,30 @@ def get_wave(data, meta, log):
...
+    for order in meta.all_orders:
         # Get trace for the given order and pupil position
         trace = get_soss_traces(pwcpos=pwcpos, order=str(order), interp=True)
+        if data.attrs['mhdr']['SUBARRAY'] == 'SUBSTRIP96' and \
+                meta.trace_offset is None:
+            # PASTASOSS doesn't account for different substrip starting rows;
+            # therefore, reset default trace offset to -10 pixels.
+            # SUBSTRIP96: SUBSTRT2 = 1803
+            # SUBSTRIP256: SUBSTRT2 = 1793
+            meta.trace_offset = -10
```
taylorbell57, line 153: Is it always precisely 10 pixels - your message seemed to suggest this wasn't always 10 px. Do we need something like the WFC3 2D-image correlation code?

src/eureka/S3_data_reduction/s3_meta.py
taylorbell57: It seems to me like we could add a subarray-specific default for `orders`, `all_orders`, and `trace_offset` here, rather than requiring user-input values or hard-coding defaults inside the `niriss.read` function (which makes tracking-down the default values tricky)

taylorbell57 (changes requested):

src/eureka/S3_data_reduction/niriss.py
kevin218: The value is constant for a given visit, but not between visits. So far, values of -11 or -12 have worked best for the visits I've looked at.

src/eureka/S3_data_reduction/s3_meta.py
kevin218: Documentation added.

In exploring your 2nd point, I'm not sure how we would know what substrip mode is being used when calling `s3_meta.S3MetaClass()` when it's the first step in `s3_reduce.py`.  The ideal solution is to have PASTASOSS compute the trace correctly, at which point, we can get rid of the code in `niriss.get_wave()`.

docs/source/ecf.rst
```diff
@@ -434,6 +434,18 @@ subarray_halfwidth
...
+''''''
+Only used for NIRISS. List of spectral orders to be reduced.
+
+trace_offset
+''''''''''''
+Only used for NIRISS. Manually adjust trace position for SUBSTRIP96 mode.  Default is -10 pixels.
+
+src_ypos
+''''''''
+Only used for NIRISS. Spectral trace will be shifted to given vertical position.  Must be same size as meta.orders.
```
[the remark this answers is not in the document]
kevin218: Good catch!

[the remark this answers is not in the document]
kevin218: Actually, wait... this was intentional.  For NIRISS, we don't want to call `source_pos_wrapper`.  Instead, we specify where we want the spectrum to be on the detector.

PR #782, opened 2025-06-30
Adds Fig 4106 (MAD Outliers)

kevin218:
I've created a new figure in Stage 4 that plots the MAD values for the binned LCs, unbinned LCs, and deviation of the unbinned LCs from the band-integrated LC.  The purpose of this plot is to visually identify outlier pixel columns that can then be masked in subsequent runs of Stage 4 (using `mask_columns` in the ECF).  What constitutes an outlier is still quite subjective, so I didn't write any code that automatically identified and masked outliers.

I added an example of Fig 4106 (and 4105, which was missing) to the documentation.  Note that Fig 4104 is still missing from the documentation, but I can't generate it with my NIRISS dataset.

I also tracked down a bug that had the star flux and BG flux being appended to the same list (`meta.mad_s4_binned`).


docs/media/S4_out/fig4106_MAD_Outliers.png
taylorbell57, line 1: Some instruments have MAD values that pretty steeply change with wavelength. Instead of (or in addition to) a figure like this, I often make a version after boxcar smoothing to get rid of broad trends and help identify individual outliers.

What do you think?

taylorbell57 (changes requested):

docs/media/S4_out/fig4106_MAD_Outliers.png
kevin218: Here is where I'm at with the new figure:
Figure_4106
Thoughts?  Suggestions?

How do you feel about automatically masking the identified outlier columns?  The alternative is to print out the list and let the user decide in a subsequent iteration of Stage 4.

taylorbell57: Awesome - I love that. I'm inclined to have a sigma threshold in the ECF and then automatically clip based on that, but to plot and log which columns were clipped so users can investigate and tweak their thresholds. Looks like what you have largely does that already

src/eureka/S4_generate_lightcurves/s4_meta.py
```diff
@@ -134,6 +134,8 @@ def set_defaults(self):
         self.maxiters = getattr(self, 'maxiters', 20)
         self.boundary = getattr(self, 'boundary', 'fill')
         self.fill_value = getattr(self, 'fill_value', 'mask')
+        # Used in Fig 4106
+        self.mad_sigma = getattr(self, 'mad_sigma', 7)
```
taylorbell57, line 138: Please add this parameter to docs/source/ecf.rst

src/eureka/S4_generate_lightcurves/s4_genLC.py
```diff
@@ -326,9 +326,19 @@ def genlc(eventlabel, ecf_path=None, s3_meta=None, input_meta=None):
...
+            # Plot MAD values to help identify outliers
+            if meta.isplots_S4 >= 1 and not meta.photometry:
+                outliers = plots_s4.mad_outliers(meta, lc, spec)
+                if np.any(outliers):
+                    log.writelog(f'Identified {np.size(outliers)} outlier'
+                                 f' columns in Fig 4106.',
+                                 mute=(not meta.verbose))
+                    meta.mask_columns = np.union1d(meta.mask_columns,
+                                                   outliers).astype(int)
+
```
taylorbell57, line 338: I don't think this code block should depend on the isplots_S4 setting

src/eureka/S4_generate_lightcurves/plots_s4.py
```diff
@@ -331,6 +318,99 @@ def lc_driftcorr(meta, wave_1d, optspec_in, optmask=None, scandir=None):
...
+    # Identify only high outliers from residuals
+    masked_mad = sigma_clip(residual_mad, sigma_upper=meta.mad_sigma,
+                            sigma_lower=100, maxiters=meta.maxiters,
+                            masked=True, copy=True)
+    masked_dev = sigma_clip(residual_dev, sigma_upper=meta.mad_sigma,
+                            sigma_lower=100, maxiters=meta.maxiters,
+                            masked=True, copy=True)
+    x_mad_outliers = x_mask[np.where(masked_mad.mask)[0]]
+    x_dev_outliers = x_mask[np.where(masked_dev.mask)[0]]
+    outliers = np.union1d(x_mad_outliers, x_dev_outliers)
```
taylorbell57, line 377: These calculations should be moved outside of the plotting function as the MAD clipping shouldn't depend on whether or not the plot was made, and this code is not purely related to plotting

taylorbell57 (changes requested):

src/eureka/S4_generate_lightcurves/outliers.py
```diff
@@ -0,0 +1,84 @@
...
+    # Identify only high outliers from residuals
+    masked_mad = sigma_clip(residual_mad, sigma_upper=meta.mad_sigma,
+                            sigma_lower=100, maxiters=meta.maxiters,
+                            masked=True, copy=True)
+    masked_dev = sigma_clip(residual_dev, sigma_upper=meta.mad_sigma,
+                            sigma_lower=100, maxiters=meta.maxiters,
+                            masked=True, copy=True)
+    x_mad_outliers = x_mask[np.where(masked_mad.mask)[0]]
+    x_dev_outliers = x_mask[np.where(masked_dev.mask)[0]]
+    outliers = np.union1d(x_mad_outliers, x_dev_outliers)
```
taylorbell57, line 67: We probably want to maintain support for `meta.mad_sigma = None` which doesn't mask any outliers

taylorbell57 (approved):

PR #785, opened 2025-07-07
Implements common-mode systematics model

kevin218:
This relatively small PR implements a new model that utilizes results from a white LC fit to remove common-mode instrument systematics.


src/eureka/S5_lightcurve_fitting/s5_meta.py
```diff
@@ -97,6 +97,8 @@ def set_defaults(self):
         self.num_planets = getattr(self, 'num_planets', 1)
         self.compute_ltt = getattr(self, 'compute_ltt', None)
         self.force_positivity = getattr(self, 'force_positivity', False)
+        # Path to ECSV file containing common mode variations
+        self.common_mode_file = getattr(self, 'common_mode_file', None)
```
taylorbell57, line 101: I recommend adding the following:
```
if self.common_mode_file is not None:
    # Require the common mode name to be specified if a common_mode_file is provided
    self.common_mode_name = getattr(self, 'common_mode_name')\
```

taylorbell57 (changes requested):

PR #788, opened 2025-07-11
Report upper limits on nightside flux when using force_positivity

kevin218:
When setting `force_positivity = True`, the currently reported nightside flux is skewed when pushing up against the zero flux boundary.  For these channels, we should actually be reporting upper limits on `fn`.

This PR appends two additional columns to the Stage 6 Astropy table: the 3-sigma upper limit on `fn` and a boolean where True indicates the `fn` value should be reported as an upper limit.

@taylorbell57 If you think it would be useful, I can also look into implementing something similar for the dayside flux, `fp`, and the phase curve amplitude, `pc_amp`.  Are there any other parameters where this might be useful?

To Do List:
- [x] Documentation
- [x] Generalize to also report upper limits on `fp` etc?


taylorbell57:
Great thinking!

I think generalizing this for `fp` would be quite helpful for rocky exoplanet eclipse measurements. It's less obvious to me that it'd be useful for `pc_amp`, but it'd likely be worth adding at the same time. I can't think of any other parameters that this would be helpful for off the top of my head

kevin218:
@taylorbell57 I ended up only implementing this for `fn` and `fp`.  There wasn't a simple way of determining whether or not `pc_amp` should be reported as an upper limit since it can sometimes exceed `fp` when there is a strong PC offset.

PR #800, opened 2025-08-20
Adding support for wavelength-shared parameters

taylorbell57:
I made a new `s5_meta.wl_groups` parameter which will allow for some fitted parameters to be shared between only "channels" with the same wavelength. The values of this array should be integer indices. To clarify, one could set `s5_meta.wl_groups = [0, 0, 0, 1, 1]` for a dataset comprised of 3 F1500W observations followed by 2 F1280W observations (s5_meta.wl_groups need not be sorted, but it does need to match the order of `inputdir+inputdirlist`).

An example use case of this is fitting all MIRI imaging observations of TRAPPIST-1 planets, which include F1280W and F1500W observations. For those data, one may want the eclipse depth to be shared between all observations in the same filter, while the orbital parameters would be shared between all observations. With the current code, one could not simply set `fp` to be shared (which forces it to be shared across all observations), but with the code in this PR you can have an `fp` (implicit `_wl0`) which could be used for F1500W data and an `fp_wl1` which could be used for F1280W data while you could have the orbital parameters (like `per`) shared between all observations.

This new code works similarly to other parameter auto-completions, where if you don't explicitly set a `_wl#` parameter then it is assumed that you want to re-use the wl=0 parameter. For example, if you just set one `per` in your EPF, then there will not be a separate `per_wl1` made. But if you make a `fp` and a `fp_wl1` then the two will be used for wl=0 and wl=1 channels accordingly.

This is still a work in progress, and more work and testing is needed before this PR is ready for review


environment.yml
```diff
@@ -1,9 +1,8 @@
-name: eureka
+name: eureka_temp
```
kevin218: Why the name change?

environment.yml
```diff
@@ -1,9 +1,8 @@
...
 channels:
     - conda-forge
-    - defaults
 dependencies:
-    - python==3.13.0
+    - python==3.12
```
kevin218: Why are we rolling back from 3.13 to 3.12?  Also, the setup.cfg uses python >=3.11,<3.14.

src/eureka/S6_planet_spectra/s6_spectra.py
```diff
@@ -722,7 +726,7 @@ def load_s5_saves(meta, log, fit_methods, n_samples=1):
         A list of sample arrays, one for each channel.
     """
     if 'dynesty' in fit_methods:
-        if meta.run_dynamic:
+        if hasattr(meta, 'run_dynamic') and meta.run_dynamic:
```
kevin218: `run_dynamic` should be defined in s6_meta.py so that we don't have to check that is exists using `hasattr`.

src/eureka/S5_lightcurve_fitting/models/CommonModeModel.py
```diff
@@ -1,54 +1,50 @@
...
+            and ``common_mode_name`` attributes.
+        log : logedit.Logedit
+            The current log in which to output messages from this function.
         **kwargs : dict
             Additional parameters to pass to
             eureka.S5_lightcurve_fitting.models.Model.__init__().
-            Can pass in the parameters, longparamlist, nchan, and
-            paramtitles arguments here.
+            Expects 'meta' (with .common_mode_file and .common_mode_name)
+            and 'log'.
```
kevin218: Seems unnecessary given Lines 15-17.

src/eureka/S5_lightcurve_fitting/models/CentroidModel.py
```diff
@@ -7,32 +7,33 @@
...
 
-    This can be used to do a linear decorrelation against the x position
-    (axis='xpos'), y position (axis='ypos'), x width (axis='xwidth'),
-    or y width (axis='ywidth').
-
+    Linear decorrelation against x position (axis='xpos'), y position
+    (axis='ypos'), x width (axis='xwidth'), or y width (axis='ywidth').
     """
-    def __init__(self, **kwargs):
+    def __init__(self, axis=None, **kwargs):
```
kevin218: Make this non-optional?

src/eureka/S5_lightcurve_fitting/models/GPModel.py
```diff
@@ -168,26 +128,34 @@ def eval(self, fit_lc, channel=None, gp=None, **kwargs):
...
+            res_good = residuals[good]
 
-            # Create the GP object with current parameters
+            # Build or reuse the GP object
             if input_gp is None:
                 gp = self.setup_GP(chan)
             else:
                 gp = input_gp
+                # If caller passed a pre-built GP, we may still need
+                # to the compute inputs.
```
kevin218: `to compute the inputs`

src/eureka/S5_lightcurve_fitting/models/GPModel.py
```diff
@@ -348,39 +316,35 @@ def get_kernel(self, kernel_name, k, c=0):
...
+        """Compute the GP log-likelihood.
 
         Parameters
         ----------
         fit_lc : ndarray
-            The fitted model.
-        channel : int; optional
-            If not None, only consider one of the channels. Defaults to None.
+            The fitted (non-GP) model.
+        channel : int, optional
```
kevin218: `int; optional`

src/eureka/S5_lightcurve_fitting/s5_fit.py
```diff
@@ -88,12 +89,16 @@ def fitlc(eventlabel, ecf_path=None, s4_meta=None, input_meta=None):
...
             # Load in the S4 metadata used for this particular aperture pair
-            if meta.data_format == 'eureka':
+            if (meta.data_format == 'eureka' and
+                    (meta.spec_hw != spec_hw_val or meta.bg_hw != bg_hw_val)):
+                meta.spec_hw = spec_hw_val
+                meta.bg_hw = bg_hw_val
                 meta = load_specific_s4_meta_info(meta)
+            elif meta.data_format != 'eureka':
+                meta.spec_hw = spec_hw_val
+                meta.bg_hw = bg_hw_val
```
kevin218: I think I commented on a similar piece of code above about always overwriting `meta.spec_hw` and `meta.bg_hw` and thus taking it out of the for loop.

kevin218 (changes requested):

environment.yml
taylorbell57: Ah oops, that was accidental

environment.yml
taylorbell57: Python 3.13 doesn't allow numpy==1.26.4 which I think one of our package versions need or needed (see #809).

src/eureka/S5_lightcurve_fitting/models/CommonModeModel.py
```diff
@@ -1,54 +1,50 @@
...
         """
-        # Inherit from Model class
+        if meta is None or log is None:
+            raise ValueError("CommonModeModel requires 'meta' and 'log' "
+                             "arguments.")
```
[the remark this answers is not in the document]
taylorbell57: My concern had revolved around argument ordering in the current places we call all these functions, and I'm not 100% sure we always use keyword arguments. But thinking about it again, we must be so I agree

src/eureka/S5_lightcurve_fitting/s5_fit.py
taylorbell57: I don't see any such comment - can you clarify what you mean?

kevin218: Ok, never mind.  Please disregard this comment.

kevin218 (changes requested):

kevin218 (approved):

taylorbell57:
@kevin218, sorry - this needs approval again. The automated tests don't test Stage 1 and I had missed that the number of arguments returned from the ramp_fit function had been reduced by one. That's been patched, as well as one source of the pkg_resources deprecation warning (I've also got PRs on a bunch of other packages to resolve that warning with their packages so we stop getting pestered)

kevin218 (approved):

PR #821, opened 2025-11-21
Enabling Celerite2 SHO GP kernels.

jbrande:
### Summary
This PR enables users to select SHO kernels for their GPs when using Celerite2. The SHO kernel takes in the same log-amplitude `A` and log-timescale `m`, as well as the quality factor of the oscillator `Q`. 

### Why
The existing Matern 3/2 kernel is useful and extensible, but does not necessarily give any physical intuition about the system being observed. A stochastically driven damped harmonic oscillator can reasonably model physical processes commonly seen in transit lightcurves, with the appropriate choice of quality factor. E.g. `Q = 1/sqrt(2)` can describe stellar granulation in lightcurves.

### Key Changes
`GPModel.py` now allows for `SHO` kernel choice for `celerite` GPs. and automatically parses the EPF for all three parameters `A, m, Q`. 

`docs/source/ecf.rst` also now describes the parameters and explains the motivation for the new GP kernel.

`demos/JWST/S5_template.ecf` and `demos/JWST/S5_fit_par_template.epf` also give useful examples for the new GP kernel.

### Testing
I've run some tests using the SHO kernel on some very small datasets, but haven't yet exhaustively tried different `Q` values, or tried to get equivalent results by setting `Q=0.5` (which should collapse to the Matern 3/2 kernel). I also haven't tested multi-channel datasets or shared GP parameters yet.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -3,7 +3,7 @@ channels:
     - conda-forge
     - defaults
 dependencies:
-    - python==3.13.0
+    - python==3.12.0
     - astropy
     - astroquery
     - bokeh[version='<3.0']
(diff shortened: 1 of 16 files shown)


taylorbell57 (changes requested):
Didn't we decide that the `*2` on the line:
```
metric = np.exp(self.coeffs[c, k, 1]*2)
```
in the george code was a typo?

Or does the metric used by george use a different unit than the other codes?

jbrande:
> Didn't we decide that the `*2` on the line:
> 
> ```
> metric = np.exp(self.coeffs[c, k, 1]*2)
> ```
> 
> in the george code was a typo?
> 
> Or does the metric used by george use a different unit than the other codes?

Yeah, the george API specifies that the metric is squared, where celerite2 and tinygp don't. The typo was in the original placeholder code for the tinygp kernels, where it was copied as-is from the george implementation and as such also squared the metric, when it shouldn't have. That's been fixed in this PR.

PR #837, opened 2026-04-17
Dependency policy

taylorbell57:
This PR adds Eureka's first explicit dependency-policy + compatibility-testing layer on top of the recent packaging/env modernization work.

## Summary

This PR adds Eureka's first explicit dependency-policy + compatibility-testing layer.

The point of this PR is to make our dependency/support story more explicit and more honest. In particular, it distinguishes between:

- declared dependency metadata in `pyproject.toml`
- a tested oldest-practical constraints recipe
- the generated conda environment
- the small permanent CI lanes that actually define tested behavior

This is also broadly aligned with the direction of [Scientific Python SPEC 0](https://scientific-python.org/specs/spec-0000): prefer realistic dependency windows, test them explicitly, and avoid implying support for broad version ranges that are not actually exercised. This PR is using that as guidance, not as a hard rule, since some of the astronomy/JWST/HST dependencies still need exceptions.

## Why this PR matters

The packaging modernization work made our dependency metadata much clearer, but it also made the next gap more obvious: we still needed a better answer to "what do we actually support?"

That matters for Eureka because we have:
- a mixed pip/conda install setup,
- optional extras with different roles,
- astronomy-specific dependencies that do not always move in lockstep with the broader Scientific Python ecosystem,
- occasional installation headaches caused by breaking changes from our explicit dependencies and their implicit dependencies, and
- a desire not to over-claim compatibility that we do not actually test.

So the goal here is to add a small compatibility layer on top of the packaging work so that the support story is tied more directly to CI-backed evidence.

### 1. A tested oldest-practical constraints recipe

Adds:

- `ci/constraints/oldest-practical-py311.txt`

This file is the current tested oldest-practical Python 3.11.0 recipe used by the new compatibility workflow.

### 2. A permanent compatibility workflow

Adds:

- `.github/workflows/dependency-compat.yml`

This intentionally small workflow includes:

- an `oldest-practical py311` job on Ubuntu / Python `3.11.0`
- install via:
  - `pip install -c ci/constraints/oldest-practical-py311.txt -e ".[jwst,hst,test]"`
- `pytest tests`
- a lightweight docs build sanity job using:
  - `pip install -e ".[jwst,docs]"`
  - `pandoc` as a system dependency for `nbsphinx` notebook rendering
  - `sphinx-build -W -b html docs/source docs/_build/html`

### 3. A dependency policy page

Adds:

- `docs/source/dependency_policy.md`

This page explains the current dependency/support model, including:

- `pyproject.toml` as the canonical dependency metadata source
- `environment.yml` as a generated UniDep artifact
- Eureka being SPEC 0-informed, but not strict in its application of that policy
- "supported" meaning "tested in CI"
- the distinction between metadata ranges, the tested oldest-practical recipe, and the generated conda environment

### 4. Small doc integration updates

Updates:

- `docs/source/index.rst`
- `docs/source/installation.rst`
- `docs/source/contribute.md`

These changes link the new dependency policy page from the places where users and contributors are most likely to need it.

### 5. A maintainer helper script

Adds:

- `ci/constraints/check-oldest-practical-py311.sh`

This provides a lightweight way to re-run the current oldest-practical checks locally.

## Key design decisions

- `pyproject.toml` remains the canonical dependency metadata source.
- `environment.yml` remains generated from `pyproject.toml` via UniDep.
- Eureka is informed by Scientific Python SPEC 0, but not strictly bound to it.
- "Supported" is being defined more honestly as "tested in CI."
- The tested oldest-practical recipe is not the same thing as claiming that every resolver-valid in-range combination is supported.
- The permanent compatibility lane is intentionally narrow and maintainable:
  - Python `3.11.0`
  - `-c ci/constraints/oldest-practical-py311.txt`
  - `-e ".[jwst,hst,test]"`
  - `pytest tests`
- The docs CI lane is also intentionally small and only verifies that the docs build cleanly with warnings treated as errors.

## NumPy support framing

This branch also updates the current NumPy floor to:

- `numpy>=1.25,<2.0`

The intended framing is:

- `numpy>=1.25` is the honest current tested `py311` floor
- `numpy>=1.20.0` is not a realistic Python 3.11-era floor
- we are still keeping `<2.0` for now because the current practical tested outcome of the dependency stack still resolves below NumPy 2
- in other words: `<2.0` is still the right constraint to keep today, but it is something we should revisit soonish

## CI fixes included here

The new workflow exposed a few small baseline issues that are fixed in this PR:

- `mc3==3.1.4` failed to import on the Ubuntu Python 3.11.0 runner, so the oldest-practical recipe now uses `mc3==3.2.1`.
- The docs job installs `.[jwst,docs]` because `docs/source/conf.py` imports `eureka`, whose current import path needs the JWST extra.
- The docs job installs `pandoc` because `nbsphinx` needs the external executable for notebook pages.
- A couple of pre-existing docs markup warnings are cleaned up so `sphinx-build -W` can run as a permanent check.

## What this PR intentionally does not do

This PR is intentionally pragmatic and low-risk. It does not try to:

- solve every packaging or dependency problem
- add a giant compatibility matrix
- add per-extra permanent compatibility jobs
- add Dependabot yet
- prove that all lower bounds in `pyproject.toml` are now fully established
- force pip and conda to resolve identically
- redesign UniDep or environment generation

## Follow-up work

Likely follow-up work after this PR:

- decide whether any additional `pyproject.toml` lower bounds should be tightened selectively
- revisit the oldest-practical recipe as upstream dependency constraints evolve, aided by `.github/workflows/dependency-compat.yml`
- consider a follow-up Dependabot PR, probably in a staged way: start with GitHub Actions updates, then evaluate conservative Python dependency update monitoring now that Eureka has a clearer tested-floor and compatibility-checking story to run those updates against
- continue refining the dependency policy language as the support model matures


taylorbell57:
@kevin218, just a heads up that this PR is open for review now, but note that the Required checks seems to be bugged in a new way, now requiring "PR CI" rather than any of the tests nested within that. If you prefer, I can also see if it's possible to have the PR CI workflow produce a final status that passes only if all sub-tests pass

docs/source/dependency_policy.md
```diff
@@ -0,0 +1,124 @@
...
+recipes, and the generated conda environment. This page describes how those
+pieces fit together and what the project currently means by "supported".
+
+## Canonical dependency metadata
+
+`pyproject.toml` is the canonical source of Eureka!'s dependency metadata.
+Direct dependency declarations, optional dependency groups, and packaging
+configuration should be maintained there.
+
+Eureka currently defines optional dependency groups including:
```
kevin218: Same typo in 5+ locations in this document: Eureka -> Eureka!, Eureka's -> Eureka!'s

kevin218 (dismissed):
I really like this new dependency policy!

Also, I was able to get rid of the "PR CI" hold up.

kevin218 (approved):

PR #841, opened 2026-04-25
Added the channel-specific calibrated light curve data to the S5 .txt output file

asafkaya:
With these changes, the S5 .txt output file now contains, in addition to the previous products, the _calibrated_ flux for spectroscopic light curves for outside Eureka! access. This allows the user to easily plot astrophysical model spectroscopic light curves over the calibrated data in scientific reports.

By the calibrated flux, I am referring to the `lc.flux/model.syseval() - GP` object that is created for example inside the `plot_fit()` function in `S5_lightcurve_fitting/plots_s5.py`. 

Please note that I have tested this new branch with limited trial runs, and make sure to check for broader compatibility.

### DIFF
--- a/src/eureka/S5_lightcurve_fitting/fitters.py
+++ b/src/eureka/S5_lightcurve_fitting/fitters.py
@@ -1363,12 +1363,19 @@ def save_fit(meta, lc, model, fitter, results_table, freenames, samples=[]):
             individual_models.append([comp.name, comp.eval(fit)])
     individual_models = np.array(individual_models, dtype=object)
 
+    model_noGP = model.eval(incl_GP=False)
+    gp = model.GPeval(model_noGP)
+
+    model_sys = model.syseval()
+
+    lcdata_cal = lc.flux/model_sys - gp
+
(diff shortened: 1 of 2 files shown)


taylorbell57:
I'll try to take a look at this in the next week or two, but noting here that we'll want to call the "calibrated" light curve something different since "calibrated" is already being used to mean absolutely calibrated flux (where we work out how bright the system is in units like mJy). I think something like "cleaned" or something would be less confusing. You're welcome to make such a change yourself, but I'll also be able to do so when I get to reviewing this PR

asafkaya:
> I'll try to take a look at this in the next week or two, but noting here that we'll want to call the "calibrated" light curve something different since "calibrated" is already being used to mean absolutely calibrated flux (where we work out how bright the system is in units like mJy). I think something like "cleaned" or something would be less confusing. You're welcome to make such a change yourself, but I'll also be able to do so when I get to reviewing this PR

Thanks! I had picked "calibrated" because that was the label used for the middle panel of 5101 figures. Whatever you think is best works though.

PR #844, opened 2026-05-08
S4: Rescaling the mu-grid in the Phoenix grid to more accurately align with the true edge of the star

erinmmay:
The Phoenix stellar grid defines mu=0 to be outside of the photosphere of the star due to the geometry of the model. This causes issues when using the model as-is to calculate limb darkening parameters. See Section 2.2 and Figure 4 here for a detailed discussion: https://academic.oup.com/mnras/article/450/2/1879/985166

This PR adapts code from David Sing to rescale the mu-grid prior to fitting limb darkening parameters to the stellar intensity profile. @astroRez assisted in coding this up. The PR includes a new figure to show the rescaled and original stellar intensity profile as integrated over the entire bandpass considered in S4.

fig4305_RescaledPhoenix

### DIFF
--- a/demos/JWST/S4_template.ecf
+++ b/demos/JWST/S4_template.ecf
@@ -38,6 +38,7 @@ logg            4.0     # Surface gravity in log g
 minmu			0.2		# minimum mu to use when calculating LD in ExoTiC-LD
 exotic_ld_direc /home/User/exotic-ld_data/ # Directory for ancillary files for exotic-ld, download from: https://zenodo.org/doi/10.5281/zenodo.6047317
 exotic_ld_grid  stagger # You can choose from kurucz (or 1D), stagger (or 3D), mps1, or mps2 model grids, or custom (to use custom_si_grid below)
+# rescale_phoenix  True. #Option to rescale the mu grid for phoenix models. See Section 2.2: https://academic.oup.com/mnras/article/450/2/1879/985166
 # custom_si_grid  /home/User/path/to/custom/stellar/intensity/profile  #Custom Stellar Intensity profile. For examples see Eureka/demos/JWST/Custom_Stellar_Intensity_Files
 # exotic_ld_file  /home/User/exotic-ld_data/Custom_throughput_file.csv # Custom throughput file, for examples see Eureka/demos/JWST/Custom_throughput_files
 # Options for SPAM
--- a/demos/JWST/S4_template.ecf
+++ b/demos/JWST/S4_template.ecf
@@ -38,7 +38,7 @@ logg            4.0     # Surface gravity in log g
 minmu			0.2		# minimum mu to use when calculating LD in ExoTiC-LD
 exotic_ld_direc /home/User/exotic-ld_data/ # Directory for ancillary files for exotic-ld, download from: https://zenodo.org/doi/10.5281/zenodo.6047317
 exotic_ld_grid  stagger # You can choose from kurucz (or 1D), stagger (or 3D), mps1, or mps2 model grids, or custom (to use custom_si_grid below)
-# rescale_phoenix  True. #Option to rescale the mu grid for phoenix models. See Section 2.2: https://academic.oup.com/mnras/article/450/2/1879/985166
+# rescale_phoenix  True # Option to rescale the mu grid for phoenix models. See Section 2.2: https://academic.oup.com/mnras/article/450/2/1879/985166
 # custom_si_grid  /home/User/path/to/custom/stellar/intensity/profile  #Custom Stellar Intensity profile. For examples see Eureka/demos/JWST/Custom_Stellar_Intensity_Files
 # exotic_ld_file  /home/User/exotic-ld_data/Custom_throughput_file.csv # Custom throughput file, for examples see Eureka/demos/JWST/Custom_throughput_files
 # Options for SPAM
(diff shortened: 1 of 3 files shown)


taylorbell57:
This new feature should be put behind a flag that can be controlled in the s4_meta. The default can certainly be to apply this correction if you feel that is warranted, but it should not be forced in case someone needs to reproduce an old result or a result produced by someone else. In a perfect world, we should also submit this as a bug or PR to ExoTiC-LD's repo so that it gets fixed/addressed at the source of the issue

erinmmay:
Yes, sounds good on the flag. After I submitted this I had the same thought.

Re: a PR/bug for ExoTIC-LD: We could, but apparently there are already talks with the Phoenix people to just release new versions of the model that address this (eventually). Either way, in the meantime, we have a working solution in Eureka! while a different option comes up. And it'll be easy to remove from Eureka! if ExoTIC-LD or Phoenix addresses it themselves.

src/eureka/S4_generate_lightcurves/generate_LD.py
```diff
@@ -98,10 +99,44 @@ def exotic_ld(meta, spec, log, white=False):
...
                                    custom_mus=s_mus,
                                    custom_stellar_model=custom_si)
+
     else:
         sld = StellarLimbDarkening(meta.metallicity, meta.teff, meta.logg,
                                    meta.exotic_ld_grid, meta.exotic_ld_direc)
+     
+    # Now if Phoenix, rescale the mu grid                         
+    if meta.exotic_ld_grid == 'phoenix' and meta.rescale_phoenix:
+        print('Rescaling Phoenix')
```
kevin218, line 109: Instead of a print statement, use `log.writelog('Rescaling Phoenix', mute=(not meta.verbose))` so that it also appears in the log file.

src/eureka/S4_generate_lightcurves/generate_LD.py
```diff
@@ -241,3 +276,62 @@ def spam_ld(meta, white=False):
...
+    """Rescale the Phoenix spherical-model mu grid before fitting.
+
+    Phoenix intensities include very sharp, low-mu limb points.  This follows
+    the rescaling approach commonly used for these spherical grids:
+
+        mu_scaled = (mu - mu_cri) / (1 - mu_cri)
+
+    where mu_cri is selected just inside the largest intensity-gradient point.
+    The resulting profile is interpolated onto a fixed 0..1 mu grid before the
+    limb-darkening law is fit.
```
kevin218, line 291: This is an excellent description, but the doc still needs `Parameters` and `Returns`.

src/eureka/S4_generate_lightcurves/plots_s4.py
```diff
@@ -494,3 +494,44 @@ def plot_extrapolated_throughput(meta, throughput_wavelengths, throughput,
     plt.savefig(meta.outputdir+fname, bbox_inches='tight', dpi=300)
     if not meta.hide_plots:
         plt.pause(0.2)
+        
+@plots.apply_style
+def plot_rescaled_phoenix(meta,sld, sld_new):
+    '''Make a plot of rescaled phoenix mus/intensities (Fig 4305).
```
kevin218, line 500: I think 4304 is the next available number.

src/eureka/S4_generate_lightcurves/plots_s4.py
```diff
@@ -494,3 +494,44 @@ def plot_extrapolated_throughput(meta, throughput_wavelengths, throughput,
...
+
+    Parameters
+    ----------
+    meta : eureka.lib.readECF.MetaClass
+        The metadata object.
+    sld : original sld object from ExoTIC phoenix grid
+    sld_new : sld object after re-scaling mu/intensity.
```
kevin218, line 507: check other docs for correct formatting.

demos/JWST/S4_template.ecf
```diff
@@ -38,6 +38,7 @@ logg            4.0     # Surface gravity in log g
 minmu			0.2		# minimum mu to use when calculating LD in ExoTiC-LD
 exotic_ld_direc /home/User/exotic-ld_data/ # Directory for ancillary files for exotic-ld, download from: https://zenodo.org/doi/10.5281/zenodo.6047317
 exotic_ld_grid  stagger # You can choose from kurucz (or 1D), stagger (or 3D), mps1, or mps2 model grids, or custom (to use custom_si_grid below)
+# rescale_phoenix  True # Option to rescale the mu grid for phoenix models. See Section 2.2: https://academic.oup.com/mnras/article/450/2/1879/985166
```
kevin218, line 41: Don't forget to add a detailed description in `docs/source/ecf.rst`.  This would be the best place to reference Nestor's paper for more information.

kevin218 (changes requested):

taylorbell57:
@kevin218, had you intended to merge this PR before your requested changes had been implemented?

kevin218:
I implemented the changes myself and tried to pushed them to this PR, but it looks like they went straight to main. I think it's because Erin is working from a clone rather than a branch. The good news is that the changes have been implemented and this PR can be closed.

taylorbell57:
Ah gotcha, I had a similar issue with ExoCTK recently. Eventually you should look at change the protections against pushes directly to the main branch to be controlled by a Ruleset rather than a Branch Protection rule (which doesn't protect admins from accidental pushes to main). I'll send a slack message with more details

PR #849, opened 2026-05-26
Support glob-filtered allapers inputdirs for Stages 4-6

taylorbell57:
Allow allapers=True runs to use glob-style inputdir values to narrow previous-stage aperture/background folders before the usual allapers loop runs. The glob is expanded relative to topdir through existing inputdir handling, and matched folders are mapped back to ap/bg pairs so Stages 4, 5, and 6 process only the requested combinations.

Add logging for expanded allapers glob searches and a clear error when no candidate folders match. Preserve existing behavior for normal inputdir values and for glob-like inputdir values when allapers is False.

Document nested Eureka output examples such as
`Stage3/S3_*_run*/ap5_bg*`, and add focused tests for normal allapers behavior, glob filtering, multiple background matches, no-match errors, and allapers=False behavior.


src/eureka/lib/manageevent.py
```diff
@@ -3,12 +3,167 @@
...
+
+    candidates = getattr(meta, 'allapers_inputdir_candidates', None)
+    if candidates is None:
+        return meta
+
+    pair_dirs = {}
+    for folder in candidates:
+        for part in reversed(folder.rstrip(os.sep).split(os.sep)):
+            match = re.search(r'(?:^|_)ap(?P<ap>[^_/]+)_bg'
+                              r'(?P<bg>.+?)(?:_run\d+)?$', part)
```
kevin218: Add documentation (either here or in the docstring) explaining the naming convention expected by the ap/bg extraction regex.

tests/test_allapers_inputdir.py
```diff
@@ -0,0 +1,118 @@
...
+        tmp_path / 'Stage3' / 'S3_2026-01-01_testevent_run1' / 'ap5_bg7',
+        5, 7)
+
+    meta = _make_current_meta(
+        tmp_path, 'Stage3/S3_*_run*/ap5_bg*', allapers=False)
+    s3_meta, _, _ = me.findevent(meta, 'S3', allowFail=False)
+
+    assert not hasattr(meta, 'allapers_inputdir_candidates')
+    assert s3_meta.spec_hw == 5
+    assert s3_meta.bg_hw == 7
```
kevin218, line 150: Copilot suggests adding an additional test case:

```
def test_allapers_glob_folders_match_but_no_ap_bg_overlap(tmp_path):
    """Glob matches folders, but no ap/bg values match previous stage ranges."""
    # Create folders with ap10_bg10 (outside prev stage's ranges)
    _save_previous_meta(
        tmp_path / 'Stage3' / 'S3_2026-01-01_testevent_run1' / 'ap10_bg10',
        10, 10)
    
    meta = _make_current_meta(tmp_path, 'Stage3/S3_*_run*/ap10_bg*', allapers=True)
    
    with pytest.raises(AssertionError, match='none of their ap/bg values matched'):
        me.filter_allapers_inputdir(meta)
```

kevin218 (changes requested):

kevin218 (approved):

PR #853, opened 2026-06-23
Add code to overplot empirical trace for NIRSpec/NIRCam/MIRI

jbrande:
NIRISS/SOSS median flux plots have the nominally-computed spectral trace from PASTASOSS overplotted on them. While NIRSpec, NIRCam, and MIRI don't have precomputed spectral traces, we can still overplot the empirically-identified center-of-mass spectral trace as found in `straighten.find_median_column_shifts()`. This PR takes the smoothed trace and overplots in on Fig. 3308 for non-NIRISS instruments.

An example plot looks like this (NRS2 data): 
fig3308_file0_MedianFrame

### DIFF
--- a/src/eureka/S3_data_reduction/plots_s3.py
+++ b/src/eureka/S3_data_reduction/plots_s3.py
@@ -15,6 +15,7 @@ warnings.filterwarnings("ignore", message='Ignoring specified arguments in '
                                           'this call because figure with num')
 
 from .source_pos import gauss
+from .straighten import find_column_median_shifts
 from ..lib import util, plots
 
 
@@ -706,9 +707,9 @@ def median_frame(data, meta, m, medflux, order=None):
                vmin=vmin, vmax=vmax, interpolation='nearest',
                extent=[xmin, xmax, ymin, ymax], cmap=cmap)
 
-    # Overplot nominal spectral trace for NIRISS SOSS
-    # (TODO: Probably also useful for NIRSpec, etc.)
-    if meta.inst == "niriss":
+    # Plot spectral traces over median frame
+    if meta.inst == "niriss": # nominal trace for NIRISS/SOSS
+        # SUBSTRIP96 only deals with a single order
         if data.attrs['mhdr']['SUBARRAY'] == 'SUBSTRIP96':
             order = 1
             trace = data["trace"].sel(order=order)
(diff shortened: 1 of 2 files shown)


src/eureka/S3_data_reduction/plots_s3.py
```diff
@@ -717,15 +718,23 @@ def median_frame(data, meta, m, medflux, order=None):
...
             plt.plot(data.flux.x.values, trace,
                      "--k", lw=1, label=f"Trace Order {order}")
-        else:
+        else: # SUBSTRIP256 handles both orders
             for order in meta.all_orders:
                 trace = data["trace"].sel(order=order)
                 plt.plot(data.flux.x.values, trace,
                          c="white", lw=2)
                 plt.plot(data.flux.x.values, trace,
                          "--k", lw=1, label=f"Trace Order {order}")
```
kevin218: Can we combine this if/else statement into one set of plotting commands? I think the `else` statement should work for SUBSTRIP96.

src/eureka/S3_data_reduction/straighten.py
```diff
@@ -18,11 +18,17 @@ def find_column_median_shifts(data, meta, m):
...
     m : int
         The file number.
+    plot_3308 : boolean
+        Flag to determine if this method is being called to 
+        get the smoothed spectral trace to plot over Fig. 3308
 
     Returns
     -------
     shifts : ndarray (2D)
         The shifts to apply to each column to straighten the trace.
```
kevin218: Currently only returned if plot_3308 is False. Either update the docs here or have it always returned on Line 97.

kevin218 (changes requested):
Be sure to fix the Flake8 issues.

src/eureka/S3_data_reduction/straighten.py
jbrande: Tweaked to make clearer.

src/eureka/S3_data_reduction/plots_s3.py
jbrande: Yes, I'm not sure how I missed this originally, but looping over `orders` rather than `all_orders` handles this. Done.

src/eureka/S3_data_reduction/straighten.py
kevin218: `shifts` is still only returned if plot_3308 is False. That's fine, but that information should be captured in the doc string.

kevin218 (changes requested):

PR #855, opened 2026-06-23
Improving WFC3 support

taylorbell57:
## Summary

This PR collects a set of bug fixes and quality-of-life improvements for WFC3 reductions, with most of the changes focused on making Stage 3 and Stage 4 behave more robustly for HST/WFC3 data products. It also includes a few related Stage 5 fitting and plotting fixes that surfaced while exercising the WFC3 workflow end-to-end.

## Main changes

- Improves WFC3 read bookkeeping for FLT inputs by setting `nreads_full` consistently and using `meta.nreads` when assigning integration ranges.
- Handles WFC3 scan-direction ordering more carefully when selecting reference frames, accounting for both `iref` parity and the first scan direction.
- Makes WFC3 timestamp handling more robust when one or more JD values are NaN, including the case where the first timestamp is missing.
- Applies manual masks earlier in Stage 3 so detector-coordinate mask regions are applied before downstream background subtraction and extraction steps.
- Carries WFC3 median-frame aperture data into optimal extraction so `fittype='meddata'` works for WFC3, and raises a clear error if meddata is requested but unavailable.
- Preserves WFC3 DQ handling for both multi-read and FLT-style inputs.
- Fixes the logged WFC3 centroid coordinate order.
- Updates WFC3 read summing in Stage 4 to use masked arrays, combine partially masked reads more gracefully, and only mask a wavelength bin when all contributing reads are masked.
- Adjusts white-light mask aggregation and drift-correction plot channel edge labels in Stage 4.

## Stage 5 and related fixes

- Adds `stretch_scale` as an `emcee` configuration option, defaulting to `2.0`, and passes it to `EnsembleSampler`.
- Moves the residual-distribution diagnostic from Fig. 5302 to Fig. 5502 and only generates it for `isplots_S5 >= 5`, since it really isn't a particularly informative figure.
- Fixes one-dimensional limb-darkening coefficient reshaping when loading coefficients from file.
- Makes `HSTRampModel` use the minimum available time as the local zero point and supplies safer defaults for the HST orbital ramp parameters `h4` and `h5`.
- Logs what `dynesty` resolves `run_sample='auto'` to, which should make sampler behavior easier to diagnose.
- Ensures `sigma_thresh` in S4cal metadata is stored as a list.

## Documentation and demos

- Updates the HST/WFC3 demo runner to set `CRDS_SERVER_URL` to the HST CRDS server.
- Updates Stage 5 output documentation and media references for the residual-distribution plot renumbering.
- Cleans up several diagnostic plot details, including WFC3/S3 aspect ratios and legend handling.


src/eureka/S3_data_reduction/plots_s3.py
```diff
@@ -107,9 +109,9 @@ def lc_nodriftcorr(meta, wave_1d, optspec, optmask=None, scandir=None,
         im1 = ax1.pcolormesh(np.arange(meta.n_int), wave_1d,
                              normspec.swapaxes(0, 1), vmin=meta.vmin,
                              vmax=meta.vmax, cmap=cmap)
-        im2 = ax2.imshow(normspec.swapaxes(0, 1), origin='lower',
-                         aspect='auto', extent=[0, meta.n_int, pmin, pmax],
-                         vmin=meta.vmin, vmax=meta.vmax, cmap=cmap)
+        im2 = ax1.pcolormesh(np.arange(meta.n_int), optspec.x.values,
```
kevin218: I think this should be `ax2.pcolormesh` since the previous version was `ax2.imshow`

kevin218 (changes requested):

PR #857, opened 2026-06-23
Add isort import sorting

taylorbell57:
## Summary

This PR adds `isort` to the project tooling and applies the resulting import-ordering changes across the repository.

The large file count is expected: almost all changed files only have mechanical import sorting from `isort`. Reviewers can focus on the small set of manual changes listed below.

## Manual changes

- `.pre-commit-config.yaml`
  - Adds the `pycqa/isort` pre-commit hook so import ordering is checked locally and in pre-commit workflows.

- `pyproject.toml`
  - Adds the project `isort` configuration:
    - `line_length = 79`
    - `known_first_party = ["eureka"]`
    - `src_paths = ["src"]`
  - Adds `isort` to the `dev` dependency group.

- `src/eureka/S5_lightcurve_fitting/models/BatmanModels.py`
  - Changes the `Model` import from the package-level import to the explicit class import:
    - from `from . import Model`
    - to `from .Model import Model`
  - This avoids importing the module object when the code needs the `Model` class. This file also has normal `isort` cleanup around the imports.

## Generated changes

All other modified Python files are mechanical import-ordering changes produced by running `isort` with the new project configuration. These should not change runtime behavior.

## Verification

Ran:

```bash
python -m isort src tests demos docs/source/conf.py --check-only
git diff --check origin/main..HEAD
```

Both checks passed.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -19,6 +19,11 @@ repos:
       - id: check-added-large-files
       - id: debug-statements
 
+  - repo: https://github.com/pycqa/isort
+    rev: 5.13.2
+    hooks:
+      - id: isort
+
   - repo: https://github.com/pycqa/flake8
     rev: 7.3.0
     hooks:
--- a/demos/HST/run_eureka_wfc3.py
+++ b/demos/HST/run_eureka_wfc3.py
@@ -1,4 +1,5 @@
 import os
+
 import eureka.lib.plots
 import eureka.S3_data_reduction.s3_reduce as s3
 import eureka.S4_generate_lightcurves.s4_genLC as s4
--- a/demos/JWST/CalibratedStellarSpectra/run_eureka.py
+++ b/demos/JWST/CalibratedStellarSpectra/run_eureka.py
@@ -1,4 +1,5 @@
 import os
+
 import eureka.lib.plots
 import eureka.S1_detector_processing.s1_process as s1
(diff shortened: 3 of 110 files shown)


taylorbell57:
One CI follow-up for this PR: this adds `isort` to `.pre-commit-config.yaml`, but the current GitHub Actions workflows do not run `isort` or `pre-commit`. The existing flake8 workflow only runs `flake8`, so import sorting would be enforced locally but not on PRs.

I think the cleanest option is for @kevin218 to enable `pre-commit.ci` for this repo. The config already has a `ci:` block and skips `flake8`, so `pre-commit.ci` would run the lightweight formatting/check hooks, including `isort`, while leaving flake8 to the existing GitHub Action. It would also optionally auto-fix PRs and do monthly hook autoupdates.

If we do not want to rely on `pre-commit.ci`, then we should add a GitHub Actions job that runs:

```bash
python -m pip install pre-commit
pre-commit run --all-files
```

My preference would be `pre-commit.ci` as the primary enforcement, since it matches the existing `.pre-commit-config.yaml` and avoids duplicating hook logic in Actions.

To enable `pre-commit.ci`, you will need to install or enable the GitHub App from https://github.com/apps/pre-commit-ci. Once the app has access to the repo, it will automatically read `.pre-commit-config.yaml` and start running on new PRs; no extra workflow file is needed.

kevin218 (approved):

PR #870, opened 2026-08-07
Fix nested output directories causing FileNotFoundError on multi-aperture runs

taylorbell57:
## Bug

`MetaClass.__setattr__` (`src/eureka/lib/readECF.py`) auto-derives
`outputdir_raw` from whatever `outputdir` is assigned:

```python
elif item == 'outputdir' and hasattr(self, 'topdir'):
    if self.topdir in value:
        value = value.replace(self.topdir, '')
    self.outputdir_raw = os.path.join(value, '')
    value = os.path.join(self.topdir, *value.split(os.sep))
```

Stages 3/4/5/6 reuse a single `meta` object across their loop over
(spec_hw, bg_hw) combinations, and each iteration does
`meta.outputdir = util.pathdirectory(meta, 'S#', run, ap=.., bg=..)` — a
fully-resolved absolute path. Assigning it re-triggers the setter above,
which strips `topdir` and stores the *already fully-resolved run+ap+bg
subfolder* back into `outputdir_raw`. On the next (ap, bg) iteration,
`util.pathdirectory`/`makedirectory` rebuild the root directory from that
corrupted `outputdir_raw`, so the new run+ap+bg folder gets nested **inside**
the previous one instead of being created alongside it. Only the flat
(non-nested) directory was pre-created by the dir-creation pre-loop, so
opening the per-aperture log file for the 2nd+ combination raises
`FileNotFoundError`.

This affects any Stage 3 run with more than one `photap`/`skyin`/`skywidth`
combination, and identically affects Stage 4/5/6 whenever `allapers=True`
processes more than one aperture pair from a prior stage. It is not caught
by the existing test suite since no test currently exercises a multi-value
aperture grid end-to-end. The Stage 1/3-4 optimizers are unaffected since
each sweep evaluates a single combination via a freshly deep-copied `meta`.

## Fix

- Added an optional `outputdir_raw` kwarg to `util.pathdirectory()` and
  `util.makedirectory()`. When provided, it's used instead of reading
  `meta.outputdir_raw`, so callers can supply an immutable, clean base path
  that isn't affected by the `outputdir` setter's side effect. Omitting it
  preserves the previous behavior, so unaffected callers (S1/S2/S4cal,
  optimizers) need no changes.
- In `s3_reduce.py`, `s4_genLC.py`, `s5_fit.py`, and `s6_spectra.py`: capture
  the clean base directory once, right after `set_defaults()`/before any
  aperture loop, and pass it explicitly to every `makedirectory()`/
  `pathdirectory()` call in both the directory-creation loop and the main
  processing loop.

## Tests

Added `tests/test_outputdir_raw.py`:
- `test_pathdirectory_without_base_reproduces_nesting_bug` documents the old
  hazard (nesting occurs if the base isn't passed explicitly).
- `test_pathdirectory_with_explicit_base_stays_flat` confirms the fix keeps
  three consecutive aperture/annulus directories as flat siblings.

Full `pytest tests/` suite (49 tests, including MIRI/NIRCam/NIRSpec/NIRISS
end-to-end pipeline runs) passes with no regressions.

### DIFF
--- a/src/eureka/S3_data_reduction/s3_reduce.py
+++ b/src/eureka/S3_data_reduction/s3_reduce.py
@@ -119,6 +119,9 @@ def reduce(eventlabel, ecf_path=None, s2_meta=None, input_meta=None):
     # create directories to store data
     # run_s3 used to make sure we're always looking at the right run for
     # each aperture/annulus pair
+    # Cache the clean base dir since meta.outputdir_raw gets overwritten
+    # below each time meta.outputdir is set to a per-pair directory
+    base_outputdir_raw = meta.outputdir_raw
     meta.run_s3 = None
     for spec_hw_val in meta.spec_hw_range:
         for bg_hw_val in meta.bg_hw_range:
@@ -126,8 +129,9 @@ def reduce(eventlabel, ecf_path=None, s2_meta=None, input_meta=None):
             # Directory structure should not use expanded HW values
             spec_hw_val, bg_hw_val = util.get_unexpanded_hws(
                 meta.expand, spec_hw_val, bg_hw_val)
-            meta.run_s3 = util.makedirectory(meta, 'S3', meta.run_s3,
-                                             ap=spec_hw_val, bg=bg_hw_val)
+            meta.run_s3 = util.makedirectory(
+                meta, 'S3', meta.run_s3, ap=spec_hw_val, bg=bg_hw_val,
+                outputdir_raw=base_outputdir_raw)
 
     # begin process
     for spec_hw_val in meta.spec_hw_range:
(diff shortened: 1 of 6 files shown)


src/eureka/S4_generate_lightcurves/plots_s4.py
```diff
@@ -518,26 +519,22 @@ def plot_rescaled_phoenix(meta, sld, sld_new):
...
+             label='Original Phoenix Intensity Profile')
+    plt.plot(sld_new.mus, sld_new.I_mu, color='tomato', lw=1.0,
+             marker='o', ms=6,
+             label='Rescaled Phoenix Intensity Profile')
+
+    plt.xlim(-0.05, 1.05)
+    plt.ylim(-0.05, 1.05)
+
+    plt.ylabel('I($\\mu$)')
+    plt.xlabel('$\\mu$')
```
kevin218, line 533: Is there a reason to switch from `\` to `\\` here or is this just AI being AI?

taylorbell57: It's for flake8 which had been complaining that `\m` is not a valid character — there were several flake8 issues introduced by your accidental push to `main` before

taylorbell57: The alternative is to make it a raw string like `r'I($\mu$)'` which results in the same thing as `'I($\\mu$)')`

kevin218 (approved):

PR #876, opened 2026-08-14
Rename mad

### LINKED ISSUE #852: [Enhancement]: Rename MAD to MAeD or MAED
### Instrument

Other (any stage)

### What is your suggestion?

The MAD label is confusing because people assume that is means median absolution deviation (which is the standard definition) when it actually computes the median absolution element difference (via np.ediff1d). Relabelling MAD to MAed or MAED would make it clear that we're performing a different calculation.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

arnav-agrawal:
Updated MAD to MAED everywhere in the code, including variable names, figure captions, and comments. Also updated the term Median Absolute Deviation to Median Absolute Element Difference wherever the acronym was fully typed out. Note that Mean Absolute Deviation was left as is, as I assume that is different.

Regression test reference files were updated and overwritten to account for this change.


kevin218:
Mean absolute deviation may be just a typo. Do you recall where it was mentioned? If not, I'll look into it more on Monday.

arnav-agrawal:
> Mean absolute deviation may be just a typo. Do you recall where it was mentioned? If not, I'll look into it more on Monday.

Yeah if you look at some of the ECF files in the Files Changed Tab (e.g. `demos/JWST/CalibratedStellarSpectra/S3_nircam_wfss_template.ecf`), there is a`mean` option for `bg_method`). In that code comment it says `Mean Absolute Deviation`.

src/eureka/S3_data_reduction/hst_scan.py
```diff
@@ -307,7 +307,7 @@ def makeflats(flatfile, wave, xwindow, ywindow, flatoffset, n_spec, ny, nx,
             ilow = np.nonzero(flat_norm < 1)
             # Make distribution symetric about 1
             dbl = np.concatenate((flat_norm[ilow], 1+(1-flat_norm[ilow])))
-            # MAD
+            # MAED
```
kevin218, line 310: This should still be MAD

src/eureka/S3_data_reduction/hst_scan.py
```diff
@@ -316,7 +316,7 @@ def makeflats(flatfile, wave, xwindow, ywindow, flatoffset, n_spec, ny, nx,
             ihi = np.nonzero(flat_norm > 1)
             # Make distribution symetric about 1
             dbl = np.concatenate((flat_norm[ihi], 2-flat_norm[ihi]))
-            # MAD
+            # MAED
```
kevin218, line 319: This should still be MAD

src/eureka/S3_data_reduction/hst_scan.py
```diff
@@ -401,7 +401,7 @@ def makeBasicFlats(flatfile, xwindow, ywindow, flatoffset, ny, nx, sigma=5,
         ilow = np.nonzero(flat_norm < 1)
         # Make distribution symetric about 1
         dbl = np.concatenate((flat_norm[ilow], 1+(1-flat_norm[ilow])))
-        # MAD
+        # MAED
```
kevin218, line 404: Same

src/eureka/S3_data_reduction/hst_scan.py
```diff
@@ -410,7 +410,7 @@ def makeBasicFlats(flatfile, xwindow, ywindow, flatoffset, ny, nx, sigma=5,
         ihi = np.nonzero(flat_norm > 1)
         # Make distribution symetric about 1
         dbl = np.concatenate((flat_norm[ihi], 2-flat_norm[ihi]))
-        # MAD
+        # MAED
```
kevin218, line 413: Maybe we need a helper function if this code is repeated 4 times

kevin218 (changes requested):
I think these changes will break backwards compatibility. We may want to have code in Stage  4 that creates MAED from MAD of it doesn't exist.

src/eureka/S3_data_reduction/hst_scan.py
arnav-agrawal: Changed these MAEDs back to MAD. Created a helper function at the bottom of the file. 

I could delete the second occurrence of the `std` calculation (lines 319 and 413) since it's in scope of the previous definition, but I left it for clarity for the `std` calculation used for outlier rejection from high side and the low side.

arnav-agrawal:
> I think these changes will break backwards compatibility. We may want to have code in Stage 4 that creates MAED from MAD of it doesn't exist.

Makes sense. For backwards compatibility do you mean the ability to run old ECFs with the latest code, new ECFs with old code, or both? I assume both.

arnav-agrawal:
> I think these changes will break backwards compatibility. We may want to have code in Stage 4 that creates MAED from MAD of it doesn't exist.

Updated the code so that it works with old ECFs and old output products (the HDF5s from S3 and S4) in case people want to start from Stage 5 for example.

kevin218 (approved):

PR #877, opened 2026-08-17
Explicit SOSS X and Y offset parameters

jbrande:
Adding custom X offset capabilities to SOSS stage 3 analyses. While we previously allowed for custom Y offsets, we now let users pass in arbitrary X offsets. The code checks the FITS headers for `XOFFSET` and `YOFFSET` values in the dither information, which are superseded by explicit `trace_xoffset` and `trace_yoffset` values in the ECF. The defaults for these parameters are `None`, and no offset is applied from the header if the dither is less than 1/100 of a pixel (or ~0.7 mas).

Also, fixes S3 fig. 3308 to properly do trace plotting for Orders 1 and 2 in SUB96 vs SUB256 by checking `meta.orders` as opposed to `meta.all_orders`.

### DIFF
--- a/src/eureka/S3_data_reduction/niriss.py
+++ b/src/eureka/S3_data_reduction/niriss.py
@@ -147,12 +147,28 @@ def get_wave(data, meta, log):
     log.writelog(f"  The NIRISS pupil position is {pwcpos:3f} degrees",
                  mute=(not meta.verbose))
 
-    # Calculate offset amount from header (given in arcsec) and
-    # NIRISS pixel scale (0.0653 arcsec/px in X direction)
-    pixscale = 0.0653
-    trace_xoffset = data.attrs['mhdr']['XOFFSET'] / pixscale
-    log.writelog(f"  There is an X offset of {trace_xoffset:.3f} pixels",
-                 mute=(not meta.verbose))
+
+    # Check meta for trace offsets, and if not present,
+    # calculate offset amounts from header (given in arcsec) and
+    # NIRISS pixel scale (0.0653 arcsec/px in X direction, 0.0658 in Y)
+    trace_xoffset = meta.trace_xoffset
+    if trace_xoffset is not None:
+            trace_xoffset = data.attrs['mhdr']['XOFFSET'] / 0.0653
+
+    trace_yoffset = meta.trace_yoffset
+    if trace_yoffset is not None:
+        trace_yoffset = data.attrs['mhdr']['YOFFSET'] / 0.0658
+
+    if np.abs(trace_xoffset) > 0.001:
+        log.writelog(f"  There is an X offset of {trace_xoffset:.3f} pixels",
+                    mute=(not meta.verbose))
+    else:
+        trace_xoffset = 0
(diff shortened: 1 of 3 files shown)


src/eureka/S3_data_reduction/niriss.py
```diff
@@ -147,12 +147,28 @@ def get_wave(data, meta, log):
...
-    pixscale = 0.0653
-    trace_xoffset = data.attrs['mhdr']['XOFFSET'] / pixscale
-    log.writelog(f"  There is an X offset of {trace_xoffset:.3f} pixels",
-                 mute=(not meta.verbose))
+
+    # Check meta for trace offsets, and if not present,
+    # calculate offset amounts from header (given in arcsec) and
+    # NIRISS pixel scale (0.0653 arcsec/px in X direction, 0.0658 in Y)
+    trace_xoffset = meta.trace_xoffset
+    if trace_xoffset is not None:
```
kevin218: Should be: `if trace_xoffset is None:`

src/eureka/S3_data_reduction/niriss.py
```diff
@@ -147,12 +147,28 @@ def get_wave(data, meta, log):
...
+            trace_xoffset = data.attrs['mhdr']['XOFFSET'] / 0.0653
+
+    trace_yoffset = meta.trace_yoffset
+    if trace_yoffset is not None:
```
kevin218: Should be: `if trace_yoffset is None:`

src/eureka/S3_data_reduction/niriss.py
```diff
@@ -171,11 +187,11 @@ def get_wave(data, meta, log):
...
             trace = get_soss_traces(pwcpos=pwcpos,
                                     order=str(order), interp=True)
-        if data.attrs['mhdr']['SUBARRAY'] == 'SUBSTRIP96' and \
-                meta.trace_yoffset is None:
+        #if data.attrs['mhdr']['SUBARRAY'] == 'SUBSTRIP96' and \
+        #       meta.trace_yoffset is None:
             # PASTASOSS doesn't account for different substrip starting rows;
             # therefore, set trace offset to best guess (-12 pixels).
-            meta.trace_yoffset = -12
+        #    meta.trace_yoffset = -12
```
kevin218: Delete commented out code

src/eureka/S3_data_reduction/niriss.py
kevin218: Should we do the following to keep the -12 default for SUBSTRIP96? Looking at `data.attrs['mhdr']['YOFFSET'] / 0.0658`, I see that it is 0 for SUBSTRIP96, which will yield an incorrect result.
```
if trace_yoffset is None:
    if data.attrs['mhdr']['SUBARRAY'] == 'SUBSTRIP96':
        trace_yoffset = -12
    else:
        trace_yoffset = data.attrs['mhdr']['YOFFSET'] / 0.0658
```

kevin218 (changes requested):

PR #879, opened 2026-09-03
merge uncals script

hdiamondlowe:
This rough script will merge uncal segments. I run it before starting Eureka! Stage 1. Written for MIRI photometry, not sure how easily it will map to other instruments/modes.


taylorbell57:
Thanks for submitting this Hannah! As I'd mentioned on Slack, I'll take a look at this in the coming weeks (perhaps not until October) and rework it to work for all our supported JWST modes; having this initial draft will make it way easier to do so, so thank you!

